PR #60, opened 2015-01-12
Temporarily removed the links to the handin 

icanberk:
Handin page since it doesn't exist anymore.

Ideally, these buttons should pop-up a handing modal. Will work on this.


dlbucci:
PS These links are not really the Rails way to do links.  Use link_to and pass it a block if you want more complex HTML inside of it.

PR #62, opened 2015-01-12
Remove Waves.js from Home Layout

dlbucci:
It breaks links on certain browsers.

### DIFF
--- a/app/views/layouts/home.html.erb
+++ b/app/views/layouts/home.html.erb
@@ -6,7 +6,6 @@
 
     <%= stylesheet_link_tag 'style' %>
     <%= stylesheet_link_tag 'animate' %>
-    <%= stylesheet_link_tag 'waves' %>
 
     <%= external_stylesheet_link_tag "bootstrap", "3.0.0" %>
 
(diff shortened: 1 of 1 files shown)


Z1MM32M4N:
@dlbucci, I see you've assigned me here. Is there something I should be reviewing here, or was this just a heads-up? It looks like waves.js is somehow connected to sorting tables? I'm not quite sure what conditions need to be met to merge in this change.

dlbucci:
@Z1MM32M4N No, I just thought I'd give you a pull request :)  Waves.js appears to be entirely cosmetic, and is supposed to animate link clicks.  Instead, it appears to break links.  You can probably just pull.

Z1MM32M4N:
Sounds good :D

Z1MM32M4N:
@dlbucci @icanberk What kinds of things was Waves.js used for? It looks like not all the remnants of Waves.js were removed, as some files still reference them.

icanberk:
@Z1MM32M4N it was only to do material design style animations on the buttons. (https://github.com/fians/Waves) We can really get rid of all of it for now.

Z1MM32M4N:
ok.

PR #85, opened 2015-01-14
Fixed job feedback page

### LINKED ISSUE #84: Jobs page is incorrectly appending .img to VM name
If the user is requesting rhel.img, the job parameters section of the feedback report is incorrectly listing the VM as rhel.img.img. However, this is only a cosmetic problem, as the front end is sending the correct VM image name to Tango.

ymzong:
Fixes #84 by removing <tt>.img</tt> after VM Image Name.

### DIFF
--- a/app/views/jobs/getjob.html.erb
+++ b/app/views/jobs/getjob.html.erb
@@ -15,9 +15,9 @@
 
 <h2>Job Parameters</h2>
 <ul>
-<li>VM image: <kbd><%= @job[:rjob]["vm"]["name"]%>.img</kbd>
-<li><kbd>timeout</kbd>: <%= @job[:rjob]["timeout"] %> secs
-<li><kbd>maxOutputFileSize</kbd>: <%= @job[:rjob]["maxOutputFileSize"] %> bytes.
+<li>VM image: <%= @job[:rjob]["vm"]["name"]%>
+<li>Job Timeout: <%= @job[:rjob]["timeout"] %> secs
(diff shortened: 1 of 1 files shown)


app/views/jobs/getjob.html.erb
```diff
@@ -31,10 +31,10 @@
 </tr></table>
 
 <% if @cud.user.administrator?  then %>
-<li><kbd>notifyURL</kbd>: <kbd><%= @job[:rjob]["notifyURL"] %> </kbd>
+<li><kbd>notifyURL</kbd>: <%= @job[:rjob]["notifyURL"] %>
```
dlbucci: Why is this the only place you kept <kbd>?

ymzong: @dlbucci Because notifyURL is kind of a "reserved word", while I changed the other lines to normal English whenever possible...

dlbucci: Why couldn't this be changed to normal English? It just bugs me that one of five is special like this, but it's not a huge issue.

ymzong: @dlbucci I actually feel the same -- what do you suggest me to change this line into? I thought `Notify URL` might confuse people... What about `Callback URL`, or other options?

dlbucci: I think Notify URL would be fine, especially since it's just making the property look nice.  But you can just make a decision and merge when you're ready.

PR #90, opened 2015-01-15
Clean up trailing white spaces

ymzong:
By using `find . -name '*.rb' -exec sed -i 's/[[:blank:]]*$//g' '{}' \;` in shell.

### DIFF
--- a/app/controllers/admins_controller.rb
+++ b/app/controllers/admins_controller.rb
@@ -3,15 +3,15 @@ require('fileutils')
 
 
 class AdminsController < ApplicationController
-  
+
   CMU_ROSTER_COLUMNS = 29
 
   action_auth_level :show, :instructor
   def show
-    @options = 
-      [{"name"=>"Edit course", path: [:edit, @course], 
+    @options =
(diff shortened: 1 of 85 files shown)


dlbucci:
Jesus.  Why did we need this?

Z1MM32M4N:
Trailing whitespaces are generally bad style/practice and from time to time they can make certain tools get confused, for example with revision control systems and text-only email systems.

The one thing I always check for when mass-removing whitespace from a project is that no whitespace was removed from anything that could be treated as a string, like a heredoc or some sort of markup (Markdown, HTML, etc.). I scanned over the changes and didn't find any instances where the removal of the whitespaces would have impacted something like this, so I'd say we're good to go. It's probably best to merge something like this sooner rather than later to avoid weird merge conflicts.

Z1MM32M4N:
@ymzong Just to confirm, did you mean to close and not merge this PR?

ymzong:
@Z1MM32M4N That's right -- currently it already requires manual merging. I'd rather leave it when code changes slow down.

PR #92, opened 2015-01-15
Filter import roster for well-formed entries

### LINKED ISSUE #67: Empty lines in class roster breaks Import Roster
As in `admins_controller.rb#L468`, any blank lines present in the csv file would break the roster-parsing routine.

To fix the issue, we should silently ignore empty lines and report errors for incomplete lines. (Currently both would lead to `500 Internal Error`.)

ymzong:
Fixes #67.

Filters the import roster such that only lines with the "correct" number of cells are processed. This is mainly used to ignore empty lines / incomplete lines.

### DIFF
--- a/app/controllers/admins_controller.rb
+++ b/app/controllers/admins_controller.rb
@@ -4,7 +4,7 @@ require('fileutils')
 
 class AdminsController < ApplicationController
   
-  CMU_ROSTER_COLUMNS = 29
+  NEW_ROSTER_COLUMNS = 29
 
   action_auth_level :show, :instructor
   def show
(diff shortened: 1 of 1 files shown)


app/controllers/admins_controller.rb
```diff
@@ -475,6 +475,8 @@ def detectAndConvertRoster(roster)
       # First Name(6), (skip)(7), Andrew ID(8), (skip)(9), School(10), 
       # Major(11), Year(12), (skip)(13), Grade Policy(14), ...
 
+      # Sanitize roster input, ignoring empty / incomplete lines.
+      parsedRoster.select! { |row| row.length == CMU_ROSTER_COLUMNS }
```
dlbucci: Earlier on, we check that the first element in parsedRoster is this length too.  Can't we merge these two statements?  (Also, maybe we should rename the constant to ROSTER_COLUMNS.  It's based off of CMU, but there's no reason other schools couldn't prepare data in the same format)

ymzong: @dlbucci I changed `CMU_ROSTER_COLUMNS` to `NEW_ROSTER_COLUMNS`.

However, we still need to check `length == NEW_ROSTER_COLUMNS` twice because the first time is when we distinguish if the roster is of new format or old format; the second time is when I filter for incomplete rows, assuming the roster is of new format.

PR #93, opened 2015-01-15
update user information when uploading roster

andrewwuan:
Also fix a bug that school, major, and year are skipped

### DIFF
--- a/app/controllers/admins_controller.rb
+++ b/app/controllers/admins_controller.rb
@@ -102,19 +102,32 @@ class AdminsController < ApplicationController
                 # Add this user to the course
                 # Look for this user
                 email = newCUD[:email]
+                first_name = newCUD[:first_name]
+                last_name = newCUD[:last_name]
+                school = newCUD[:school]
+                major = newCUD[:major]
+                year = newCUD[:year]
                 
                 if ((user = User.where(email: email).first).nil?)
                   # Create a new user
-                  user = User.roster_create(email, newCUD[:first_name],
-                                            newCUD[:last_name])
+                  user = User.roster_create(email, first_name, last_name, school,
+                                        major, year)
                   if (user.nil?)
                     throw NewUserCreationException
                   end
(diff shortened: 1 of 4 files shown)


dlbucci:
What do all the various colors mean?

icanberk:
Are those exceptions defined somewhere? (eg. GreenCUDExistInCourseException)

andrewwuan:
@dlbucci Colors are visual aid for uploadRoster page. Green means that roster contains new users, red means that user is going to be dropped, and black means that user exists before

andrewwuan:
@icanberk Nope.. I'll change them to string style

icanberk:
Can we also add that as a comment somewhere in the code?

dlbucci:
And has this been tested?  We do not want this feature breaking.

andrewwuan:
@dlbucci Yep it has been tested

andrewwuan:
@icanberk added

app/controllers/admins_controller.rb
```diff
@@ -102,19 +107,32 @@ def uploadRoster
...
+                  user.school = school
+                  user.major = major
+                  user.year = year
+                  user.save
                 end
                 
                 # Make sure this user doesn't have a cud in the course
                 if (@course.course_user_data.where(user: user).first)
-                  throw NewUserExistInCourseException
+                  throw GreenCUDExistInCourseException
```
dlbucci: And here.  (As a side note, Ruby has a raise and a throw? WTF? What hasn't this language absorbed from every other language ever?)

andrewwuan: Yep it has raise/rescue and throw/catch

dlbucci:
Do we also need to add a rescue somewhere?  From what I can tell, it's a different mechanism from try/catch.

andrewwuan:
@dlbucci The whole transaction block has a rescue statement at the end

PR #104, opened 2015-01-15
Removed Ruby backtrace for autograde feedback string parsing error

### LINKED ISSUE #86: Backtrace on autograding feedback
If parseautoresult() fails because the autograding code doesn't generated a JSON autoresult, then the feedback returned to the user shouldn't include the lengthy trace, which users find very confusing.

ymzong:
Fixes #86.

Also includes minor code cleanup.

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -639,9 +639,6 @@ class SubmissionsController < ApplicationController
 
       end
     rescue Exception => e
-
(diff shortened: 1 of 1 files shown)


dlbucci:
So were we printing the stack trace in Autolab2?

ymzong:
@dlbucci According to the code we probably were, but I can't think of many scenarios where it would actually help the instructor...

PR #105, opened 2015-01-16
Fix table sorting in Manage Accounts page. #1

Z1MM32M4N:
I removed Prototype.js, which was shadowing jQuery, and then made a call
to the `sorttables` function, which adds listeners on the table headings
to toggle sorting.

Prototype.js doesn't look like it has a place in our code, from what I
can tell. Correct me if I'm wrong. I'll be attempting to remove it in
favor of jQuery whenever I get the chance. Until it's "safe," I'll be
preferring `jQuery(...)` over `$(...)`, even though this is generally
not the best style.

I also disabled the functionality to float table headers at the top of
the screen, because it was causing errors and the current "solution" was
really hacky. I might look into enabling this feature later.

### DIFF
--- a/app/assets/javascripts/sorttable.js
+++ b/app/assets/javascripts/sorttable.js
@@ -1,16 +1,16 @@
-/* 
+/*
     sorttable.js
     Enables the sorting of tables on a per column basis.
     Sorting is stable and does not modify the elements being
     sorted in any way other than their relative position.
-    
+
     Usage:
         Give a table the class "sortable" and make sure that
         all of the elements you want sorted are wrapped
         in a <tbody> tag, it will take care
         of the rest. Whenever a user clicks on a particular
         heading, the table will be sorted based on that column.
-    
+
     Includes the sortElements jQuery plugin:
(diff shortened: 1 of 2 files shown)


Z1MM32M4N:
Does anyone know of any other tables in the app that need sorting fixed?

I'm still not that familiar with the app, so I don't know where tables are being used.

Z1MM32M4N:
@autolab/developers Does anyone know of any other tables in the app that need sorting fixed?

I'm still not that familiar with the app, so I don't know where tables are being used.

dlbucci:
Try 'grep -nr "<table" Autolab/app/views' ?

Z1MM32M4N:
```
$ ag '<table' | wc -l
72
```

:scream:

dlbucci:
Eh... we could also add it on a case-by-case basis.

Z1MM32M4N:
I added a note about gradually fixing the issue in #1, which we can leave open until we're confident the tables are all sortable. For now, let's merge this in.

PR #109, opened 2015-01-16
Sort all columns with the sortable class.

### LINKED ISSUE #108: Scoreboards should be sortable
Scoreboards should be sortable by columns (including submission timestamp, and points).

See old Autolab scoreboard for example.

Z1MM32M4N:
Fixes #108.

I'm assuming that any tables that include the class="sortable" expect
their contents to be sorted. The way to do this is

  a) include the sorttable.js file
  b) call the function `sorttables();` once the DOM has loaded

I've added step (b) to the sorttable.js file, so that it is done
automatically on all pages that include the JavaScript file.

sorttables.js is an external library, so if we ever update it, this will
get overwritten. However, it is much simpler than pecking and hunting
through the source code to include the three lines required to sort
columns in all the files that require it.

### DIFF
--- a/app/assets/javascripts/sorttable.js
+++ b/app/assets/javascripts/sorttable.js
@@ -183,3 +183,8 @@ var sorttables = (function($) { return function() {
 		});
 	});
 }})(jQuery);
+
+// Once the DOM has loaded, make the table columns sortable
+jQuery(document).ready(function() {
+  sorttables();
+});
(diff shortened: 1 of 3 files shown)


Z1MM32M4N:
Like I mentioned in #108, I haven't tested this, as I can't access a scoreboard locally. Any suggestions?

PR #111, opened 2015-01-16
Updates X-Frame-Options header to ALLOWALL(Fixes #110)

### LINKED ISSUE #110: Autolab denies embedding as iFrame by external sites
Professor Kosbie:

> Hi all.  We prefer to include autolab in an iframe in the 112 website, so our students get a "112 curated" view of autolab.  This, for example, let's them see the links to the schedule and the assigned hw while also viewing the autolab site.

> This has worked just fine for a long time.

> But with the move to the new Autolab, it no longer works.

ymzong:
Fixes #110.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -82,5 +82,8 @@ module Autolab3
     config.action_view.field_error_proc = Proc.new { |html_tag, instance| 
       "<div class=\"field_with_errors has-error\">#{html_tag}</div>".html_safe
     }
+
+    # Allow embedding as iFrame on external sites
+    config.action_dispatch.default_headers.merge!({'X-Frame-Options' => 'ALLOWALL'})
   end
 end


Z1MM32M4N:
This opens us up to click jacking attacks. I'm not sure how much we care about this, but I'm also not quite sure what the advantages of being able to embed a site in an iFrame are either.

@autolab/developers Thoughts?

dlbucci:
Kosbie wanted to be able to do that. If you want to be more careful about it (restrict most sites and create a whitelist), go ahead, just make sure the site works for 15-112.

droh:
Kosbie likes using things like autolab and piazza in ifram on his 15-112
page because he feels it gives the students the feeling that they've never
left the 112 environment.

-- David O'Hallaron
-- Prof of CS and ECE, Carnegie Mellon University
-- GHC 7517, x8-8199
-- http://www.cs.cmu.edu/~droh

Z1MM32M4N:
Should we maybe then only allow from *.cmu.edu pages? Is that something we can do?

ymzong:
@Z1MM32M4N According to [this](https://developer.mozilla.org/en-US/docs/Web/HTTP/X-Frame-Options), the only reliable options are `DENYALL`, `SAMEORIGIN`, or `ALLOWALL`.

icanberk:
Would SAMEORIGIN work since they are both cs.cmu.edu ?

ymzong:
@icanberk 
> SAMEORIGIN: The page can only be displayed in a frame on the same origin as the page itself.

I guess it means that an Autolab page can only appear as iframe on the exact same Autolab page, which is not very helpful...

icanberk:
"origin" here refers to the domain name. (http://en.wikipedia.org/wiki/Same-origin_policy)

ymzong:
@icanberk Cool! But that still doesn't help us much, does it? The host `https://autolab.cs.cmu.edu` is used exclusively by Autolab...

dlbucci:
But it can include subdomains as well, correct?

ymzong:
@dlbucci You can see some examples that @icanberk sent. According to the examples,

> exact match [of host name is] required.

Z1MM32M4N:
After reading that article, it looks like unfortunately our best bet long-term is to either wait for browsers to catch up with the spec or to encourage Kosbie to not use iFrames, though neither of these is all too desirable.

PR #125, opened 2015-01-18
Updated Nickname Restrictions

dlbucci:
This was tested and works on my Mac OS X.  I don't know if settings for the DB need to be changed, but I think this should be fine. (the nickname column actually has 255 characters, so we can allow longer names.)

### DIFF
--- a/app/models/course_user_datum.rb
+++ b/app/models/course_user_datum.rb
@@ -61,11 +61,8 @@ class CourseUserDatum < ActiveRecord::Base
   def valid_nickname?
     if not nickname
       true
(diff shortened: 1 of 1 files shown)


dlbucci:
PS Do we want nicknames to be CUD-specific or User-specific?

icanberk:
I think user-specific is good enough.

dlbucci:
Well they are CUD-specific at the moment, so we could change that.

icanberk:
Oh, I think cud-specific is much better. I just said good enough for user-specific one because it's not that urgent.

PR #128, opened 2015-01-19
Use uniform margin on `.card`s

### LINKED ISSUE #124: Courses don't wrap correctly when screen is resized
When the screen is resized, the spacing between courses becomes inconsistent.

Screenshot attached.

10433203_815692498477014_3528490103397129615_n

/cc @aarongut

Z1MM32M4N:
Commit Summary
--------------

ec652d9 (Jacob Zimmerman, 2 minutes ago)
   Stripped whitespace

cf04d88 (Jacob Zimmerman, 23 seconds ago)
   Use uniform margin on `.card`s. Fixes #124

### DIFF
--- a/app/assets/stylesheets/style.css.scss
+++ b/app/assets/stylesheets/style.css.scss
@@ -563,10 +563,6 @@ div.card .btn.main:hover {
   background-color: rgba(153, 0, 0, 0.9)
 }
 
-div.card:nth-child(3) {
-  margin-right: 0;
-}
-
 @media (max-width: 768px) {
 
   .rolodex {
--- a/app/assets/stylesheets/style.css.scss
+++ b/app/assets/stylesheets/style.css.scss
@@ -228,7 +228,7 @@ div.main-header img {
 
   .action-links.navbar-nav li {
     display: block;
-    border-bottom: 1px solid #fff;  
+    border-bottom: 1px solid #fff;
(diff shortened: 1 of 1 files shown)


Z1MM32M4N:
The change kind of gets lost in the midst of stripping whitespace, but the real changes occur on lines 566 - 569. If you'd like a more refined diff, I separated the request into two commits so that the real changes become more apparent.

ymzong:
From what I can observe, now `Go to Course Page` link appears (vertically) at the bottom of its row, instead of at the middle.

Z1MM32M4N:
I don't quite know what you mean, could you post a screenshot?

ymzong:
Sure:

screenshot from 2015-01-18 21 33 09

Z1MM32M4N:
I can't reproduce this. I'm on Chrome on a MacBook Air. I'm testing the change locally (with the actual patch applied) as well as on autolab.cs.cmu.edu (by applying my change from the Inspector. I just pulled.

The only thing I can think is that you're either running an out-of-date version of the code base, have some other local changes that affect this, or are using a different browser. Any other ideas?

Either way, I think that the button being at the bottom is less visually distracting than the cards not wrapping correctly.

icanberk:
Yeah, it happens for me too sometimes. But I think we can fix the smaller bugs later.

PR #134, opened 2015-01-19
Module Updates

dlbucci:
Removed UserModule Key-Value store and added Groups.


dlbucci:
There's too much going on here, so I'm gonna start over.

PR #137, opened 2015-01-20
Use static padding in favor of min-height on cards

### LINKED ISSUE #135: "Go to Course Page" button is sometimes misplaced
Sometimes the "Go to Course Page" button looks like this:

It should be centered in the whitespace at the bottom of the card.

Z1MM32M4N:
Fixes #135.

### DIFF
--- a/app/assets/stylesheets/style.css.scss
+++ b/app/assets/stylesheets/style.css.scss
@@ -481,7 +481,6 @@ div.card {
   margin-right: 20px;
   margin-top: 15px;
   vertical-align: top;
-  padding:0;
   webkit-box-shadow: 0 0 8px rgba(0,0,0,0.10);
   -moz-box-shadow: 0 0 8px rgba(0,0,0,0.10);
   box-shadow: 0 0 8px rgba(0,0,0,0.10);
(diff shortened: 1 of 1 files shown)


Z1MM32M4N:
@ymzong I believe this fixes the issue you were having with card button stylings.

dlbucci:
How does this look with multiple courses?

Z1MM32M4N:
No worse than it does before. The heights are ragged when you have multiple open assignments (as they were before), but at least the buttons don't touch the bottom line.

icanberk:
Actually, why did we remove `min-height: 215px;` ?

Z1MM32M4N:
I removed the `min-height` because I don't think it's actually necessary. Here are the cases:

1. There are no open assessments. In this case, the `min-height` comes into play, and there's about 40px below the "Go to Course Page" button.
2. There is one open assessment. In this case, the relative spacing doesn't change, because the heights just work out that way.
3. There are two open assessments. In this case, the `min-height` doesn't come into play, forcing the button to be flush with the bottom.

After writing out those cases, though, I've convinced myself that keeping in the `min-height` would make cases 1 and 2 have cards of the same height, which probably makes things look best 95% of the time.

I'll add it back in, with the `padding-bottom` unchanged.

Z1MM32M4N:
That being said, there's still some weirdness if the card title is three lines long.

PR #144, opened 2015-01-20
Revert "Adds small margin-top to submit buttons"

Z1MM32M4N:
This reverts commit 2762e78ad21874e925673cee30d1664911814d8a.

I only tested this change on the Change Password form. There are other
forms where this change doesn't work correctly, for example the Edit
Assessment form when the Delete this Assignment button is also present.

The issue I was trying to fix (in one fell swoop rather than looking at
each case separately), was that sometimes the submission button is two
close to the end of the form (i.e., touching the last input). Rather
than pushing the button down uniformly, all the inputs should have
margins after them forcing things that come later to be in a more
appropriate place. I'll look into making these changes, but things look
better with this change reverted.

### DIFF
--- a/app/controllers/course_user_data_controller.rb
+++ b/app/controllers/course_user_data_controller.rb
@@ -1,11 +1,11 @@
 class CourseUserDataController < ApplicationController
-  
+
   action_auth_level :index, :student
   def index
     @requestedUser = @cud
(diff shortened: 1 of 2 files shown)


Z1MM32M4N:
Whoops, accidental PR.

PR #145, opened 2015-01-20
Reveal course profile in navbar. Fixes #139.

### LINKED ISSUE #139: Students have no option to change their nicknames
Currently, the only way a student can change their nickname is by having a TA or Admin change it.

The route is there, but you have to go to Course Admin > Manage Accounts and then click on the email of one of the students > Edit Information. Notice that that whole workflow assumes you're a Course Admin.

If this is by design, then go ahead and close the issue. In the past though, people have always been able to change their nicknames at any time.

Z1MM32M4N:
Adds a dropdown menuitem titled "Course Profile" whenever the navbar is
rendered in the context of a course.

### DIFF
--- a/app/views/layouts/_navbar.html.erb
+++ b/app/views/layouts/_navbar.html.erb
@@ -12,7 +12,7 @@
 
 	<div class="collapse navbar-collapse" id="bs-example-navbar-collapse-1">
 		<ul class="nav navbar-nav navbar-right action-links" role="navigation">
-            
+
 			<% if session[:sudo] && @cud then %>
 			<li>
 				<%= link_to "(#{@cud.display_name}) Reset user",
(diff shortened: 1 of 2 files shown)


app/views/layouts/_navbar.html.erb
```diff
@@ -59,6 +59,11 @@
 						<%= link_to "Account", edit_user_path(current_user), :title => "View your account info" %>
 					</li>
 					<% end %>
+					<% if @course and @cud %>
+					<li role="presentation">
+						<%= link_to "Course Profile", course_course_user_datum_path(@course, @cud), :title => "View your user profile for the current course" %>
```
Z1MM32M4N: The only reason why I figured out how to add this was because I saw it referenced somewhere else in the project, but I couldn't find where this method (`course_course_user_datum_path`) was defined. My hunch is that Rails is auto-generating this method, but I would love clarification!

dlbucci: Run 'rake routes' to get a list of routes in Autolab, as well as helper functions to create those routes, when they are available.  course_course_user_datum_path is indeed one of the paths defined by our route schema.  For what it's worth, you could have also done [@course, @cud] in place of the function call and got the same result.

PR #146, opened 2015-01-21
Fixed 2 submission counting queries

dlbucci:
Fixes two erroneous counting queries. There may be more.

### DIFF
--- a/app/controllers/assessment/handin.rb
+++ b/app/controllers/assessment/handin.rb
@@ -91,7 +91,7 @@ module AssessmentHandin
       @assessment = @course.assessments.find(params[:assessment_id])
     end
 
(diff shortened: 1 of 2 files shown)


ymzong:
Fixes #147?

dlbucci:
@ymzong It should, but I'm also less than confidant that all of our queries are working as intended.

ymzong:
Could you use a simple `grep` to search for similar queries and fix them as
well (if they exist)?

Then I'll merge it to master :)

dlbucci:
I don't think there are any more worrisome queries of this type.

ymzong:
@dlbucci Awesome, thanks!!

PR #153, opened 2015-01-21
Started using rack-ssl-enforcer gem for forcing SSL

icanberk:


icanberk:
This gem in general provides us with more flexibility.

dlbucci:
How well has this been tested?

icanberk:
Pretty well tested on sawshark. Although we might need to change some apache configs after pulling to milkshark.

dlbucci:
We will.  We were using this gem before though?

icanberk:
No we never used it before

dlbucci:
So can we test this gem?

icanberk:
I tested on sawshark.

Issue #126, opened 2015-01-18
AFS Submit Script Broken


namanbharadwaj:
The example script given (`Autolab/script/autolabSubmit.sh`) is broken. In particular, it seems that the `unofficial.fish.ics.cs.cmu.edu` URL can't see "new Autolab" courses and gives me `ERROR: invalid course`.

15-210 uses this for all labs, so getting a fix for this or a new URL that works would be nice.

dlbucci:
The unofficial_submit is now a route for autolab, so I think the problem is that the example submit script needs to be rewritten, but I'm not 100% sure that everything else is working.

ymzong:
We really want to fix the issue asap -- I see it bothering lots of students in 15-122 working on Windows... They should be able to run `handin filea.c0 fileb.c0` on `andrew`, but now they need to:
- make a `tgz` out of the submission files on afs;
- use a transfer client to obtain a local copy;
- manually submit the `tgz` to Autolab.

robsimmons:
The "handin" file is actually /afs/andrew.cmu.edu/course/15/122/bin/handin, which calls  /afs/andrew.cmu.edu/course/15/122/bin/handin/autolabSubmit-master.sh, which follows the following protocol:

Queries: http://unofficial.fish.ics.cs.cmu.edu/officialSubmit.rb?course=15122-s15&user=rjsimmon&assessment=scavhunt
Receives an AFS path in response
Places the handin tarball (handinscavhunt.tgz) in that directory
Queries: http://unofficial.fish.ics.cs.cmu.edu/officialSubmit.rb?course=15122-s15&user=rjsimmon&assessment=scavhunt&submit=handinscavhunt.tgz

I'm happy to hack my script, but I just don't know a lot of things, starting with what my source number is: "15122-s15" and "14" both give me "invalid course".

dlbucci:
This can't be resolved until we fix our SSL setup: unofficial submit needs to use HTTP to get around the user needing HTTPS certificates, but our current Apache setup always forces HTTPS.  We should be switching to Rails-enforced SSL instead of Apache, which will let Rails not force SSL for some routes.  Then we can rewrite this script.

icanberk:
Hi @robsimmons, I'm working on fixing this as soon as possible. The unofficial route is working now, but the url is different. It's the part of the app now, so it can be reached through:

[...]

PR #163, opened 2015-01-22
Fixed and cleaned up methods for log and local submissions

icanberk:


icanberk:
This should not be merged yet, but I wanted to push it to get your comments.

icanberk:
This is up for review now

dlbucci:
How well has this been tested? We seem to still be getting error reports.

icanberk:
I think those errors were from while I was developing.

dlbucci:
The ones mentioned at the end of #126?

icanberk:
Gonna work on those soon.

dlbucci:
Alright, well I don't think we're any worse off with this in then, so I'll merge.

Issue #69, opened 2015-01-14
DK displaying gradebook


droh:
Seth uploaded the roster and then visited the gradebook, which DK'd attempting to run the courseAverage function. I fixed it by manually reloading the course.rb file, but the fact that I had to do it is a bug.

ymzong:
I cannot reproduce the bug in Seth's course after loading a user with a dummy roster -- can someone try reproduce it and post the error message?

dlbucci:
@droh was this fixed when the permissions on the course directory were fixed?

droh:
No, his permission bits were fine. This was fixed when I reloaded the
course.rb file, but it shouldn't have happened. Somehow, we want to ensure
that the course.rb file is always loaded when a course is created.

-- David O'Hallaron
-- Prof of CS and ECE, Carnegie Mellon University
-- GHC 7517, x8-8199
-- http://www.cs.cmu.edu/~droh

andrewwuan:
@droh Just to make sure, the reload should be done after "a course a created" rather than someone "upload a roster" right?

droh:
Yes, course.rb reload should be done after course created.

On Monday, January 19, 2015, An Wu <notifications@github.com> wrote:

> @droh <https://github.com/droh> Just to make sure, the reload should be
> done after "a course a created" rather than someone "upload a roster" right?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/autolab/Autolab/issues/69#issuecomment-70574660>.
>


-- 
Sent from my iPhone.

PR #175, opened 2015-01-24
Reload course config after course is created

andrewwuan:
https://github.com/autolab/Autolab/issues/69


dlbucci:
That message... ( ͡° ͜ʖ ͡°)

dlbucci:
Also, do we have any other examples of using class methods in Controllers like this?  The function seems like it would be more appropriate in the course.rb file.

PR #177, opened 2015-01-24
move reload_course_config to course model

andrewwuan:
@dlbucci https://github.com/autolab/Autolab/pull/175#issuecomment-71333094 This makes sense. I wasn't sure because extend is used in the method.
@icanberk Can you make sure that AdminsController.extend is what we want?


dlbucci:
AdminsController.extend should add the methods in mod to the AdminsController, which is what was happening before, so this should work fine.

PR #182, opened 2015-01-25
Set shibboleth attribetus when user registers

andrewwuan:


app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -95,5 +99,19 @@ def shibboleth
...
     end
   end
+
+  private
+
+  # choose_shib_attr
+  # Shibboleth may return semi-column seperated attributes.
+  # The function returns the first one if attributes_str is not nil, otherwise
+  # it returns "(blank)" as a placeholder
+  def choose_shib_attr(attributes_str)
```
andrewwuan: Since shibboleth attributes are unordered, we can only choose one (here it's the first) hoping it makes sense.
Choice of attribuets like school, department, or year can be improved by having a list of them in advance and compare, but I'll do it in another PR.

andrewwuan:
This is tested with my own CMU account

app/controllers/users/omniauth_callbacks_controller.rb
dlbucci: Does this mean some attributes come in as lists?  Like if a person has multiple majors or whatnot?

andrewwuan: Yep double major folks

PR #190, opened 2015-01-26
Fix official+unofficial submit

icanberk:


icanberk:
Fixes issue #126

app/models/submission.rb
```diff
@@ -91,6 +91,8 @@ def saveFile(upload)
       # Sanity!
       upload['file'].rewind
       File.open(path,"wb") { |f| f.write(upload['file'].read)}
+    elsif upload['local_submit_file']
+      File.open(path,"wb") { |f| f.write(IO.read(upload['local_submit_file']))}
```
dlbucci: Why are there two different methods for reading the file here?

app/controllers/assessments_controller.rb
```diff
@@ -1299,10 +1299,16 @@ def local_submit
         @cud = CourseUserDatum.find_cud_for_course(@course, @user.id)
         @submission = Submission.create(:assessment_id=>@assessment.id,
           :course_user_datum_id => @cud.id);
-        @submission.filename = File.join(internalDir, handinFile)
-        @submission.save!
+        upload = {'local_submit_file'=>File.join(internalDir, handinFile)}
 
-        afterHandin(@submission)
+        @submission.saveFile(upload)
+        #afterHandin(@submission)
```
dlbucci: Dead code here.

app/models/submission.rb
icanberk: upload['file'] is a class that we're not using in the case of local submit. so there is not way to do `.read`

dlbucci:
If we're gonna say this fixes #126, this should fix the scripts/autolabSubmit.sh too.

icanberk:
Is 213 using that script? Otherwise, can we just get rid of it?

app/models/submission.rb
dlbucci: So upload['file'] is an actually file-like object, but upload['local_submit_file'] is a string?  Could we get a comment for that.

dlbucci:
122 and 210 are using that script for command line submission.  We should have an official local submit script.

namanbharadwaj:
We (210) actually have our own script that follows the protocol, so as long as the protocol is documented, getting rid of the official script is perfectly fine. I believe that 122 does a similar thing (@robsimmons).

PR #191, opened 2015-01-26
Up the feedback column limit

dlbucci:
Should help with the feedback issues. Ups the size from 64KB to 16MB
Will need a db migration that won't be reversible.


dlbucci:
This is gonna be painful because of Ilter's remote_handin_path migration already being run, so I'll just make a new branch.

Issue #193, opened 2015-01-27
Handin script enhancements


robsimmons:
15-122 Handin is working again!

1) Without assignment names, this script is all incredibly brittle from semester to semester - I have to edit the script to have the right assignment ID, rather than being able to leave it as "scavhunt" from semester to semester. Having a course id of 14 instead of 15122-s15 isn't as big of a deal.
2) Not DKing when the handin path is wrong/not set would be nice
3) Having remote file path entirely in assignment configuration, instead of part of the course configuration, doesn't make much sense to me. A course-specific prefix and an optional assignment-specific postfix (which isn't needed for the `/afs/andrew.cmu.edu/scs/cs/autolabEmail/handin/` path that 122 uses now). I guess maybe this is one of these things like autograde_image that I set once and forget... I dunno. I'd just ask that you rethink this
4) If you're going to have the remote file path in course-wide configuration, remote_path should be exported in the yml file like autograde_image

remzr7:
Backlogging

PR #194, opened 2015-01-27
Added an error message for case where the directory is not defined

icanberk:


icanberk:
Addresses point 2 in issue #193

Issue #181, opened 2015-01-25
Upgrade Datetime Pickers


jez:
### Required Features
- [x] Linked datepickers (start time can never come after end time)
- [x] Cross browser support
- [x] Option to cut/copy/paste dates in plain text
- [x] Upgrade all current interfaces which use dropdowns
- [x] Include a "now" or "today" button
- [x] Easy to grok times (from a user's perspective)
- [x] Allow arbitrary precision for picking dates (to the minute)

This will likely be implemented using [Bootstrap 3 Datepicker][1], which includes support for all of these date and time options.

[1]: http://eonasdan.github.io/bootstrap-datetimepicker/

icanberk:
@Z1MM32M4N how's this going?

jez:
I've been stuck on how to get this to work with Rail's custom form builders. I'd like to be able to just write `f.datetime` and have it include all the necessary HTML. For whatever reason, when I try to modify the existing `FormBuilderWithDateInput` class, it raises a method not found exception.

dlbucci:
Good idea. I think changing FormBuilders requires a server restart, so you know.

Z1MM32M4N:
@dlbucci Rather, when I try to add the field `builder: FormBuilderWithDateInput` to the appropriate form (without changing the form builder) is when I get errors.

dlbucci:
What's the method it's not finding?  Are you modeling the form after the assessments#edit form (which might be the only place we use that builder)?

Z1MM32M4N:
pm'ed you with a lengthy explanation.

Z1MM32M4N:
So turns out that a server restart did the trick (as referenced in the above pull request!).

Z1MM32M4N:
The next issue is more of a design decision, so I'm wondering what people's thoughts are.

There's a project that serves the assets through the [Rails asset pipeline][1], but it's [out of date][2] right now (1 version behind).

We could very easily download the required assets (Bootstrap 3 DatePicker + Moment.js) and manage installation and updating manually, or we could load the assets through our Gemfile. I'm in favor of the second, but there's no guarantee that this will happen any time soon.

Thoughts? Our options are:

1. Manage all static assets manually
1. Download manually now and patch up Gemfile later
1. Don't do extra work now and install through Gemfile when it becomes available

[1]: https://github.com/TrevorS/bootstrap3-datetimepicker-rails
[2]: https://github.com/TrevorS/bootstrap3-datetimepicker-rails/issues/23

[...]

Issue #101, opened 2015-01-15
Inconsistent font sizes throughout the app


droh:
Throught the new system, the font sizes of different elements are very inconsistent. It would be very helpful for someone to a global review of the consistency of the fonts throughout the system.

icanberk:
Merged to #231

PR #199, opened 2015-01-27
Restyled Edit Problem page using FormBuilder

Z1MM32M4N:
Makes progress towards #101 (inconsistent font sizes) and #181 (practice using
FormBuilders!).

### DIFF
--- a/app/assets/stylesheets/problems.css.scss
+++ b/app/assets/stylesheets/problems.css.scss
@@ -1,0 +1,7 @@
+form.edit_problem {
+  max-width: 600px;
+
+  div.action_buttons {
+    margin: 25px 0 15px 0;
+  }
+}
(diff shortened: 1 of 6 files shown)


dlbucci:
This only works on the edit form and not on the new form.  Also, a breadcrumb back to problems editing page from the new/edit pages would be a nice feature, though not a requirement.

Z1MM32M4N:
878fe56 should fix the New Problem styles, and 6cca38b should add breadcrumbs.

dlbucci:
Could you also de-tabify the _navbar layout? Other than that, this looks good.

Z1MM32M4N:
Do you mean the Edit Assessment tabs (basic, handin, ...), like into a single page?

dlbucci:
No, I mean the _navbar layout has tabs instead of 2 spaces, which is why it looks all wonky.

Z1MM32M4N:
Oh, whoops, I didn't even check the diff. Good catch; it should be fixed now.

Issue #202, opened 2015-01-27
Grade submissions page: feedback box no longer shows Grader ID


droh:
The feedback box should show the grader's email address, so that students know who is giving them their feedback. This was working during the beta test last semester, but somehow disappeared in the current version.

dlbucci:
This will be fixed by #205

PR #205, opened 2015-01-27
Fixed Grader Info in Submission Popover

### LINKED ISSUE #201: Grade Submissions page: Adding feedback to a score sometime fails
When trying to add feedback to a score, the feedback box pops up and then DKs inside the box the first time, and thereafter, clicking the score doesn't pop up any feedback box. This doesn't occur for every course or assessment. To consistently reproduce this, visit 18447-s15, lab1,a*****r@andrew, and try to add feedback to his score (see screenshot below).

The problem only shows up for certain assessments, even in the same course. For example, 15122-s15,Written 1 is fine for all students and all problems. However, for 15122-s15,Quiz 0, it fails for all students. 

5640038c-a63e-11e4-905a-a355aa6152c2

dlbucci:
Fixes #201 and #202.

### DIFF
--- a/app/controllers/assessment/grading.rb
+++ b/app/controllers/assessment/grading.rb
@@ -286,8 +286,9 @@ public
   end
 
   def score_grader_info
-    grader = Score.find(params[:id]).grader
-    if grader
+    score = Score.find(params[:score_id])
+    grader = (if score then score.grader else nil end)
+    if grader then
       render :text => "#{grader.first_name} #{grader.last_name} (#{grader.email})"
     else
       render :nothing => true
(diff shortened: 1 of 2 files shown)


app/controllers/assessment/grading.rb
```diff
@@ -286,8 +286,9 @@ def submission_popover
   end
 
   def score_grader_info
-    grader = Score.find(params[:id]).grader
-    if grader
```
icanberk: who wrote this? how did this ever work without `then`?

ymzong: @icanberk Apparently the new Rails overrode `params[:id]` with the id of the controller, i.e. assessment.

dlbucci: @icanberk ruby if statements don't need to have a then unless the block is on the same line (if true then blah end), but I think it's good practice to have then's at the end of if statements.

Issue #64, opened 2015-01-13
Sign up is Confusing


icanberk:
After signing up, the system should print a message directing users to
visit their inbox and look for the confirmation email

PR #216, opened 2015-01-29
Include all the flash messages to layout

icanberk:
This fixes Issue #64

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -57,15 +57,12 @@
         <% end %>
 
         <!-- Flashes -->
-      <% if flash[:error] %>
-        <p class="error"><%= flash[:error].html_safe %></p>
-      <% end %>
-      <% if flash[:success] %>
-        <p class="success"><%=flash[:success].html_safe%></p>
-      <% end %>
-      <% if flash[:info] %>
-        <p><%= flash[:info] %></p>
-      <% end %>
+        <div id="flashes">
+          <% flash.each do |name, msg| %>
+          <%= content_tag :div, sanitize(msg), id: "flash_#{name}",
+              class: "#{name}" %>
+          <% end %>
+        </div>
         <!-- End Flashes -->
 
         <!-- Persistent Announcements -->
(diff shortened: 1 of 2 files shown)


app/views/layouts/application.html.erb
```diff
@@ -66,6 +66,12 @@
       <% if flash[:info] %>
         <p><%= flash[:info] %></p>
       <% end %>
```
dlbucci: Do we still need this flash info?

icanberk: I'm not actually sure. Didn't remove them just in case..

app/views/layouts/application.html.erb
```diff
@@ -66,6 +66,12 @@
...
+        <div id="flashes">
+          <% flash.each do |name, msg| %>
+          <%= content_tag :div, raw(msg), id: "flash_#{name}",
```
dlbucci: Why does msg need to be raw?

icanberk: Yeah, I think devise passes some html message occasionally that we don't want to escape.

dlbucci: I'm wary of this at the moment as I just had a big todo about not letting scoreboards be raw.  Can we at least use 'sanitize' (http://api.rubyonrails.org/classes/ActionView/Helpers/SanitizeHelper.html#method-i-sanitize) instead? That will at least strip out script tags.

icanberk: Done

app/views/layouts/home.html.erb
```diff
@@ -29,7 +29,7 @@
       <div id="pageBody" class="container">
 
         <!-- Flashes -->
-        <div id='flashes'>
+        <div id="flashes">
           <% flash.each do |name, msg| %>
           <%= content_tag :div, raw(msg), id: "flash_#{name}",
```
dlbucci: Sanitize here, while we're at it. (Actually, you could also make a partial template for this, but you don't need to)

icanberk: Done

PR #219, opened 2015-01-30
Turned production config to a template and added to gitignore

icanberk:


icanberk:
The idea is to put all the private keys etc. into production configs and keeping it away from the repo.

After people install the dev version, they'll need to fill the `production.template.rb` to start running the prod environment.

PR #220, opened 2015-01-30
Ground Work for SVN

dlbucci:
Take a look at this and see if this will work.  Setting repositories and all that will work fine.  It should be up to the instructors to override the 'checkoutWork' method of the Svn module, and everything should work.  I'm not certain that the module is even still needed actually, but I'm not totally sure.


icanberk:
This looks good to me. @andrewwuan should take another look before merging.

andrewwuan:
@dlbucci Have you tested this? I'm more concerned with how the user access the svn handin page because we changed the mechanism of handing in...

dlbucci:
The SVN handin page is now loaded as a partial on assessments#show, just like the normal handin page.  All of the links and functions work now, the main question is whether instructors can still override 'checkoutWork', which I believe they can, but am not 100% sure about.

andrewwuan:
@icanberk Does checkoutWork also needs to be clean up with module?

icanberk:
Yes if we'd like the instructor to add the method in assessment config. I
actually just created a helper method for that. I can fix the checkoutWork
part as well before pushing it.

dlbucci:
Actually, that module could probably be emptied, seeing as both of those functions need to be overwritten for SVN to work anyway.  We probably still want to leave it their in case labs are requiring Svn.rb.  Then again, if we removed it, we'd get errors, but professors would fix their files...

icanberk:
@dlbucci let's not do that until we have a good way of warning the instructors.

PR #221, opened 2015-01-30
Cleanup autograde module and check for overwriting autogradeDone

icanberk:
Overwriting autogradeDone is important because Buflab is using this functionality(in a very hack-y way)


dlbucci:
If the Autograde module is empty, is there a way we can notify professors of this when its required in an assessment file?  Like, print a message saying "require Autograde does nothing, so please remove it"?

Also, I'm seeing the config_module check in more than a few places.  Do we want to make a helper method for it?

icanberk:
It's a good idea. Should we show it on assessment page? Should it stay there until the instructor deletes it?

And yeah, a helper method would actually be useful.

dlbucci:
Yeah, just have it show up like an announcement or flash message.  And having it stay there would be good.  In fact, we might be able to extend Announcements into a DeprecationWarning model that only displays for Instructors and lets them know when their code is out of date.  That could be a nice feature for the devs.

PR #223, opened 2015-01-31
Fix various Manage Accounts bugs

### LINKED ISSUE #207: Useful sequence numbers on Manage Accounts page
Would it be possible for the "No." column at the left to __not__ resort when we resort by other columns?

### LINKED ISSUE #212: Manage accounts: Search on email, first, and last
On the Manage Accounts page, the search operation is only looking at the email address. It should include First and Last name as well.

Z1MM32M4N:
### Commit Summary

ac2c179 - Set row index counter using CSS

This way, we don't have to worry about reindexing when we filter rows,
because the browser will take care of it.

Fixes #207

edc00d9 - Clean up SCSS for tables

3d49510 - Search case insensitively in Manage Accounts.

Fixes #212

### DIFF
--- a/app/assets/stylesheets/style.css.scss
+++ b/app/assets/stylesheets/style.css.scss
@@ -431,37 +431,43 @@ div.field_with_errors {
 
 /* TABLE STYLES.  Styles for the multiple tables that we have for some reason */
 /* Pretty Border Tables. I vote we make this the generic table style. */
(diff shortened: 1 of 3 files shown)


Z1MM32M4N:
It might be more manageable to look at the diffs for the individual commits. They're grouped according to the change they introduce. Overall, there are three "big" changes.

icanberk:
Didn't know about these CSS row numbering properties. Nice.

Issue #225, opened 2015-01-31
Standardize Static Assets


jez:
The static asset discipline is a little unnerving. Here's a short list of things I'd like to see improved:
- [x] One (1) version of jQuery, ideally jQuery 2.x, used throughout the app

Right now, we are loading both jQuery 1 and 2. One is being loaded over a CDN, and I think another is being loaded through the asset pipeline. Which leads me into
- [ ] better usage of Rails' static asset pipeline

If used correctly, it can ease the efforts of minification as well as managing and upgrading dependencies. I don't think that many plans for maintenance or upgrading have been made with respect to this area, so I'd like to make sure that we have
- [x] Up-to-date versions of static assets, to include
  - [x] Bootstrap
  - [x] jQuery

Later, this list will include whatever library we choose for displaying tables (#224), as well as date/time pickers (#181). Keeping assets up to date is like taking your vitamins: you do it so you don't get sick down the road.

This is a pretty broad-sweeping issue, which means it'll be hard to make measured progress towards it. However, I still think it's a good idea to get these ideas out into the open so we can make a good plan going forwards.

icanberk:
I think jQuery stuff is pretty standardized at this point but you should take another look and update this issue.

dlbucci:
It would be nice if #316 was fixed, though.

dlbucci:
What remains to be done for the last check?  If it's substantial, can we make a new issue and close this one?

jez:
I'm working on this for the process of #345. Because this issue is inspiring my solution for that one, we may as well keep it open, but we can close it if you want.

icanberk:
Any progress on this?

jez:
Yes, I have a branch for this and have started working on it, but I won't be able to work on anything related to Autolab until at least Tuesday.

remzr7:
Standardized static assets with recent pull requests

PR #226, opened 2015-01-31
Initial Work on Datetime Pickers

Z1MM32M4N:
### Commit Summary

c279e97 - Purge all date & time assets

76b10b8 - [bundle install] Add date time dependencies

You will need to run

    bundle install

after merging this commit.

76c1e75 - Stripped whitespace from application_helper.rb

b400e3d - Remove duplicate version of jQuery (#225)

c9249e5 - Use Bootstrap grid over manual width styling

Also edited create_edit.css.scss for SCSS style guidelines.

38c7733 - Incorporate Bootstrap 3 Datetimepicker (#181)

Functionality Added

 - The custom form builder `FormBuilderWithDateTimeIntpu` adds support
   for date selectors to be added with `f.date_select` and
   `f.datetime_select`.
 - The Edit Assessment > Handin page was refactored to use the above
   functionality
 - The `initialize_datetimepickers.js` file contains the logic needed
   to initialize any datetime picker on any page. It still needs to be
   included whenever datetimepickers will be on the page.

Notable Additions

 - Removes commented out code relating to Waves.js
 - Loads moment.js and bootstrap-datetimepicker.{css,js} on all pages
   the using application.{css,js} files.

7e3eb77 - Patch styles for Bootstrap 3 Datepicker (#181)


Z1MM32M4N:
These changes have been tested to work for the Edit Assessment > Handin page.

Currently, the only line items of #181 that this doesn't resolve are linked date pickers (which is slightly more complicated because we have four fields which need to be arranged chronologically, not two) upgrading all date pickers (the framework is there, but not the changes nor testing).

icanberk:
This is cool.

Are the icons at the bottom of the picker coming from the library? They will probably be a little confusing to the users. Can we add "Current Date" and "Set Time"/"Set Date" texts next to them?

icanberk:
@Z1MM32M4N if it's going to require too much work, we can leave it for the later versions. Just let me know. I really want to close the issue this weekend :)

Z1MM32M4N:
I might be able to hack it together. I was pouring over the source to
figure out what classes to style, so I have a few ideas. Certainly I could
do it by forking the project, but I might be able to get it with even less
work.

icanberk:
Forking the project is probably unnecessarily too much work at this point.

Maybe you can add content to an :after class...

Z1MM32M4N:
Yep, that's exactly what I had in mind.

Z1MM32M4N:
This should be good to go.

PR #233, opened 2015-02-01
Remove libarchive from submissions#view

dlbucci:


app/models/submission.rb
```diff
@@ -205,12 +205,44 @@ def get_scores(released_only, problem_id_to_name)
...
-    require "libarchive"
-    archive = Archive.read_open_filename self.handin_file_path
-    while header = archive.next_header do
-      return header.pathname if archive.header_position == position 
+    require 'rubygems'
+    require 'rubygems/package'
+    require 'zlib'
+    require 'zip'
+
+    archive_type = IO.popen(["file", "--brief", "--mime-type", @filename],
```
ymzong: @dlbucci The code is breaking at Line 213 `@filename`. Apparently this variable is not defined in this context...

dlbucci: I'm guessing that should be submission.filename.  This was pretty much copied from the getFileAt function in the submissions_controller.

PR #234, opened 2015-02-02
Made autograde module be importable from labs + cleaned up logs

icanberk:
(ie. buflab can call sendJob now)

### DIFF
--- a/app/controllers/assessment/autograde.rb
+++ b/app/controllers/assessment/autograde.rb
@@ -1,5 +1,5 @@
 module AssessmentAutograde
-  require 'autoConfig'
+  require 'autoConfig.rb'
 
   # 
   # autogradeAfterHandin - submits an autograding job to Tango when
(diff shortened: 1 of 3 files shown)


app/controllers/assessment/autograde.rb
```diff
@@ -228,4 +235,83 @@ def autograde?(submission)
     not (submission.nil? or submission.handinFile().nil?)
   end
 
+  def extend_config_module
```
icanberk: This code is repeated in submissions_controller and assessments_controller but I haven't had the chance to clean it up yet. I think it'll work if we just remove that method from but no time to test.

dlbucci:
Should we really be allowing this?  It seems like it's a step backwards in our attempt to remove modules...

icanberk:
I'm not sure if it's a step backwards. We just allow custom rb files send jobs to Tango. Although I think overwriting autogradeDone is pretty bad. Anyhow, we really need this for buflab coming out on Thursday...

icanberk:
We need to rewrite buflab with a better logic in the future, but Dave didn't want to take the risk at this point.

dlbucci:
Yeah, the fact that we need to rewrite it makes me think this is a step back, or at least a side step, but I understand.

Issue #43, opened 2014-11-19
Database Schema should be Cleaned Up


dlbucci:
Before spring 15, we should remove columns and tables we are not using.

PR #242, opened 2015-02-03
Removed _old lines from schema

icanberk:


icanberk:
@dlbucci how do I test this?

icanberk:
Resolves #43

icanberk:
So I tested this by creating a db, migrating it and populating it. Everything seems to work well and we don't have the _old columns in the db anymore. I think this is good to go.

db/schema.rb
```diff
@@ -148,7 +144,6 @@
     t.boolean  "dropped",              limit: 1,   default: false
     t.string   "nickname",             limit: 255
     t.boolean  "course_assistant",     limit: 1,   default: false
-    t.float    "tweak_old",            limit: 24,  default: 0.0
     t.boolean  "absolute_tweak",       limit: 1,   default: true
```
dlbucci: I believe this (absolute_tweak) can also be removed.

dlbucci:
Isn't there a migration file for this?  I don't think just changing the schema will affect the database.

icanberk:
Hmm.. this actually seems to be the wrong practice.
http://stackoverflow.com/questions/11301009/is-it-considered-safe-to-manually-edit-schema-rb-in-rails

I will just migrations that will remove these keys.

(Rails is so weird)

icanberk:
@dlbucci this should be good to go now

icanberk:
Fyi, schema.rb is autogenerated. I'm not sure why it diffed a lot of those lines.

dlbucci:
The _backup columns in CUD can also be removed.  There's another absolute_tweak column that I believe can also be removed.

icanberk:
Are you sure we are not using submission.absolute_tweak  ?

dlbucci:
And the feedback_file columns in the scores table can also be removed, since we don't really use that anymore, although maybe check that we aren't.

dlbucci:
I think tweaks have been moved into the Tweak model, so the column is no longer used, but you can check to be sure.

icanberk:
I see. It doesn't seem to be used anywhere.

`feedback_file` seems to be in use btw.

dlbucci:
Also, absolute_tweak is only mentioned in the schema and old migrations.

icanberk:
Done

dlbucci:
Code is there to handle feedback_file values, but I don't think that column can be set anywhere in the code.  If you want, we can leave it in for now, and remove it after a check.

icanberk:
Yeah, let's not touch it for now.

dlbucci:
And you've made sure this migration will run with the new additions?  Because we absolutely do not want a migration to fail in production.

icanberk:
Yep

PR #252, opened 2015-02-05
Temp Fix for some encoding errors

dlbucci:
I call this temporary because there are still some issues displaying
UTF-8 characters and, overall it just feels hacky.  That said,
submissions with UTF-8 characters won't break the page now, which is
progress on #251.

I'd suggest a more thorough examination of our current encoding
situation in the future.

### DIFF
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -137,7 +137,7 @@ class Submission < ActiveRecord::Base
     conditions[:position] = position if position
     annotations = self.annotations.where(conditions)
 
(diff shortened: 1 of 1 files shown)


icanberk:
@dlbucci next time can you name the fixes that will go into master as "hotfix_"

dlbucci:
Never.  Oh wait, sure.

icanberk:
This is pulled to prod. You should also merge this into `develop` now.

RONNCC:
Can we open another issue to make this less hacky later? :P

PR #253, opened 2015-02-05
Expose linked relationships among datetimepickers

jez:
Makes progress towards #181.

To use, simply enumerate the IDs of datetimepickers which should be
forced to be less than a particular `f.datetime_select` in the `:less_than`
field and those which should be greater in the `:greater_than` field.

The `initialize_datepickers.js` file must be manually included on every
page that uses datetimepickers. We could easily make this a file that is
loaded by default for every page if that's preferable.


icanberk:
"merge 13 commits into master from feature_datetime_validation"

This should be to `develop`, right?

icanberk:
Nvm

jez:
One step ahead of you :P

PR #254, opened 2015-02-05
Expose linked relationships among datetimepickers

jez:
Makes progress towards #181.

To use, simply enumerate the IDs of datetimepickers which should be
forced to be less than in the `:less_than` field on the FormBuilder, and
those which should be greater in the `:greater_than` field.

The `initialize_datepickers.js` file must be manually included on every
page that uses datetimepickers. We could easily make this a file that is
loaded by default for every page if that's preferable.


icanberk:
This is cool :+1:

PR #255, opened 2015-02-05
Update datetime picker in Edit Course. Closes #181

### LINKED ISSUE #181: Upgrade Datetime Pickers

jez:
I'm pretty sure that this is all the datetime pickers. Obviously, if one
comes up that I've missed, let me know.

Also, as a part of the process of fixing the datetime pickers, I've been
restyling forms to look more like native Bootstrap. At some point, I
hope that we can just use the `FormBuilderWithDateTimeInput` as our
default form initializer, as it contains a bunch of convenience methods
for adding help text and similar things.

### DIFF
--- a/app/form_builders/form_builder_with_date_time_input.rb
+++ b/app/form_builders/form_builder_with_date_time_input.rb
@@ -10,7 +10,7 @@ class FormBuilderWithDateTimeInput < ActionView::Helpers::FormBuilder
       options[:class] = "form-control #{options[:class]}"
       field = super name, *(args + [ options ])
 
-      wrap_field name, field, options[:help_text]
+      wrap_field name, field, options[:help_text], options[:display_name]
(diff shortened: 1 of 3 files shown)


jez:
If I'm not mistaken, this should finish up #181.

PR #263, opened 2015-02-08
Fixed Unrecognized Archive Type errors

### LINKED ISSUE #222: 'libarchive' uses need to be removed
I'm pretty sure we use a different library at this point, but 'libarchive' is required in a few places.  (submissions#view for a tar file is one place)

dlbucci:
Fixes #222, I suppose.


icanberk:
Please merge to `develop` as well

PR #265, opened 2015-02-09
Feature groups

icanberk:


app/controllers/assessment/autograde.rb
```diff
@@ -29,30 +24,31 @@ def autogradeAfterHandin(submission)
...
   # Tango server via an REST API.
   #
-  # Required global input variables:
-  # @course, @assessment, @submission
+  # submissions must have at least one element
   #
-  def sendJob
-    extend_config_module()
-    assessmentDir = File.join(AUTOCONFIG_COURSE_DIR,@course.name,@assessment.name)
+  def sendJob(course, assessment, submissions)
```
icanberk: Why are we passing multiple submissions here but only using submissions[0] throughout?

app/controllers/groups_controller.rb
```diff
@@ -0,0 +1,150 @@
...
+
+  action_auth_level :show, :student
+  def show
+    @aud = @assessment.aud_for @cud.id
+    if !@cud.instructor? then
+      puts @aud.group_id, params[:id]
+      if @aud.group_id == nil then
+        redirect_to action: :new and return
+      elsif @aud.group_id != params[:id].to_i then
+        puts "this some old bull shit"
```
ymzong: Wait... what? :trollface:

dlbucci: Professionalism, that's what.

app/controllers/assessment/autograde.rb
dlbucci: So for groups, we only want to submit one job to be autograded, because, at least with compilers, jobs can take a really long time and there's no need to grade the same submission twice.  But we still need to pick a submission to be the canonical submission that actually gets autograded, so we just go with the first one, since it's guaranteed to be there.

dlbucci: Also, on line 119, we save the dave for each submission in the group.  This lets us know which submissions to save the autoresults for in the autograde_done callback.

dlbucci:
At this point, I think the CRUD design is about where I want it to be, it's just about adding the proper links and whatnot.  Then we'll need to test the autograding changes and this should be good.

app/controllers/assessment/handin.rb
```diff
@@ -74,19 +76,69 @@ def validateHandin()
...
+    return validateForGroups()
   end
 
-  def saveHandin()
-    @submission = Submission.create(:assessment_id => @assessment.id,
-                                    :course_user_datum_id => @cud.id,
-                                    :submitter_ip => request.remote_ip)
-    @submission.saveFile(params[:submission])
-    return @submission
+  def validateForGroups
```
icanberk: Can we comment what this function does?

app/controllers/assessments_controller.rb
```diff
@@ -2000,78 +1970,6 @@ def get_assessment
     end
   end
 
-  def extend_config_module
```
icanberk: I'm all for making this extend_config_module DRY, but I'm not sure if it's appropriate to put it in autograde.rb

What it essentially does is to let instructors overwrite assessment configs and some of the methods that get overwritten are not really related to autograding. (e.g. there are methods for overwriting scoreboard headers) I guess this works for now since only submissions_controller and assessments_controller uses this method and autograde.rb is included in both, but I'd suggest moving it into something like application helpers. What do you think?

app/controllers/groups_controller.rb
```diff
@@ -0,0 +1,214 @@
...
+        aud.membership_status = AssessmentUserDatum::UNCONFIRMED
+        aud.save!
+      end
+      @group.destroy!
+    end
+    respond_with(@course, @assessment, @group)
+  end
+ 
+  action_auth_level :import, :instructor
+  def import
```
icanberk: How are we planning to export/import these groups? What's the use case you had in mind?

app/controllers/submissions_controller.rb
```diff
@@ -1,8 +1,5 @@
 class SubmissionsController < ApplicationController
 
-  autolabRequire File.join(Rails.root, 'app/controllers/assessment/autograde.rb')
```
icanberk: how does this work if we're not including autograde.rb?

app/controllers/assessment/handin.rb
dlbucci: Sure thing.

app/controllers/assessments_controller.rb
dlbucci: It was already in autograde.rb though.  And assessments_controller.rb.  And submissions_controller.rb.  And it was the same code each time.  Helpers are really meant for views, not general use.  Also, didn't we deprecate the scoreboard headers thing?

app/controllers/groups_controller.rb
dlbucci: That's a feature that was previously available, and I guess this function will just try its best to make the groups look the same.

app/controllers/submissions_controller.rb
dlbucci: I moved all of the autograding code out of submissions_controller.rb, so there's no need for it to be imported. (There aren't any functions in autograde that this calls, as far as I know).

icanberk: My bad. I knew about this. I really like this idea.

dlbucci:
Okay, so now, I'm hoping I don't have to change any front end MVC stuff.  The only thing left is testing the MVC, autograding tests, and removing the partners module in lib/module.

icanberk:
What's left with this? Anything I can help with?

dlbucci:
I'm just gonna update local_submit and log_submit, and remove partners and UserModule files.  Testing will help, but that might naturally come when we put it in development.

dlbucci:
I'm gonna call this alpha-ready.  People should pull and test, because Autograding is sort of essential to our service.

icanberk:
Looks to me like things are working. Shall we merge this into develop and keep testing on staging?

icanberk:
Also, it'd be nice to squash the last 5 commits.

dlbucci:
I'll do that.  And if we get this into develop, more eyes will be on it, so we can pick up bugs more easily.

icanberk:
Sounds g.

dlbucci:
Well that didn't work at all.

dlbucci:
My attempt to squash did not work too well, and I once again learned to never git push -f

jez:
This doesn't help now, but I was reading an interesting article about how to avoid some of these issues with merging long running branches: it's a pretty good read. 

https://medium.com/@porteneuve/fix-conflicts-only-once-with-git-rerere-7d116b2cec67

PR #267, opened 2015-02-09
Fixed CSV and Student Gradebook Ordering

### LINKED ISSUE #249: Assignment sort
Assignments within a category are sorted correctly (by, I think, assignment release date? whatever, by one of the the dates) in the instructor gradebook, on the main page, etc.

In the CSV gradebook export and the student gradebook, they're sorted by what appears to be sequence ID number, which is not what we want. (See the relationship between quizzes 1, 2, and 3 in 15-122 for an example.)

dlbucci:
Fixes #249


icanberk:
Please merge in to `develop`

Issue #269, opened 2015-02-10
Cumbersome manage accounts navigation


droh:
If an instructor modifies some attributes of some user from the manage accounts page, it's very cumbersome to get back to the manage accounts to modify the attributes of the next user, primarily because the breadcrumb at the top of the page doesn't include "Manage accounts".

PR #273, opened 2015-02-10
Set up breadcrumbs array

dlbucci:
I'm gonna make a PR for discussion.  This is a suggestion for how breadcrumbs could be implemented to be more powerful.  We'd need to update some controllers (submissions_controller) to build the @breadcrumbs array, but we can make it as deep as we want to, and not need to rely on a bunch of instance variables to build it.
It makes progress towards #269.
DO NOT MERGE.


app/controllers/assessments_controller.rb
```diff
@@ -1982,6 +1982,7 @@ def get_props
   def get_assessment
     @assessment = Assessment.find params[:assessment_id] || params[:id]
     @course = @assessment.course
+    @breadcrumbs << (view_context.current_assessment_link)
```
icanberk: What does `view_context.current_assessment_link` do?

dlbucci: view_context lets you use view helper functions (like link_to, form_for, etc.) inside of controller code.  This is calling the 'current_assessment_link' helper method.

icanberk:
I like the idea of powerful breadcrumbs but I still don't see how this works. Can you elaborate more?

For e.g. how to we pass custom titles to `@breadcrumbs`?

dlbucci:
If you want to add a breadcrumb, you append a string or HTML to the end of the array.  Then every item in the array gets added to a span.item in the breadcrumb bar, in the order the items were appended.

icanberk:
I understand the rendering part.

So if I'm in viewFeedback method for example, do I have to pass the course_name and assessment_name and title?

What if there is a page that is one level deeper than viewFeedback, do I have to pass the course_name,  assessment_name, parent_title and title?

dlbucci:
The course is always passed as a link, and any page from the assessments_controller will have the assessment.  We may need to pass the assessment in other controllers, which is one drawback, but I don't think it would be hard to place that in a before_action that loads models.

jez:
I like this. It's definitely nicer than having all of our breadcrumb constructors hard coded.

Correct me if I'm wrong, but this removes functionality from the current breadcrumbs (from a user's perspective). The old one used to tell you when you were editing an assessment and editing a problem.

dlbucci:
Hence this being a proof a concept.  We'll need to add more breadcrumbs in more controllers/pages, but overall, it'll be more robust.

jez:
So then, the take-away so far is just that we've replaced a bunch of if statements with a `.each`, right? Do we have a timeline on when we want this finished by?

dlbucci:
I've been trying to get groups done, but after that I'll work on this.  It's just a minor fix for the Manage Submissions page, actually.

icanberk:
We have our plates pretty full at this point. We don't need to rush this for v1.1

dlbucci:
I think this should be ready for prime time, if someone wants to pull and test to make sure no obvious pages or links are broken.

jez:
Just a thought: what if we did breadcrumbs on the frontend? It would work like this: we store a list of pages that a user has visited in a list, push into that list when the user "descends" into a page, and pop from the list when a user circles back to a page they've already visited.

### Pros

- It accounts for some of our non-linear workflows, but still lets people easily navigate to the page that they last visited
- Dynamically creates breadcrumbs (rather than hard-coding them)

### Cons

- Potentially higher implementation complexity
- Have to handle what constitutes "circling back"; i.e., under what circumstances should we pop from the stack

dlbucci:
Eh, I'm not a big fan of relying on JS like that.  On a page like the gradesheet or the submissions#view, there's enough complexity that we can require JS for them to work, but I don't like the idea of breadcrumbs, an essential part of the navigation, failing if a user disables javascript.

Having them always be able to go back is nice, but we can't always rely on a user starting at the courses page, then going to the assessments page, and then to the gradesheet, or whatever.

Also, I don't think our current structure ever gets too deep, and this PR simplifies the breadcrumb creation by unifying how assessments are loaded across 6 or 7 controllers.

jez:
Oh, that's a good point about them not necessarily starting at the courses page. It's also understandable to want to make the site work as much as possible with JavaScript disabled.

icanberk:
It could be nice to do it with JS but I think it's too much effort for the amount of benefit at this point.

Issue #42, opened 2014-11-19
AssessmentCategories should probably be an Assessment field


dlbucci:
It's pretty useless as is.

icanberk:
@dlbucci have you done anything about this? otherwise you can assign it to me.

dlbucci:
No, you can take it, but having thought about it, I think there may be instructor code that determines scores based on AssessmentCategories, so I'm not certain this would be too simple.  At the very least, score computations need to still take category into account.  (SQL GROUP BY might help though)
Also, a fix for #142 may be more important, or at the very least can be done at the same time as this.

icanberk:
Good point. I started working on it, but we need to review all that grading code which seems nontrivial right now.

jez:
Closed by #276

PR #276, opened 2015-02-12
Replace AssessmentCategory model with assessment_name field

icanberk:


icanberk:
I'm not sure if this was worth it, but I've done it already.

This needs to be tested with a large enough data set since it touches gradebook.

app/models/assessment.rb
```diff
@@ -30,8 +29,9 @@ class Assessment < ActiveRecord::Base
                             :greater_than_or_equal_to => -1, :allow_nil => true
   validates_numericality_of :max_grace_days, :only_integer => true,
                             :greater_than_or_equal_to => 0, :allow_nil => true
-  validates_presence_of :name, :display_name, :due_at, :end_at, :start_at, :quiz, :quizData,
-                        :grading_deadline, :category_id, :max_size, :max_submissions
+
+  validates_presence_of :name, :display_name, :due_at, :end_at, :start_at,
+                        :grading_deadline, :max_size, :max_submissions
```
dlbucci: Don't we still want each assessment to have a category name?

app/models/course.rb
```diff
@@ -142,6 +141,13 @@ def students
     course_user_data.where(course_assistant: false, instructor: false, dropped: [false, nil])
   end
 
+  def assessment_categories
+    assessments.pluck(:category_name).uniq.sort
+  end
+
+  def assessments_with_category(cat_name)
```
dlbucci: I think assessments_in_category or assessments_by_category might be more descriptive names, but it's your call.

app/models/course.rb
```diff
@@ -142,6 +141,13 @@ def students
     course_user_data.where(course_assistant: false, instructor: false, dropped: [false, nil])
   end
 
+  def assessment_categories
+    assessments.pluck(:category_name).uniq.sort
```
dlbucci: uniq is an alias for distinct, which seems to be the newer, cooler way to specify this, so maybe go with that?  No big deal.

db/migrate/20150130054629_add_category_to_assessments.rb
```diff
@@ -0,0 +1,18 @@
...
+    
+    Assessment.all.each do |assessment|
+        sql = "SELECT name FROM assessment_categories WHERE id=" + assessment.category_id.to_s
+		records_array = ActiveRecord::Base.connection.execute(sql)
+	   	
+	   	records_array.each do |cat|
+    		assessment.category_name = cat[0]
+			assessment.save
+	    end
+    end
```
dlbucci: No big deal, but indentation is off here.  What's wrong with adding the category name column, then running something like
```
Assessment.all.each do |assessment|
  assessment.category_name = assessment.category.name
  assessment.save!
end
```

db/schema.rb
```diff
@@ -90,9 +87,10 @@
     t.boolean  "has_partners",       limit: 1
     t.boolean  "has_scoreboard",     limit: 1
     t.boolean  "has_svn",            limit: 1
-    t.string   "remote_handin_path", limit: 255
     t.boolean  "quiz",               limit: 1,     default: false
     t.text     "quizData",           limit: 65535
+    t.string   "remote_handin_path", limit: 255
+    t.string   "category_name",      limit: 255
```
dlbucci: Would we maybe want this to be just 'category'?  the '_name' seems unnecessary, but it's no big deal.

icanberk: I think it's good to be more explicit. (also makes it easy to refactor/debug for me)

db/migrate/20150130054629_add_category_to_assessments.rb
icanberk: Problem was that we removed AssessmentCategory from our codebase, so `category.name` does not work.

I'll fix the indentation.

dlbucci: Ah, well this should be fine, but yeah, I'm not sure we want to drop AssessmentCategories so quickly.  We may want to keep that table around for a bit until we're sure that grades aren't getting screwed up...

app/models/assessment.rb
icanberk: Yeah, I think we need to force this since a lot of things like category average depends on this.

icanberk:
I've tested this with real data on sawshark. Everything seems to work well. I've also almost documented a way of (manually) testing gradebook changes.

app/models/assessment.rb
dlbucci: Right, so add it back to the validates_presence_of?

icanberk: My bad! Fixed.

db/schema.rb
```diff
@@ -40,15 +40,12 @@
...
-    t.string  "name",      limit: 255
-    t.integer "course_id", limit: 4
-  end
-
   create_table "assessment_user_data", force: :cascade do |t|
     t.integer  "course_user_datum_id", limit: 4,               null: false
     t.integer  "assessment_id",        limit: 4,               null: false
     t.integer  "latest_submission_id", limit: 4
+    t.integer  "cgdub",                limit: 4
+    t.integer  "special_type",         limit: 4,   default: 0, null: false
```
dlbucci: Where are these columns coming from?

app/models/grade_matrix.rb
```diff
@@ -103,8 +102,8 @@ def matrix!
 
       @course.assessment_categories.each do |cat|
         a = cud.category_average(cat, @as_seen_by)
-        cat_avg_by_cat["#{cat.id}"] ||= {}
-        cat_avg_by_cat["#{cat.id}"]["#{cud.id}"] = a
+        cat_avg_by_cat["#{cat}"] ||= {}
+        cat_avg_by_cat["#{cat}"]["#{cud.id}"] = a
```
dlbucci: This could just be cat_avg_by_cat[cat].

db/schema.rb
icanberk: No idea...

dlbucci: Can we figure that out and possibly get rid of these if we don't need them?

icanberk: Yeah, but I think we should do it in another branch

icanberk: Yeah, but I think we should do it in another branch. I'll try to cleanup for now.

db/migrate/20150203151605_cleanup_old_fields.rb
```diff
@@ -22,9 +22,17 @@ def change
...
+        remove_column :submissions, :tweak_old, :float
+    end
+    
+    if column_exists? :submissions, :ignored_old
+        remove_column :submissions, :ignored_old, :boolean
+    end
+
+    if column_exists? :submissions, :absolute_tweak  
+        remove_column :submissions, :absolute_tweak, :boolean
+    end
```
dlbucci: Has this migration already been ran in production?  If so, we should just make a new migration.

icanberk: Nope. I just ran it on staging.

icanberk: Yeah it did... Do we really need to? It essentially does the same thing.

dlbucci: It won't run if you change an old migration file that's been run.  You'd need to rollback, which introduces a bunch of potential problems.

icanberk: I don't want it to run. Those columns are already removed.

dlbucci: I see.  Why did you need to make this change at all then?

icanberk: I ran migrations on sawshark and this one was giving an error. Not sure
why. It's not a big deal - I can just revert it but I thought the changes
just make it safer.

dlbucci: I just want to be very sure that these migrations are going to run and that the schema file will be accurate.

icanberk: I ran them couple times on staging server. We will be fine.

icanberk:
Can we merge this?

dlbucci:
Could you merge develop back into this so we can automerge?  I'd like to be able to reset my DB with develop HEAD schema, run the migrations in this branch and check that they work on my own machine.

icanberk:
@dlbucci done

dlbucci:
Second note: @icanberk, I'll probably hate you for getting this in before groups, cause now I have to resolve the conflicts!

icanberk:
:love_letter:

icanberk:
Closes #42

Issue #246, opened 2015-02-04
Bomblab and Datalab scoreboards not sorting properly


droh:
The 213 bomblab scoreboard should be sorted by score, then submission date. However, it doesn't appear to be sorting properly. The sorting code is contained in bomblab.rb. Similarly for datalab, which should be sorting by score, then number of operators.

PR #277, opened 2015-02-13
Overwrites the createScoreboardEntry 

icanberk:


dlbucci:
Fixes #246 ?

icanberk:
Yes, fixes #246

dlbucci:
This is fine, but eventually I think we can merge these config_module checks into one function.

Issue #260, opened 2015-02-07
Autolab populate script creates duplicate AUDs for LabTemplate


dlbucci:
After running the populate script, the LabTemplate assessment has two AUDs for each Assessment-CUD pair.  Other assessments seem to be fine.

PR #283, opened 2015-02-16
Don't create duplicate AUD for lab template

andrewwuan:
Lab template is created after users are populated, so AUD are already created.


andrewwuan:
https://github.com/autolab/Autolab/issues/260

andrewwuan:
Going to branch from `develop` instead

PR #284, opened 2015-02-16
Fixed Autolab:Populate

dlbucci:


icanberk:
did we test this? I feel like we are also required to add a `assessment.quizData` field

andrewwuan:
There's a category_id field that we need to take care of

dlbucci:
@icanberk: quizData is no longer required, and probably shouldn't be, as most assessments aren't quizzes (we could validate it's presence for quiz assessments, but it's not necessary at the moment).
@andrewwuan: category_id just got removed, because we removed the assessment_category model, which is why the script broke.

icanberk:
@dlbucci when did we make quizData not required? (I know we should but I just haven't seen the commit)

dlbucci:
Here? a83f89df94214346add8da2569355c8eb538bb75

icanberk:
Magic. Okay.

andrewwuan:
@dlbucci Have we migrated it yet? I still see `category_id` in db/schema.rb

dlbucci:
Yeah schema's a little out of whack.  If you migrate it, category_id should be removed.

PR #286, opened 2015-02-16
Fix autopopulate

andrewwuan:
https://github.com/autolab/Autolab/pull/284


dlbucci:
I already fixed autopopulate in #284, that was a pull request, not an issue.

PR #291, opened 2015-02-17
Hotfix for MOSS Archive support

ymzong:
Tested on local Autolab instance.

Note: MOSSing with external archives is still **not** working. The next PR is for refactoring out archive-related functions and making `runMoss` much cleaner.

### DIFF
--- a/app/controllers/admins_controller.rb
+++ b/app/controllers/admins_controller.rb
@@ -367,7 +367,12 @@ e.to_s() + e.backtrace().join("<br>")
         assessments << assessment
       end
     end
-  
+
+    require 'rubygems'
+    require 'rubygems/package'
+    require 'zlib'
+    require 'zip'
+
     @mossCmd= "mossnet -d "
   
     # Create a temporary directory for this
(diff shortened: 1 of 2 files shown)


app/controllers/admins_controller.rb
```diff
@@ -402,18 +408,28 @@ def runMoss
...
+        # If we need to unarchive this file, then create archive reader
+        arch_type = params["archiveCmd"][ass.id.to_s]
+        arch_path = "#{stuDir}/#{sub.filename}"
+        if arch_type == "tar" then
+          f = File.new(arch_path)
+          archive_extract = Gem::Package::TarReader.new(f)
+          archive_extract.rewind
+        elsif arch_type == "zip" then
+          archive_extract = Zip::File.open(arch_path)
+        elsif arch_type == "tar.gz" then
```
dlbucci: What about tgz files?  Wouldn't checking the mimetype for the file be a more accurate way to do this?

dlbucci:
How much work would getting external tar files working again be?  If it's not working, can we either remove the field on the moss page, and leave a note to the effect of "Currently, external tar uploads are not working. We hope to resolve this issue soon."?

app/controllers/admins_controller.rb
ymzong: Currently, it's the instructor's job to pick an archive type (`plain file`, `tar`, `tgz`, or `zip`), and we extract the archive. Those extensions are only an internal way to represent the type of the archive that the instructor picked, rather than the actual extension.

ymzong:
@dlbucci To make external tar working, I'd rather first implement a standalone library for archives. The reason is that an external tarball can contain an arbitrary number of student submissions as plain files or archive files, and it would be too messy to deal with them all in `runMoss`.

Sure, I can add a note there.

icanberk:
@ymzong please also merge this to `develop` and delete the branch.

ymzong:
@icanberk Done!

krha:
@ymzong 

Is it working now? I can submit an external tar file at moss, but the result does not make sense. The output of using external tar shows a list of file that matches with high probability as follows:
  > /opt/autolab/build/Autolab/tmp/kiryongh@andrew.cmu.eduMoss20150316-10576-1y3vgbp/external_input/submissions/ (73%)	

However, I cannot find this corresponding file at the external tarball. I don't think this feature is working now. Please let me know how I can use the external tar to compare against. Thanks.

ymzong:
Hi @krha, I will take a look at this later tonight and let you know.

krha:
Hi @ymzong 

Thanks. I don't mean to rush you, but It would be great if you can take a look asap since I need to report the result soon. Thanks.

PR #295, opened 2015-02-20
Implemented instructor email feature + cleanup admin functionality

icanberk:


icanberk:
1) I created an autolab_admin controller where we will be putting our autolab admin related features
2) Improved the `CourseMailer` to have more descriptive method names. (This should actually become `AutolabMailer` in the future)
3) Implemented the auth for administrator only. This was previously failing for pages where we don't have any course. I'm hoping that this will be useful as we have more admin only functionality. (eg. page to view system logs, systemwise stats etc.)

dlbucci:
I would prefer if course-specific functionality in admins_controller was moved into courses_controller, and then admins_controller became the autolab-specific admin controller.  It's silly to have two 'admins' controllers, when most of the functionality of the current admins_controller rightfully belongs in the courses_controller.

icanberk:
That actually makes sense.

dlbucci:
Also, what actions were failing?

icanberk:
By default, we call `authorize_user_for_course` for every auth. If there is no course, like the page where we email to all instructors, the actions were failing..

dlbucci:
But it wasn't happening on any pages before?  Couldn't you do a skip_before_action on other methods where there isn't a course (like courses#index, or users#index)?

icanberk:
@dlbucci moved old admin stuff into course and added autolab admin menu to admins_controller per your request.

app/controllers/admins_controller.rb
```diff
@@ -2,603 +2,38 @@
...
-                      major: cud.major,
-                      year: cud.year,
-                      grade_policy: cud.grade_policy,
-                      lecture: cud.lecture,
-                      section: cud.section,
-                      color: "red"}
-            @cuds << newCUD
-          end
-        end
+      @cuds = CourseUserDatum.select(:user_id).distinct.where(:instructor=>true)
```
dlbucci: This should probably just email instructors of courses that are currently running on autolab. No point emailing professors who aren't currently using Autolab (or at least, there should be an option to limit the emails to a current professors)

app/controllers/admins_controller.rb
```diff
@@ -2,603 +2,38 @@
...
-    else
-      # No header row. Columns are:
-      # Semester(0), Email(1), Last Name(2), First Name(3), School(4), 
-      # Major(5), Year(6), Grade Policy(7), (skip)(8), Lecture(9), 
-      # Section(10), ...
-      return parsedRoster
-    end
-  end
-      
 end
```
dlbucci: 604 to 39.  Nice.

app/controllers/admins_controller.rb
icanberk: Dan, do you know how hard it is to write that query???

app/controllers/courses_controller.rb
```diff
@@ -172,6 +255,454 @@ def userLookup
 
   end
 
+  action_auth_level :users, :instructor
+  def users
```
dlbucci: RESTfully speaking, this should just be course_user_data#index

app/controllers/courses_controller.rb
```diff
@@ -172,6 +255,454 @@ def userLookup
...
+
+
+  # Upload a CSV roster and import the users into the course
+  # Colors are associated to each row of CUD after roster is processed:
+  #   green - User doesn't exist in the course, and is going to be added
+  #   red - User is going to be dropped from the course
+  #   black - User exists in the course
+  action_auth_level :uploadRoster, :instructor
+  def uploadRoster
+    if request.post? then
```
dlbucci: Check the routes, but this check may be unnecessary (somewhat out of scope of this pr).

app/controllers/courses_controller.rb
```diff
@@ -17,6 +21,85 @@ def index
...
+
+       {"name"=>"Reload course config file", path: course_reload_path(@course),
+        "title"=>"Do this each time your modify the course.rb file"},
+
+       {"name"=>"Bulk release all grades", path: bulkRelease_course_course_user_datum_gradebook_path(@course, @cud),
+        "title"=>"Release all grades for all assessments"}
+
+       # TODO: re-add CSV export
+       # {"name"=>"Export grades as CSV","controller"=>"gradebook","action"=>"csv"}
+    ]
```
dlbucci: Maybe out of scope for this pr, but these should all be links in the view.  I kind of hate these action lists we have all over the place.

icanberk: It needs to be somewhere...

config/routes.rb
```diff
@@ -128,10 +132,7 @@
       end
     end
     
-    resource :admin, only: :show do
-      collection do
-        get 'throwException'
-      end
+      get 'admin'
```
dlbucci: Do you think this should be something like 'settings' or 'options'? It's a bit confusing to have two admin routes, but it's no big deal.

app/controllers/courses_controller.rb
dlbucci: Yeah, but these are just for creating links, and if you need to list them anywhere, just bake them into the view to reduce clutter in the controllers.  You don't need to do it now, but we should in the future.

app/controllers/admins_controller.rb
icanberk: Do we have a concept of active/inactive course?

icanberk:
```
def temporal_status(now = DateTime.now)
    if now < start_date
      :upcoming
    elsif now > end_date
      :completed
    else
      :current
    end
  end
```

Does this work?

dlbucci: select(user_id,course_id).distinct.joins(:courses).where(courses:"end_date>Time.now", instructor:true)?
Or since you are already looping over all cuds, you could remove ones with courses with end_dates that have passed.

dlbucci: That function looks good.  Does it go in courses.rb?

icanberk: It's already there

icanberk:
@dlbucci donzie

dlbucci:
And you've tested that the query works and the emails get sent?

icanberk:
Query is tested. I assume emails get sent cuz we didn't change the way we send them.

icanberk:
I mean I tested the CourseMailer methods work.

dlbucci:
Okay, good enough for me.

PR #305, opened 2015-02-21
Moved autoconfig into config and renamed it to autogradeConfig

icanberk:


app/controllers/jobs_controller.rb
```diff
@@ -1,7 +1,7 @@
-require 'autoConfig'
-
 class JobsController < ApplicationController
 
+  autolabRequire File.join(Rails.root, 'config/autogradeConfig.rb')
```
dlbucci: This seems to miss the point of using join. I'd use 'Rails.root.join("config", "autogradeConfig.rb")'

icanberk:
@dlbucci not sure if Rails.root.join is a thing. We seem to use `File.join(Rails.root, ...` throughout the codebase so I changed it to that.

dlbucci:
It absolutely is a thing, I've used it in many places.
See: https://github.com/autolab/Autolab/blob/develop/app/controllers/assessment/autograde.rb#L29

icanberk:
Just for you, Dan.

icanberk:
@mihirpandya you should be aware of this change.

PR #306, opened 2015-02-21
Moved JS and CSS into proper content_for

dlbucci:
Feeling pretty salty right about now.


.gitignore
```diff
@@ -19,3 +19,4 @@ tmp/
 *.swp
 db/*.sqlite3
 tags
+.DS_Store
```
icanberk: Thanks god.

app/views/assessments/_gradesheet.html.erb
```diff
@@ -1,19 +1,24 @@
...
+  <%= javascript_include_tag "KeyTable" %>
 
-<script type="text/javascript">
-  andrewID_col = 1
-  problem_count = <%= @assessment.problems.length %>
-  first_problem_column = 5;
-  first_non_searchable_column = 3 
-</script>
+  <script type="text/javascript">
+    andrewID_col = 1
```
icanberk: Seems like we should fix this at some point.

dlbucci: Fix what, the gradesheet? This fixes it.  The scripts trusted that jquery had been loaded, but since they weren't in the content_for, jquery was loaded after the application scripts ran.

icanberk: I mean the `andrewID`

icanberk: Actually, do you mind fixing it in this PR? You just need to change the class="andrewID" to email, and change this line.

dlbucci: Sure thing, cappytahn.

app/assets/javascripts/annotations.js
```diff
@@ -0,0 +1,371 @@
...
+  var path = '<%= course_assessment_submission_annotations_path(@course, @assessment, @submission) %>';
+  var createPath = function(annotation) {
+    return path + '.js';
+  };
+  var updatePath = function(annotation) {
+    return path + '/' + annotation.attr('id').replace(/[^0-9]*/, '') + '.js';
+  };
+  var deletePath = updatePath;
+
+  var problems = <%== @problems %>;
```
icanberk: This line is not working.

icanberk: Since rails does not pass the array of problems here. Perhaps we should move this into the page. What do you suggest?

icanberk: By move into the page I just mean that one line.

dlbucci: We can move the variable declarations.  I'm actually rewriting the annotations page, so I can fix it then.

icanberk: Sounds good. There are couple other variables in this script fyi.

```
+          submission_id: '<%= @submission.id %>',
+          filename: '<%= @filename %>',

```

PR #315, opened 2015-02-24
Fixed Assessments#new

### LINKED ISSUE #279: Assessment Builder doesn't import optional modules
When creating a new assignment with the Assessment Builder, the optional modules selected on the first page are not imported.

dlbucci:
Fixes #279


ymzong:
@dlbucci Could you also merge into `develop` and delete branch?

icanberk:
I don't think we have a `develop` right now. I'll make one after releasing v1.1.0

PR #318, opened 2015-02-24
Feature slickgrid submodule

dlbucci:
This WILL be a Ruby project when I'm done with it.


icanberk:
So instead of submoduling things, why don't we have a list of JS dependencies that can be easily installed? (either npm or grunt would do it, i guess)

From the article:

```

On the other hand, if the technological context allows for packaging and formal dependency management, you should absolutely go this route instead: it lets you better split your codebase, avoid a number of side effects and pitfalls that litter the submodule space, and let you benefit from versioning schemes such as semantic versioning (semver) for your dependencies.

```

icanberk:
Again from the article: "JavaScript has npm and Bower (and others, too)"

cc @jez

icanberk:
I've dealt with submodules while I was at facebook, and it's actually really painful. Especially since we assume that people will be using this repo in production, I think we should not be expecting them to git submodule update and stuff.

dlbucci:
The issue is this particular javascript repo doesn't seem to be readily available anywhere (on any package managers), short of downloading from GitHub, so cloning the repo isn't that bad (especially since it doesn't update that often).

jez:
So one thing that we could do is to package the project into the separate project that we control, turn it into a Ruby Gem, and then load the assets in through the Ruby asset pipeline. If the repo really doesn't update that often (which it doesn't because it's kind of abandoned right now...), it would be perfectly viable to maintain a Ruby Gem with the version frozen at whatever version we need to make Autolab stable.

jez:
We don't even need to push the gem to rubygems.org; we can just load it [straight from GitHub](http://bundler.io/v1.7/git.html).

jez:
Submodules are doable, we just have to explicitly remind people to run git submodule init/update all over the place.

dlbucci:
That does sound better, although I have no concept of how hard packaging a Ruby gem is.  If some one else wants to take this on, they can by all means closes this issue.

dlbucci:
But is everyone okay with removing the public javascript stuff? If so, I might re-remove those files and push to master.

jez:
You know the rails app better than I do, but `public/` seems to be pretty useless to me. I'm not familiar with the Gemfile format, but I'm willing to look into it, because I think it might be something worth doing for a couple other assets.

I was kind of disappointed that our net SLOC for the 1.1.0 merge from develop to master was positive though... This makes up for that :+1:

dlbucci:
files in public still get served if rails crashes and apache/passenger need to serve a page up, but the javascript definitely does not need to be their.

jez:
Go ahead and delete them then, I'm closing this for now.

icanberk:
I also like the idea of creating a Gem for slick grid.

It also seems that slickgrid is on bower: http://bower.io/search/?q=slickgrid

PR #322, opened 2015-02-24
Fix Contact Us and Maintenance pages

### LINKED ISSUE #314: Contact Us and Maintenance Pages Need an Update
The "Contact" link at the bottom of every page and the "Autolab is down for maintenance page" is outdated with regard to the design updates that have taken place.

- [x] Contact page
- [x] Maintenance page

jez:
Commit Summary
--------------

85d35c1
   Revamp maintenance page. Closes #314.

   I basically cURL'ed the new contact page, stripped out the old and shoved
   in the new.

0148fa7
   Add new contact us page


icanberk:
What happens if there is a 500 error and apache cannot reach the rails app?

dlbucci:
I think you'll need to add a contact action to the home_controller for this to work (not positive), and when you do, you'll also need to make sure the authentication gets skipped.

For the maintenance page, I'd called render :maintenance in the maintenance before_action (https://github.com/autolab/Autolab/blob/master/app/controllers/application_controller.rb#L123) in application_controller, that way you can use the site layout.

Also, it might be nice to check an environment variable to see if we're in maintenance mode (AUTOLAB_MAINTENANCE_MODE or whatever), that we we could activate it with you editing the file/restarting the server.

icanberk:
For skipping auth, we have a filter like `:action_no_auth `

dlbucci:
Or you can add a skip_before_filter in the controller (which are actually there for homes_controller).

jez:
Did you notice an error reaching the contact us or maintenance pages because of authentication? I didn't see any when I was testing.

dlbucci:
Neither of these would produce errors, actually (I didn't test).  My point should have been that there should be at least an empty route in home_controller (to let us know /contact is a route) and that maintenance can use a render call so that it can use the site layout, vs us having to update maintenance.html every time we make changes to the site layout.

jez:
So I figured that maintenance wouldn't be able to use a render call if the site was down for maintenance... doesn't it have to stay in `public/`? Or am I not understanding how maintenance mode works.

dlbucci:
https://github.com/autolab/Autolab/blob/master/app/controllers/application_controller.rb#L119
That method gets called before every action (see the before_action at the top of the controller).  That's how maintenance mode gets enabled, so the rails app is still running, it just redirects everything (note, it probably should not redirect autograde_done calls, that way tango jobs that start before maintenance is enabled will still be able to finish).

jez:
Alright, I think this is ready.

PR #332, opened 2015-02-24
Autograding Retries on Failure now

dlbucci:
Gonna test this on staging before we merge, but please take a look and see if anything looks wrong.  We'll need to add a config option as well.


icanberk:
I'm not sure how I feel about this. I think retry on the front-end doesn't really help much with the situation but add complexity.

What do you guys think?

cc @mihirpandya @yrkumar @droh

dlbucci:
It's obviously addressing symptoms and not the problem, but it's also added robustness.  If we can assure a decent success rate, I can't see why we wouldn't want this (plus it's not much complexity added at all).

icanberk:
That's true. We don't really lose much here. But I'd still want to hear people's opinion's.

dlbucci:
We'll table this for now.

PR #333, opened 2015-02-24
Assessment builder was not installing from directory

icanberk:


icanberk:
This will help us test doslingos now.

app/controllers/assessments_controller.rb
```diff
@@ -1671,7 +1682,7 @@ def scoreboard_prop_params
   private
   
     def new_assessment_params
-      params.require(:assessment).permit(:display_name, :category_name, :has_autograde, :has_svn, :has_scoreboard, :group_size)
+      params.require(:assessment).permit(:name, :display_name, :category_name, :has_autograde, :has_svn, :has_scoreboard, :group_size)
```
dlbucci: This really isn't okay seeing as we don't want instructors to change the name of assessments, but seeing as we allow them to change everything in edit_assessment_params, I guess it's okay for now...

PR #339, opened 2015-02-25
Unlimited multipart

dlbucci:
Should fix the server errors that moss has been causing.  Also updates a config file to be a template, so that change will need to be made during the pull to production.


icanberk:
pulled to prod.

jez:
Should `config/initializers/devise.rb.template` have been .gitignored?

dlbucci:
No, but config/initializers/devise.rb should have.

jez:
Whoops, yeah that's what I mean. It's not being ignored currently.

dlbucci:
Right, if you want to commit that change develop right now, go for it.

PR #340, opened 2015-02-25
Remove Bootstrap 2.3.1 Dependency

### LINKED ISSUE #178: Drop Bootstrap 2.3.1 Dependency
It looks like we're still using Bootstrap 2.3.1 in these views:

- `app/views/gradebooks/view.html.erb`
- `app/views/gradebooks/statistics.html.erb`
- `app/views/gradebooks/student.html.erb`
- `app/views/assessment_user_data/edit.html.erb`

As far as I can tell, the only reason why this has been kept around is for tooltips, which Bootstrap 3 also supports.

- [x] Replace Bootstrap 2.3.1 with Bootstrap 3.0.0 analogues everywhere

### LINKED ISSUE #320: Gradebook assessment names are cut off
When viewing the student gradebook, the column of assessment names is too narrow to fit even the word "Homework 0".

- [x] Fixable by making the viewport larger (it's currently capped at 970px)

jez:
Fixes #178. Fixes #320.

There were three views where Bootstrap 2.3.1 was being used:

- Assessment User Data (Edit)
    - Files affected:
        - app/views/assessment_user_data/edit.html.erb
    - This view was easy (it has very little complexity).
- Student Gradebook
    - Files affected:
        - app/assets/stylesheets/student_gradebook.css
        - app/assets/stylesheets/student_gradebook.css.scss
        - app/views/gradebooks/student.html.erb
    - Bootstrap 2.3.1 was being used both to style the layout of the
      page as well as the draw tooltips.
    - I updated the layout HTML to Bootstrap 3.0.0
    - I updated the tooltip usage to Bootstrap 3.0.0
    - I rewrote the student_gradebook.css file in SCSS
- Instructor Gradebook
    - Files affected:
        - app/views/layouts/application.html.erb
        - app/views/layouts/home.html.erb
        - app/views/gradebooks/view.html.erb
    - This page is using SlickGrid to display the instructor gradebook
      and was using jQuery UI + Bootstrap 2.3.1 Tooltip to display the
      tooltips.
    - So here's the dealio:
        - SlickGrid requires jQuery UI to sort columns
        - jQuery UI has features that enable tooltips
        - Bootstrap 2.3.1 integrates with jQuery UI tooltips
        - Bootstrap 3.0.0 has it's own tooltips implementation
        - jQuery UI interferes with Bootstrap 3.0.0 when trying to draw tooltips
    - I fixed this by adding no conflict method to the jQuery UI tooltip
      widget.
    - See here for more information
        - <http://stackoverflow.com/questions/13731400/jqueryui-tooltips-are-competing-with-twitter-bootstrap>


jez:
@icanberk Is there something left to do here? I don't know that there's anything we can do to remove jQuery UI right now, as we'd have to re-evaluate our use of SlickGrid.

Also, in the process of updating tooltips on the Student Gradebook, I fixed #320, which was an interface/design bug.

dlbucci:
Changing a grade on assessment_user_data#edit works, but displays two error messages.

dlbucci:
Score Links in the student gradebook are broken.  They effectively link to the current user's gradebook, passing the assessment name as a query variable.

dlbucci:
They should instead point to that user's history for that assessment (at least, that's what they pointed to in autolab 2).

app/views/gradebooks/view.html.erb
```diff
@@ -58,9 +58,18 @@
...
 
-<% content_for :javascripts do %>
+
+<% content_for :js_widget_patch do %>
   <%= external_javascript_include_tag "jquery-ui", "1.10.2" %>
-  <%= external_javascript_include_tag "bootstrap-tooltip", "2.3.1" %>
+  <%# jQuery UI is required for SlickGrid, but Bootstrap provides a tooltip plugin %>
+  <%# The following must come after jQuery UI but before Bootstrap 3 is loaded %>
+  <%# to resolve namespace issues %>
+  <script>
```
dlbucci: type="application/javascript" :P

jez:
I don't quite understand what you mean about the query string links. It looks to me like they're finding that assessment and linking to it's history. The query string param being passed is cud_id, not assessment name. When I click on the links, it takes me to assessment history.

app/views/layouts/application.html.erb
```diff
@@ -68,6 +68,7 @@
     </div>
 
     <%= external_javascript_include_tag "jquery", "2.1.1" %>
+    <%= content_for :js_widget_patch %>
```
dlbucci: This seems like a really hacky way to go about this, but I honestly can't think of a better way to fix this problem.  I think we should try to get rid of these layout include tags in a future PR, so that we don't need to load jQuery and bootstrap for pages that don't use it.

dlbucci:
That's for the history link glyphicon.  Click a score number and you'll be taken to your own gradebook, with ?assessment=<blah> as the query string.

dlbucci:
Besides those minor issues, this looks good, although previously the student gradebook didn't have the lines in the score tables, and that looked better IMHO.  But I'll leave that decision up to a more design-oriented dev.

app/views/layouts/application.html.erb
jez: I agree about the hackiness. My personal best-case scenario is that we fix the issue by replacing SlickGrid. It would remove our dependence on jQuery UI, and also free us from potential problems down the road (since support has been dropped). Until then, though, it appears that this solution works.

Just about every page loads Bootstrap (maybe not every page needs the JS), but certainly only some pages need jQuery.

jez:
I actually see zero links that have `?assessment=` in them. I'm looking specifically at links on this page `http://localhost:3000/courses/1/course_user_data/2/gradebook/student` (i.e, the Student Gradebook for user0_autopopulated), and hovering over the links for Final Score.

jez:
Also, I don't know what the gradebook's design looked like before, so I'm not sure what changes I'd make from a design perspective.

dlbucci:
It looked like this, but slightly smaller (I like it being bigger), and without the internal lines in each category box.

jez:
- Assessment history linking is fixed
- I tweaked the size of the history icon so that it doesn't make the tr resize on hover
- I personally like the horizontal dividing lines, and I think Ilter recently put them there, so I'm just gonna leave them

Anything else?

PR #343, opened 2015-02-26
Redirect Students who access unreleased assessments

dlbucci:


icanberk:
So `assessments_with_category` caused all the assessments to show up, and we didn't protect those pages before?

PR #348, opened 2015-02-27
Fix annotations to support edit and [ ] chars

icanberk:


app/assets/javascripts/annotations.js
```diff
@@ -39,15 +39,21 @@ var parseText = function(text) {
...
-    res2 = res[1].split(":");
-    if (res2.length === 1) {
-      return [decodeURI(res[0]), res[1].split("]")[0], ""];
+  } 
+  if (res.length > 1) {
+
+    var len = res.length;
+    var points = res.splice(len-1,1);
+    var points = points[0].split(":");
+    var comment = decodeURI(res.join("["));
```
jez: What's going on here? Are you like manually decoding the string to handle `[` ?

icanberk:
Although we have a proper annotations model, some dev at some point for some unknown reason decided to encode annotations as "[comment][problem:points]" and just use one of the columns in the table... So we need to parse the parts of it.

For the next version, we'll migrate that column and put them into separate columns but we need to deal with this for now.

jez:
Reminder: pull to prod.

PR #350, opened 2015-02-28
Feature improved submissions

icanberk:
* Upgrades the data tables row
* Makes rows selectable for custom actions
* Adds an isLatestSubmission field for submission
* Adds a "only show latest" filters
* Creates a "Regrade N Submissions" button


icanberk:
I think we can bump the patch number after this.

PR #351, opened 2015-02-28
Fix install assessment

icanberk:


icanberk:
Whoops. I accidentally checked out from my other branch so there are those extra commits but it should be fine since we'll merge both into develop.

app/controllers/assessment/autograde.rb
```diff
@@ -63,6 +63,50 @@ def regrade
     redirect_to history_course_assessment_path(@course, @assessment, cud_id: @effectiveCud.id) and return
   end
 
+
+  # 
+  # regradeAll - regrade the most recent submissions from each student
+  #
+  # RESTfully speaking, this belongs in submissions controller,
+  #   but because it uses autograding, it is easier to have it here
+  # action_auth_level :regradeAll, :instructor
```
andrewwuan: It may not be your responsibility, but why did we comment out action_auth_level for both regradeBatch and regradeAll?

icanberk: I'm not sure... @dlbucci did it so I assumed he has a good reason. we can fix in a future PR.

PR #352, opened 2015-02-28
Created the migration that will parse the text

icanberk:


icanberk:
I have tested the migration locally and it seems to work. Although I don't know why some comments are encoded( e.g. "yotestlo%5B5%5D%5B6%5D")

If you can fix/make sure migrations works, I will change annotation.js to comply with this proper model.

Loves and kisses,

Ilter

db/migrate/20150228200007_split_annotations.rb
```diff
@@ -0,0 +1,43 @@
...
+
+            points = points.split(":");
+            comment = res.join("[");
+            annotation.comment = comment
+           
+            if points.size == 1 then
+              annotation.value = points[0].delete("]").to_f
+            else
+              if points[0] == "?" then
+                annotation.problem = points[1].delete("]").to_i
```
dlbucci: Isn't this assigning the name of the problem, where as the problem is now a relation?  I.e. this should be Problem.find_by(name: points[1].delete("]").to_i)?  (Also, accounting for assessment/course scope)

dlbucci:
Darling @icanberk,
The URI encoded text occurs because we switched to using that model for escaping monday (and possibly switched back since?).  I think we can detect if we need to decode by checking for the space character (" "), which should be in most comments, and should also have gotten escaped.  So if a comment has no space, URI.decode that bitch.
-Daniel

db/migrate/20150228200007_split_annotations.rb
icanberk: No it has the id of the problem. For some reason this works.
But I guess the correct way would be

`annotation.problem = Problem.find_by_id(points[1].delete("]").to_i)`

dlbucci: Looking at the DB suggests that this is a lie, at least for some problems.  We may have accidentally switched that on monday.

icanberk: You're right. e.g. "don't leave in print statements [-0.5:CancelFlightsCorrectly]"

But we switched to IDs as of Monday...

I think I'll try to parseInt and if it fails, it means that it's just the name of the problem so I'll find by name. How does that sounds?

dlbucci: I don't think so, I found uri-encoded texts that have problem names as well.

icanberk: This is an example of after monday: "5sdfzdf[5:problem2]"

I thought problem2 would be the ID, but it might actually be the name of the problem... My b.

dlbucci: Also if you test for the " " character and it's not there, the text will probably not be uri-encoded.

icanberk: But I don't want to use `Problem.find_by(name:` because there are generic names like "problem1" so it'll find the wrong problem.

dlbucci: Hence me saying scope it to the assessment somehow (annotation.submission.assessment?).

dlbucci:
A) cleaner is subjective.  I might call it slower :P
B) We should still load it on only pages that use it.  That's what we should be doing for every javascript library, so that we keep our load time down.  If it gets used on every page, then we can load it on every page, but until then, I'd much rather we explicitly declare where it will be used by loading it on those pages.

icanberk:
It's 48 KB man. And it get's cached.

dlbucci:
That's hardly an excuse for loading it on every page, and I've never really bought the cache argument.

icanberk:
Promise I will use it on every page.

dlbucci:
That's great, but until you do, please only load it on the pages that need it.

dlbucci:
Thank you, and apologies for my dickishness on this issue.

icanberk:
So can we merge?

dlbucci:
Let me test this.  Also, I'd like to see URI decoding in the migration.

icanberk:
Can you add it and push?

dlbucci:
Sure.

icanberk:
apparently http://ruby-doc.org/stdlib-1.9.3/libdoc/uri/rdoc/URI/Escape.html#method-i-unescape

db/migrate/20150228200007_split_annotations.rb
icanberk: I only fixed it on the line below but i guess we also need `problem = annotation.submission.assessment.problems.find_by(name: problemStr)` here

dlbucci: I noticed.

dlbucci:
Got one more commit coming, then we can test.

dlbucci:
If this works, we can clean up app/models/annotations.rb next.

PR #353, opened 2015-02-28
Renamed getCategory to setCategory

icanberk:
Purpose of the page is to set the category for the assessment so it should be named accordingly.


dlbucci:
I dunno man, this is a pretty big important pull request.

icanberk:
If stuff breaks, it's your fault for not reviewing throughly.

dlbucci:
How does one review throughly?

PR #358, opened 2015-03-02
Improvements to gradesheet

icanberk:
Please don't ask me why or how this works.


icanberk:
1) Creates larger text areas for feedback
2) Saves when the TA closes the box without pressing save.

icanberk:
@jez you should just git pull and test it manually.

dlbucci:
Why or how does this work?

dlbucci:
Wait, did you switch to the backbone gradesheet?  If so, there's a corresponding gradesheet2.html.erb we should probably be using.

icanberk:
I didn't switch to backbone?

I just renamed gradesheet1 to gradesheet. 

I think I should get rid of gradesheet2.html.erb. What do you say?

icanberk:
Where did you see Backbone?

dlbucci:
Oh, the old gradesheet.js file was the JS for gradesheet2.html (sane, right?), which was a backbone gradesheet we never used in production.  That's what you deleted.  So yes, I would get rid of gradesheet2.html.erb if we are never going to use it.

icanberk:
Can someone test this?

dlbucci:
I sort of fail to see why that binding change would affect anything.

dlbucci:
Gotcha.

icanberk:
It's also good because if we dynamically add rows after initial load, it will still work.

jez:
We should set `cursor: pointer;` on the email links or set `href="#"` so that Chrome uses a pointing hand rather than a text-editing cursor.

jez:
Other than that, it looks like this is solid :+1:

icanberk:
thanks for taking care of this.

icanberk:
:100:

jez:
#teamwork :P

dlbucci:
@icanberk Didn't you combine the email and name columns? I'm not seeing that when I run develop on my machine.

PR #360, opened 2015-03-02
Annotations Bonfire

### LINKED ISSUE #359: Annotations model and code should be updated
Now that we actually use the comment, value, and problem columns, we can simplify a lot of the annotations code.

dlbucci:
Fixes #359
Should also make new annotations viewable again.


jez:
Just because it happened in this PR, is there any reason why a lot of the ruby code is 755 bs 644?

dlbucci:
Don't know.  We actually had annotations_helper (which was empty) and annotation_helper (which wasn't being used).  I emptied the latter one function at a time until it was empty and the mv'd the name, which probably explains the permissions changing.

icanberk:
Can we merge this to `develop`?

icanberk:
pulled and restarted

icanberk:
plz push to develop as well

dlbucci:
Done.

Issue #330, opened 2015-02-24
Failed autograde should not smash scores


droh:
If the autograder doesn't return a parseable json autoresult string, then the front end should not overwrite any already existing scores.

dlbucci:
Fixed by #362

PR #362, opened 2015-03-02
Don't overwrite grades on autograding failure

dlbucci:


dlbucci:
Fixes #330

app/controllers/assessment/autograde.rb
```diff
@@ -448,23 +456,17 @@ def saveAutograde(submissions, feedback)
...
-          score.released = true
-          score.grader_id = 0
-          score.save!
+          if score.new_record? then # don't overwrite scores
+            score.score = 0
+            score.feedback = feedback_str
+            score.released = true
+            score.grader_id = 0
+            score.save!
+          end
```
icanberk: So if the job got regraded, the score will not be a `new_record`. Am I correct?

dlbucci: It might be a new_record if the autograder did not set some problem scores (like style scores).  But if a TA sets a score for a problem, it will not be overwritten.

icanberk: Oh this overwrite only if it's a new_record.

What if the regrade was actually successful and we want to overwrite the grade?

dlbucci: That won't happen in the rescue block.  That code is still unchanged.

icanberk: Ok. I see

PR #374, opened 2015-03-08
Small Fixes

dlbucci:
* submissions#view failing will send users to the CUD's history, not their own
* assessments fail loudly when they fail to update
* student? is now true for administrators who aren't CAs or instructors

### DIFF
--- a/lib/pdf.rb
+++ b/lib/pdf.rb
@@ -1,0 +1,10 @@
+module PDF
+  def self.is_pdf?(file)
+    return false unless file
+    # all PDFs have '%PDF-<major-version>.<minor-version>' in them somewhere.
+    # this is used instead of MIME type detection because it should work
+    # for PDFs inside of Archives as well.
+    # the scrub call prevents invalid byte sequence errors.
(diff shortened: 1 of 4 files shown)


lib/pdf.rb
```diff
@@ -0,0 +1,10 @@
+module PDF
+  def self.is_pdf?(file)
+    return false unless file
+    # all PDFs have '%PDF-<major-version>.<minor-version>' in them somewhere.
+    # this is used instead of MIME type detection because it should work
+    # for PDFs inside of Archives as well.
+    # the scrub call prevents invalid byte sequence errors.
+    return (file.scrub("*") =~ /\%PDF-\d+\.?\d+/)
+  end
+end
```
jez: I know filetype detection is inherently janky, but what is this trying to solve? What was the original issue?

dlbucci: Well, a few courses get PDFs submitted, and before this, submissions#view would simply redirect after trying and failing to render a PDF.  But I figured it was a bit silly for us not to be able to show PDFs in the browser, so this will now send the file if a user tries to view a PDF, which means PDFs should be viewable in browser now.

Issue #346, opened 2015-02-27
Regrade-a-couple workflow broken


robsimmons:
Maybe this is related to #316 - I used to be able to hold down command and click a bunch of different "Regrade" links to regrade 5-50 assignments. (Lately this workflow is useful because of [excess bytes elided], but of course usually it's **me** that broke the autograder, necessitating a partial-mass-regrade.)

Today I noticed that this no longer works - command-clicking takes me forcibly to another page, and command-click "Open in new tab" has the unusual behavior of 404ing but, as far as I can tell, successfully spawning the regrade. Is there any chance that this old "open a new page" workflow could be preserved? It's much lighter weight than having a tickybox "select all" "regrade all selected" interface that would probably be necessary in its absence.

PR #375, opened 2015-03-08
Fix Manage Submissions Toggle

jez:
> On March 6, 2015, Iliano Cervesato wrote:
>
> Hello,
>
> in the “Manage submissions” page (15-312Q: Foundations of Programming
> Languages (Qatar) (s15) » hw03 » Manage Submissions), if I click on the links
> or buttons inside the submission table, it toggles the checkbox rather than
> taking me to that link or performing the action associated with that button.
>
> That means that I cannot look at the submission of an individual student: I
> have to download them all and then look at that student’s code.
>
> Many thanks,
>   Iliano

Commit Summary
--------------

154cfa1 Strip whitespace

f6a5a4e Toggle only when an anchor tag wasn't clicked

### DIFF
--- a/app/assets/javascripts/manage_submissions.js
+++ b/app/assets/javascripts/manage_submissions.js
@@ -116,4 +116,13 @@ jQuery(function($) {
     e.stopPropagation();
   });
 
+  $('.regrade-override').click(function(e) {
+    // Because regrade requests are sent with `data-method="post"`, we need to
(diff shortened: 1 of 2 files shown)


dlbucci:
While we're touching this file, should we fix #346 ?

jez:
@dlbucci This should be ready to go now.

icanberk:
We could actually make this an hotfix and merge to master.

PR #385, opened 2015-03-12
Add datepickers to the Announcements forms

### LINKED ISSUE #356: Announcements Form needs new Datepickers

jez:
Commit Summary
--------------

2c0aacf
- Strip whitespace

26c10e0
- Make datetimepicker support code more robust

58d09a4
- Use FormBuilder for datetimepickers. Fixes #356.

fd7558f
- Move 'confirm changes' script to dedicated file.


dlbucci:
The dates are initially blank when I load the page, and stay that way until I click them (although they appear to work).
Also the persistent announcements are shown on every page in a course, vs only assessments#index, so the description is wrong.

jez:
The dates are initially blank for me too (the form used to be that way too).

 I changed the description, so this should be good to go.

dlbucci:
The old announcement form, you mean?  Is there a way we can fix it?

jez:
What would you prefer? It's a form for creating a new announcement. There's nothing we could pre-populate it with except placeholder data, but then they'd just have to delete it.

dlbucci:
Okay, do they show up for the editing form? And does it work if you never
touch them?

jez:
Oh, it looks like it doesn't on the Edit form. I'll see if I can figure it out. My initial guess is that it's just not saving the field on submit.

jez:
This should be fixed in 364858f.

app/views/announcements/new.html.erb
```diff
@@ -1,4 +1,8 @@
 <h2>Create Announcement</h2>
-<%= form_for :announcement, url:{action:"create"} do |f| %>
-  <%= render partial:'announcementFields', locals:{f:f} %>
-<% end %>
+<div class="row">
+  <div class="col-md-6">
+    <%= form_for @announcement, url:{action:"create"}, builder: FormBuilderWithDateTimeInput do |f| %>
```
dlbucci: This line causes an error for new announcements.  The announcement variable needs to be initialized in the controller.

dlbucci:
Did you fix the create announcement error?

jez:
Yeah, but I remade the PR because of the merge conflicts.

PR #388, opened 2015-03-14
Feature debugging gems

dlbucci:
I'm creating this for people to see what running rubocop looks like.  The auto-correct turned 7000 errors into 1700.


dlbucci:
For an example of configuration settings we can change in the future:
https://github.com/bbatsov/rubocop/blob/master/config/default.yml
Also, I have not tested this, so it's not 100% certain all code still works (not that we're ever certain of that *cough*).

icanberk:
Do you want me to test this on staging?

dlbucci:
Yeah.  It's a lot of changes and the rubocop repo said auto-correct was experimental, so the results might not be accurate.

icanberk:
/opt/autolab3/build/autolab-src/app/controllers/application_controller.rb:111: syntax error, unexpected $undefined, expecting keyword_end
        " please contact the Autolab Development team at the " \ 
                                                                ^ (SyntaxError)

dlbucci:
Don't know why it did that...

dlbucci:
I'll look into fixing these issues later.

ymzong:
@dlbucci: The issue @icanberk mentioned was fixed -- it was because there was a trailing writespace after one of your backslashes, and Ruby doesn't like it.. :smile:

icanberk:
It's easy to fix... but I'm worried that there are a lot more of those.

ymzong:
Oh wait, that changed was caused by `rubocop`? I thought Dan changed it manually...

dlbucci:
That's why we should check every page and tes this thoroughly. I can't
believe ruby wouldn't allow that...

If only we had a systematic way to test that every page was working...

ymzong:
Well, the testing framework can kind of do that, but I still can't guarantee 100% coverage.

dlbucci:
So do we want to merge this?  It has to happen at some point and if we put any sweeping changes in before, it's just gonna turn into a bear.

icanberk:
Did anyone check if there are any more whitespaces after one of backslashes?

icanberk:
There seems to be more of the same errors

dlbucci:
There's at least one other place, I'll try to fix it now.  If we can get the server to start, should we merge?

icanberk:
Yeah I want to merge it

icanberk:
We good?

dlbucci:
Sawshark's a-running, and there don't seem to be obvious errors.

PR #392, opened 2015-03-15
Trim unnecessary routes.

ymzong:
Trimming down the following unnecessary routes:
- `POST admins#create`
- `GET admins#new`
- `GET admin#edit`
- `PATCH/PUT admin#update`
- `DELETE admin#destroy`

Tested to work on my local dev environment.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -12,7 +12,7 @@ Autolab3::Application.routes.draw do
     get 'vmlist'
   end
 
-  resource :admin do
+  resource :admin, only: :show do
     match 'emailInstructors', via: [:get, :post]
   end
 


config/routes.rb
```diff
@@ -12,7 +12,7 @@
     get 'vmlist'
   end
 
-  resource :admin do
+  resource :admin, only: :show do
```
icanberk: Where are you trimming the other routes?

ymzong: @icanberk This line suffices because having `resource :admin` implies the full RESTful routes. Meanwhile, limiting it with `only: :show` means I'm excluding the routes for `create`, `edit`, and `delete`.

You can also run a routes summary and check the differences.

icanberk: I see

dlbucci:
Why did this get closed? Also, using scope would probably be more appropriate.

icanberk:
We found out that there are more of these. @ymzong will send another PR.

ymzong:
I was thinking of combining many changes together -- one is the one in this
PR, and the other one is that we can remove `get 'admin'` in `resources
:users`.

You can go ahead and make the PR! I don't see anything else that needs
attention so far.

PR #394, opened 2015-03-15
Added check for auth back

icanberk:
Please test locally once again before merging.


app/controllers/courses_controller.rb
```diff
@@ -8,7 +8,6 @@ class CoursesController < ApplicationController
   skip_before_action :authorize_user_for_course, only: [ :index, :new, :create ]
   # if there's no course, there are no persistent announcements for that course
   skip_before_action :update_persistent_announcements, only: [ :index, :new, :create ]
-  skip_before_action :authenticate_for_action
```
dlbucci: Do we need to be skipping this anywhere?

icanberk: I do not think so. Do you any necessity?

PR #396, opened 2015-03-16
User LDAP to populate user information (instead of Shibboleth)

andrewwuan:
ldap_reset function is intended to be used in rails console for now to reset users whose information is wrong. I'll put it in UI in another PR.


ymzong:
:+1:

icanberk:
Umm I removed this a while ago because we don't want to depend on CMU specific LDAP.

Did you confirm that the wrong data comes from Shib?

PR #398, opened 2015-03-16
FUUUUUUUUUUUUUUUUUUUUUUUU

dlbucci:


jez:
Dan's PR titles and branch names are always on point.

ymzong:
That professionalism :smile_cat:

icanberk:
how did you test this fixes DosLingos?

dlbucci:
We were reuploading Makefile and autograde.tar, and they were getting overwritten as scp was occurring.  This was because our MD5 check wasn't working.  You need to use string keys to access the JSON values.

icanberk:
Okay, but how do you know this fixes DosLingos?

icanberk:
Does overwriting necessarily mean that it was breaking SCP?

icanberk:
I'm asking because I increased the max open file limit on greatwhite like 15mins ago. That might also be the reason it's fixed.

icanberk:
Though I still see lots of failing jobs.

dlbucci:
Hold on, I can explain on slack.

mihirpandya:
Please don't max open file limit. It's causing greatwhite to be really slow.

PR #402, opened 2015-03-17
Develop

icanberk:
Update do v1.4


spec/features/user_log_in_spec.rb
```diff
@@ -1,22 +1,32 @@
+<<<<<<< HEAD
```
ymzong: :smile_cat:

icanberk:
Merging in 10 mins.

PR #406, opened 2015-03-18
Fix External Tar Bug

dlbucci:


ymzong:
Did you test this? How external moss works is like this: inside `external.tar`, there can be:
- `submission1/`
- `submission2.tar`
- etc.

And, we need to add `externalTarDirectory/**/*` to MOSS command.

dlbucci:
It's tested and works, but it no longer works for nested archives.

ymzong:
@dlbucci I'll complete the fix now, if you are not working on it

dlbucci:
What's there to complete? Nested archives aren't really very feasible, and we already specify that on the moss page.

ymzong:
I want to refactor the code a bit -- the current function is too long

dlbucci:
You mean runMoss?  It could be split into a few functions.  If you want to do that, go for it, but since this is a hotfix, can we merge and do that in develop?

ymzong:
Oh definitely, go ahead!

PR #414, opened 2015-03-21
Add moment.js to extension date

### LINKED ISSUE #404: Format extension dates with moment.js
- [x] Dates like this on the assessment page should be formatted using the same moment.js formatting as other dates.

screen shot 2015-03-18 at 4 35 21 am

dlbucci:
Fixes #404

### DIFF
--- a/app/views/assessments/show.html.erb
+++ b/app/views/assessments/show.html.erb
@@ -112,7 +112,7 @@
   You have a <%=@extension.days %> day extension on this assessment
   </i>
   <font size=-1>
-  (Due date: <%= due_at_display(@assessment.aud_for(@effectiveCud.id).due_at) %>)
+    (Due date: <span class="moment-date-time"><%= due_at_display(@assessment.aud_for(@effectiveCud.id).due_at) %></span>)
   </font>
   <% end %>
   <% if @assessment.max_submissions != -1 then %>


app/views/assessments/show.html.erb
```diff
@@ -112,7 +112,7 @@
   You have a <%=@extension.days %> day extension on this assessment
   </i>
   <font size=-1>
-  (Due date: <%= due_at_display(@assessment.aud_for(@effectiveCud.id).due_at) %>)
+    (Due date: <span class="moment-date-time"><%= due_at_display(@assessment.aud_for(@effectiveCud.id).due_at) %></span>)
   </font>
```
jez: What in the world do we have a font tag for?

dlbucci: I don't know! It's as if some of our HTML is sub-par!

icanberk:
This is a hotfix but merged into `hotfix_extension_moment_date`? Is it actually a hotfix?

dlbucci:
What?  Are you asking if it belongs in master?  Cause there's no urgency, I meant for it to be in develop, but it is a bug fix.  I guess fix might have been more appropriate.

icanberk:
Yeah, we're supposed to use hotfix_ for urgent things that needs to go into master, but no probs.

PR #416, opened 2015-03-21
Hotfix archive autograde txt

### LINKED ISSUE #382: Assignment deletion doesn't delete autograder feedback
Not a big deal, but it causes the handin directory to have autograde.txts where it has no corresponding .tgzs, and if autograding failed for some reason this could lead to a confusing case where students were seeing feedback from deleted submissions. I'm not sure what the right behavior here is, but one reasonable thing would be that when

rjsimmon@andrew.cmu.edu_1_handin.tgz
gets moved to archive/deleted_rjsimmon@andrew.cmu.edu_95851_handin.tgz

rjsimmon@andrew.cmu.edu_1_editor_autograde.txt
gets moved to archive/deleted_rjsimmon@andrew.cmu.edu_95851_editor_autograde_.txt

dlbucci:
Fixes #382.
Chops up sendJob as well, so we should be careful.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -18,6 +18,10 @@ Metrics/LineLength:
     - http
     - https
 
+Metrics/MethodLength:
+  CountComments: false  # count full line comments?
+  Max: 20
+
 # Indent private/protected/public as deep as method definitions
 Style/AccessModifierIndentation:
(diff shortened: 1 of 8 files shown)


dlbucci:
Polling works, and using a callback url seems to work as well.

PR #423, opened 2015-03-23
Converted course urls to include name instead of id

icanberk:


icanberk:
Please pull and test on your local. ie. Navigate around

dlbucci:
Don't we want to put this into develop?

icanberk:
tros

Issue #421, opened 2015-03-23
Interaction between front-end and Tango needs to be factored out to a library


ymzong:
See autolab/Tango#53.

PR #425, opened 2015-03-24
Feature tango API

ymzong:
Added a Tango API to autolab `lib/`, and modified calls to Tango by the API.

Tested on local dev.


dlbucci:
This doesn't look like you branched of develop HEAD, which more or less rewrote sendJob?

ymzong:
Yep I just realized that -- merging... :disappointed:

lib/TangoClient.rb
```diff
@@ -0,0 +1,81 @@
...
+# Ruby API Client of Tango
+module TangoClient
+  # Httparty client for Tango API
+  class ClientObj
+    include HTTParty
+    base_uri "#{RESTFUL_HOST}:#{RESTFUL_PORT}"
+    default_timeout 10
+  end
+
+  def self.tango_handle_timeouts
```
dlbucci: Noice.

dlbucci:
According to rubocop, files should be snake_case (TangoClient.rb -> tango_client.rb).

ymzong:
@dlbucci Modified the file name and merged. Ready to go!

app/controllers/jobs_controller.rb
```diff
@@ -41,8 +29,8 @@ def index
     end
 
     # Get the complete lists of live and dead jobs from the server
-    raw_live_jobs = getCurrentJobs
-    raw_dead_jobs = getDeadJobs
+    raw_live_jobs = TangoClient.tango_jobs()["jobs"]
```
dlbucci: If you always read ["jobs"] from the result of tango_jobs, should we just return that value instead?

app/controllers/jobs_controller.rb
```diff
@@ -1,22 +1,10 @@
...
-    getJobs("0/")
-  end
-
-  #
-  # getDeadJobs - this function retrieves the recent dead jobs
-  #
-  def getDeadJobs
-    getJobs("1/")
-  end
+  include TangoClient
```
dlbucci: Is this line necessary?  I'm pretty sure this defines all of the methods in TangoClient on the JobsController, which is not what we want.  Simply requiring the file should give you access to the class methods.

app/controllers/assessment/autograde.rb
```diff
@@ -1,11 +1,14 @@
...
+
 ##
 # Contains all functions related to autograding, including functions to send autograding jobs,
 #   the callback route for when jobs are completed, and regrading routes.
 # Gets imported into AssessmentsController
 #
 module AssessmentAutograde
-  require "uri"
-  require_relative Rails.root.join("config", "autogradeConfig.rb")
+  include TangoClient
```
dlbucci: Don't think this is necessary.

dlbucci:
Have you run rubocop on this code?

app/controllers/jobs_controller.rb
ymzong: Tango returns a full json, with not only `jobs` but also `statusId`. Although we don't use it here, I thought it might be good to just have the full json and do whatever with it, just like the other endpoints.

app/controllers/assessment/autograde.rb
ymzong: Good catch! Those are leftovers of my other attempt. Fixed.

app/controllers/jobs_controller.rb
dlbucci: Could we check the statusId in TangoClient, and return a null or empty object if it's a failure status?  I think part of the reason we want this TangoClient is so that Tango details are abstracted away from Autolab, so that multiple backends can be supported.  So to me, it makes sense that the jobs function would always return an object describing the jobs.
That said, this is just a step toward a fully-abstracted TangoClient, so if you don't agree or don't want that for this commit, I understand.

ymzong:
I ran rubocop on `lib/tango_client.rb` and `app/controllers/assessment/autograde.rb`.

app/controllers/jobs_controller.rb
ymzong: I'm not sure -- @icanberk?

icanberk: I think it's a good idea to acknowledge error cases. Null would be good, but how about raising exceptions with an explanatory text? That way we can just do begin/rescue and also display human readable errors.

dlbucci: Yeah, an exception would work as well.

ymzong:
@dlbucci @icanberk 
The exception has been handled -- essentially, whenever the response times out / gives socket error / have negative `statusId` (when available), `TangoClient` would raise `TangoClient::TangoException` along with a corresponding message.

Calls to `TangoClient` are also wrapped with `begin...rescue` blocks, adding error messages to `flash[:error]`.

Since the code is rather critical, I recommend that someone test it fully on staging before merging. :smile_cat:

dlbucci:
Okay,  but wasn't the point of the exceptions so that we didn't have to return the statusId, and could instead return the meaningful part of the response from Tango?  That way we don't have to append ["jobs"] or ["files"] to calls to TangoClient functions?

ymzong:
@dlbucci Oops I forgot about it -- fixed!

lib/tango_client.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  def self.tango_prealloc(image, num, options = {})
+    tango_handle_exceptions do
+      url = "/prealloc/#{api_key}/#{image}/#{num}/"
+      ClientObj.get(url, body: options)
+    end
+  end
+
+  def self.api_key
```
dlbucci: Is there a reason for this function?  I'm not sure it would be any less clean to use RESTFUL_KEY instead.

dlbucci:
Do we want to prefix every function with tango_?  I mean, they all already belong to the TangoClient module.

icanberk:
I think we should `tango_` prefixes and then this PR is good to go. @ymzong

lib/tango_client.rb
```diff
@@ -0,0 +1,90 @@
...
+  # Exception for Tango API Client
+  class TangoException < StandardError; end
+
+  def self.tango_handle_exceptions
+    resp = yield
+    if resp.content_type == "application/json" && resp["statusId"] && resp["statusId"] < 0
+      fail TangoException, "Tango returned negative status code."
+    end
+    return resp
+  rescue Net::OpenTimeout, Net::ReadTimeout
```
dlbucci: Is there a set timeout limit? Or does this occur when we take Tango down and then try to hit it?

ymzong: The `timeout` is set on line 11, so whenever we make an API call to Tango, we expect it to finish in 10 seconds.

ymzong:
@icanberk I just cleaned up the method names. Again, we need to test this on staging before merging, otherwise...

ymzong:
(fixes #421)

icanberk:
Pulled to sawshark.

PR #428, opened 2015-03-25
Added test for autograding (half-) round-trip.

ymzong:
This test submits a file for the autograded assessment, and verifies that the job completes in a reasonable amount of time.

Note: Currently a full round-trip is not possible because the callback cannot reach the VMs for Travis CI.

### DIFF
--- a/lib/tasks/autolab.rake
+++ b/lib/tasks/autolab.rake
@@ -285,6 +285,14 @@ namespace :autolab do
       FileUtils.mkdir_p(File.join(course_dir, a.name, a.handin_directory))
     end
 
+    # Load autograding properties
+    autograde_prop = AutogradingSetup.new
+    autograde_prop.assessment_id = asmt.id
+    autograde_prop.autograde_image = "rhel.img"
+    autograde_prop.autograde_timeout = 180
+    autograde_prop.release_score = true
+    autograde_prop.save!
+
     # Load problem "autograded"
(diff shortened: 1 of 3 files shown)


icanberk:
How do we setup Tango on Travis?

Can we do polling to get the output from the VM?

ymzong:
@icanberk I'm not sure about how to set up Tango and let it run on a single machine, or if it's currently possible at all. (@mihirpandya ?)

I think we can get the score/feedback by polling, but I also heard some rumor that polling is not working (@mihirpandya ?).

By the way, do you think it's a good idea to test both frontend and Tango in the frontend Travis test suite? I'm rather neutral on this.

icanberk:
I don't think it's actually a good idea, but I don't understand how you successfully submit a file for autograding?

icanberk:
Also, we should test autograde_done method with a hardcoded json string.

ymzong:
If you look at lines 80-83 [here](https://travis-ci.org/autolab/Autolab/builds/55756453), you'd notice I set up several "secret" environment vars that point to our production Tango. Since submitting a job is one-way network connection, we can do it on Travis. However, callbacks cannot reach Travis unfortunately.

I agree, we should have an `autogradeDone` callback to `localhost` on Travis to stimulate a callback. Meanwhile, we should test the actual feedback stuffs in Tango test. What do you think?

PR #429, opened 2015-03-25
Rubocop'd A models

dlbucci:


jez:
lol "A models"

dlbucci:
What's funny about that?

jez:
I was like what's an A model and the. I realized Oh, it's literally just
_everything that starts with an a_ :P

dlbucci:
I think I know why this isn't building... I renamed "Statistics.rb" to "statistics.rb", but the OS X file system is case-insensitive, so it didn't detect the change.  So the file is still called "Statistics.rb" in the repo, despite being "statistics.rb" on my machine...

PR #433, opened 2015-03-28
Move AutogradingSetup into its own controller

dlbucci:
Don't think this is ready for primetime yet, but what do you guys think?


jez:
What's the idea behind this feature?

dlbucci:
Clean up, mostly. Our assessments_controller is HUGE, and autograding settings code really doesn't belong there, so this scales it back.

icanberk:
what if we got rid of autograding_setups completely and add columns to assessment table?

Current attributes are:
```
    t.integer "assessment_id",     limit: 4
    t.integer "autograde_timeout", limit: 4
    t.string  "autograde_image",   limit: 255
    t.boolean "release_score",     limit: 1
```

dlbucci:
I think the assessments table is complex enough already, plus more than half of our assessments don't use autograding, so that would be a bit of wasted space.

dlbucci:
This should be ready for showtime/testing.

icanberk:
I pulled this on sawshark and migrated. Seems like autograde roundtrip is working.

PR #436, opened 2015-03-28
Adds New Relic support for front-end

ymzong:


icanberk:
So what is this gonna do?

ymzong:
This is a trivial change that adds `newrelic_rpm` gem and a template config file for New Relic. With this, any deployment manager can monitor what's going on with their own Autolab instance on New Relic.

icanberk:
I was just wondering what New Relic will tell us

PR #437, opened 2015-03-28
Initial implementation of Public Course Instances

icanberk:
There are two options:
1) Is class public: If the class is discoverable by others.
2) Does class require permission: If the class is public and doesn't require permission, the user can just join the class. : If the class is public but does require permission, the user requests to join the class and the instructor needs to confirm the user.

We are creating` cudRequest` to keep track of join requests.


dlbucci:
I'd rather add a column to the cud table to the effect of 'has_joined'.  This CudRequest class just seems to add too much complexity.

icanberk:
I considered that. But I feel like that'll add more complexity because whenever we need to query CUDs we will need to make sure that we only get 'has_joined  = true'. 

Current implementation adds an extra table, but keeps everything else pretty simple imho.

What do you think?

app/controllers/courses_controller.rb
```diff
@@ -5,25 +5,59 @@
...
+    unless @course
+      redirect_to(controller: :home, action: :error) && return
+    end
+
+    if not current_user.nil?
+
+      cud = @course.course_user_data.find_by(user_id: current_user.id)
+
+      if not cud.nil?
+        redirect_to course_assessments_path(@course)
```
icanberk: @dlbucci: for some reason, when I do  `redirect_to course_assessments_path(@course) && return` it doesn't return and cause a bug. any idea why? it seems to be the style that rubocop prefers.

dlbucci: I think you need to wrap the URL in parentheses. Otherwise, the url gets combined with the return before the redirect is called.

dlbucci:
We could put a default scope on CUDs so that by default, only joined CUDs are loaded.

icanberk:
Is there a sample of it in the codebase that I can use?

icanberk:
@dlbucci halp

dlbucci:
Scores used to do it.  Just look here: http://api.rubyonrails.org/classes/ActiveRecord/Scoping/Default/ClassMethods.html#method-i-default_scope
If you use that method, whatever query you enter will be applied by default whenever you load that model.

icanberk:
scoping is pretty cool

icanberk:
Apparently when you do, `@course.course_user_data.unscoped` it disregards the relation between cud and course, and looks at all the cuds.

dlbucci:
CourseUserData.unscope.where(course: @course)?

icanberk:
I know it works. Just wanted to note it down because it took me a while to realize that was the case.

db/schema.rb
```diff
@@ -68,6 +68,7 @@
     t.datetime "updated_at"
     t.integer  "course_id",          limit: 4
     t.string   "display_name",       limit: 255
+    t.integer  "category_id",        limit: 4
```
dlbucci: Can you update your database so that this doesn't show up?  Running a rake db:reset with the current develop db/schema.rb might do it, but it's really annoying that this non-existent column keeps showing up.

db/migrate/20150403173410_add_has_joined_to_cud.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def up
+
+  	add_column :courses, :public, :boolean
+    add_column :courses, :requires_permission, :boolean
+    add_column :courses, :website_url, :string
+
+    add_column :course_user_data, :has_joined, :boolean
+
+    CourseUserDatum.all.each do |cud|
```
dlbucci: Use 'find_each' here.  Any reasons some lines in this file are 3-indented/not aligned?

dlbucci:
Make sure you run rubocop --auto-correct on this.  You don't need to fix everything it shows, but just make sure that the parts of the code that you touch are all good (ignoring method length, complexity issues, etc.).

app/controllers/application_controller.rb
```diff
@@ -19,6 +19,8 @@ class ApplicationController < ActionController::Base
   before_action :update_persistent_announcements
   before_action :set_breadcrumbs
 
+  helper_method :resource_name, :resource_class, :resource, :devise_mapping
```
dlbucci: Are these being used by devise and if so, do they need to have these names?  If that's the case, maybe leave a comment explaining that.

app/controllers/course_user_data_controller.rb
```diff
@@ -136,13 +139,21 @@ def update
...
   end
 
-  action_auth_level :destroy, :instructor
+  action_auth_no_course :destroy
   def destroy
-    @destroyCUD = @course.course_user_data.find(params[:id])
-    if @destroyCUD && @destroyCUD != @cud && params[:yes1] && params[:yes2] && params[:yes3]
-      @destroyCUD.destroy # awwww!!!
+    @course = Course.find_by(name: params[:course_name])
+    @cud = CourseUserDatum.unscoped.find(params[:id])
```
dlbucci: Do we need to reload the course here?  It should already be defined.  Also this code doesn't make sure that @cud is actually in @course.

app/controllers/course_user_data_controller.rb
```diff
@@ -136,13 +139,21 @@ def update
...
+
+    if @cud.instructor?
+      @destroyCUD = @course.course_user_data.find(params[:id])
```
dlbucci: I think this code is incorrect.  It's using params[:id] to load 2 different cuds, but really it loads the same cud twice.  We shouldn't ever load @cud, because it gets defined by session storage and represents the actual user, not any models that we're messing with.

app/controllers/course_user_data_controller.rb
```diff
@@ -198,6 +209,15 @@ def unsudo
     redirect_to([@cud.course]) && return
   end
 
+
+  action_auth_level :confirm, :instructor
+  def confirm
+    @student_cud = CourseUserDatum.unscoped.find(params[:id])
```
dlbucci: This should be scoped under the course.

app/controllers/course_user_data_controller.rb
```diff
@@ -207,18 +227,18 @@ def add_users_breadcrumb
...
                                                 tweak_attributes: [:_destroy, :kind, :value])
-    elsif @cud.instructor?
+    elsif @cud && @cud.instructor?
       params.require(:course_user_datum).permit(:school, :major, :year,
                                                 :lecture, :section, :instructor, :dropped, :nickname, :course_assistant,
                                                 user_attributes: [:email, :first_name, :last_name],
                                                 tweak_attributes: [:_destroy, :kind, :value])
     else
-      params.require(:course_user_datum).permit(:nickname) # ,
+      # params.require(:course_user_datum).permit(user_attributes: [:email, :first_name, :last_name]) # ,
```
dlbucci: Why is this commented out?

app/controllers/course_user_data_controller.rb
```diff
@@ -207,18 +227,18 @@ def add_users_breadcrumb
   end
 
   def cud_params
-    if @cud.administrator?
+    if @cud && @cud.administrator?
```
dlbucci: Why are the @cud's being nil checked?  Will @cud ever not be defined in this controller?  Also, as a minor gripe, we should check for instructors before administrators, because the latter will require a user-load from the DB.  Why are we splitting the attributes if they are the exact same object anyway?

app/views/courses/_courseFields.html.erb
```diff
@@ -11,6 +11,15 @@
   help_text: "Descriptive course name for displaying. Example: 15-213: Intro to
   Computer Systems" %>
 
+<%= f.text_field :website_url,
+  help_text: "URL to the external course webside" %>
```
dlbucci: 'webside' -> 'website'

app/controllers/courses_controller.rb
```diff
@@ -5,25 +5,51 @@
...
-    if courses_for_user.any?
+    if not courses_for_user.nil?
       @listing = categorize_courses_for_listing courses_for_user
     else
-      redirect_to(home_no_user_path) && return
+      @listing = []
     end
 
-    render layout: "home"
+    # render layout: "home"
```
dlbucci: Dead code.  Why was the index page using the home layout?

app/controllers/courses_controller.rb
```diff
@@ -5,25 +5,51 @@
...
   end
 
-  action_auth_level :show, :student
+  action_no_auth :show
   def show
-    redirect_to course_assessments_url(@course)
+
+    @course = Course.find_by(name: params[:name])
+
+    unless @course
```
dlbucci: Use the modifier form here:
`redirect_to(controller: :home, action: :error) && return unless @course`

app/controllers/courses_controller.rb
```diff
@@ -5,25 +5,51 @@
...
+        return
+      else
+        @newCUD = @course.course_user_data.new
+        @newCUD.user = current_user
+        @newCUD.tweak = Tweak.new
+      end
+
+    end
+
+    if not @course.public?
```
dlbucci: Use the modifier form.

app/controllers/courses_controller.rb
```diff
@@ -5,25 +5,51 @@
...
   def show
-    redirect_to course_assessments_url(@course)
+
+    @course = Course.find_by(name: params[:name])
+
+    unless @course
+      redirect_to(controller: :home, action: :error) && return
+    end
+
+    if not current_user.nil?
```
dlbucci: Use unless here.

app/controllers/courses_controller.rb
```diff
@@ -5,25 +5,51 @@
...
-      redirect_to(home_no_user_path) && return
+      @listing = []
     end
 
-    render layout: "home"
+    # render layout: "home"
   end
 
-  action_auth_level :show, :student
+  action_no_auth :show
```
dlbucci: I don't think this should be action_no_auth.  I think users should at least have to create an account and log in before they can start looking at courses.

app/controllers/course_user_data_controller.rb
icanberk: We need to reload the course since we have `action_auth_no_course`. Fixing the checking course issue.

app/controllers/course_user_data_controller.rb
icanberk: You're right. `params[:id]` is the cud id of the student. the previous one should be using the current_user to get the cud.

app/controllers/courses_controller.rb
icanberk: I at least want people to see the course title and the link to the course website in the future. But instead of a "Join" button we show a sign up/sign in link for them.

I can see a lot of cases where this is useful. e.g. Teacher can just send one link and tell people to sign up, instead of saying "you need to sign up for Autolab first, then go to this link and join the course"

db/migrate/20150403173410_add_has_joined_to_cud.rb
```diff
@@ -0,0 +1,27 @@
+class AddHasJoinedToCud < ActiveRecord::Migration
+
+  def up
+
+    add_column :courses, :public, :boolean
+    add_column :courses, :requires_permission, :boolean
```
dlbucci: I'd imagine you would want to set public to false and requires_permission to true for both of these columns.  (Maybe make that the default as well?)

And unrelated, but something you should know, adding three columns like this essentially copies the table three times, each time with one added column.  Since `:courses` is very small, it doesn't matter, don't change it, but for something like `:scores`, you should create a new table with all three columns added, and then copy rows, one by one.

app/views/courses/show.html.erb
```diff
@@ -0,0 +1,28 @@
+<h1><%= @course.display_name %></h1>
+<hr>
+<% if current_user.nil? %>
+<div>You need to sign in to be able to join this class.</div>
+<br>
+<div class="sign-in-panel">
+<% if Rails.env == "development" %>
+<%= link_to "Developer Login", "/home/developer_login" %><br />
+<% end %>
+<%= render "devise/shared/links" %>
```
dlbucci: Is this the only place where we're doing this?  If so, it seems like a lot of weird, non-obvious code to achieve what a single link to a sign-in page would do.

icanberk: We will likely to need the same thing in couple other places in the future.

db/schema.rb
icanberk: Migrations are trying to scope cud for `has_joined` and can't find the column now...

dlbucci: Dang.  I'm thinking the default_scope might be a bad idea, and instead we should have a named scope like `joined`.  This is based on me reading some SO answers that said default_scope is a bad idea.  How many places do we load CUDs, and would it be hard to add a scope to those places?

app/views/courses/show.html.erb
dlbucci: Like where?  I'd really rather have users get directed to a sign up page, because then we can keep a unified course loading system, and not duplicate code.

icanberk: It's really not that much of extra code. I want to keep this ability.

dlbucci: It's less that it's a lot of code and more that we're duplicating it.  For a while, we were loading assessments and submissions in different ways depending on the controller/action, which means you had inconsistencies (students could access unreleased assessments, for example).

icanberk: There is 0 logic here. We're just rendering one line. (We don't even need the `Rails.env == "development"` part. That can probably be put into that partial view.

icanberk:
@dlbucci are you happy with this?

Can you pull and test the migration as well?

dlbucci:
I'll take a look later today.

icanberk:
ping

config/routes.rb
```diff
@@ -127,6 +141,7 @@
       end
 
       member do
+        get "confirm"
```
dlbucci: This should be a post route.

icanberk:
@dlbucci fixed

dlbucci:
I get an error in self.courses_for_user if I hit the "/courses" while not logged in.  Also, there's a courses link on sign in page, but it goes to "/" and not "/courses".

icanberk:
We no longer have that link...

config/routes.rb
```diff
@@ -1,9 +1,21 @@
...
+
+    authenticated :user do
+      root 'courses#index', as: :authenticated_root
+    end
+
+    unauthenticated do
+      root 'devise/sessions#new', as: :unauthenticated_root
+    end
+
+  end
```
dlbucci: What is this block doing?  It seems to create 2 different routes that have the same path.  The result is that we get two different route helper functions pointing to the same route, but I think anyone who hits "/" will always be directed to only one of the actions.

icanberk: It is just that the same route acts differently depending on whether the user is authenticated. We can't have them in one controller because 'devise/sessions#new' code is part of the devise gem.

icanberk:
Can get this in? It has been 24 days

dlbucci:
I still don't think this is ready.  I don't see the point of the devise template, I think we should just require people to sign in and sign up before they can view anything.

app/controllers/application_controller.rb
```diff
@@ -63,13 +63,18 @@ def self.action_auth_level(action, level)
 
   def self.action_no_auth(action)
     skip_before_action :verify_authenticity_token, :authenticate_user!
-    skip_filter configure_permitted_paramters: [action]
     skip_filter maintenance_mode: [action]
     skip_filter run_scheduler: [action]
 
     skip_filter authenticate_user: [action]
     skip_before_action :authorize_user_for_course, only: [action]
```
dlbucci: I think we should break this up into 2 before_actions: set_course loads @course, and set_cud loads @cud/makes sure that the user is authorized to view the course.

icanberk:
Devise template was already there. I believe it comes with the gem.

The whole point of this feature is so that an instructor can send out a link to the course and tell people to signup. People should be able to see what they are signing up for when they receive that link.

icanberk:
Secondly, please be more considerate when you're reviewing things. It's a big waste of time for me to go back and fort with this. Instead of nitpicking on one small thing every time, you can thoroughly review it once or contribute to the branch if something is really bothering you.

app/controllers/application_controller.rb
icanberk: Good idea. We can do this in a separate PR.

dlbucci:
This isn't one small thing, it's a real issue, and I don't want this the code base the way its currently implemented.  The show page does nothing when the user isn't signed in, besides show the name of the course, which could easily be added to a flash message on the sign-in page.  This would let us keep our unified course-load code path, which helps prevent bugs and other issues.  I'm not seeing any benefit from this devise template, especially for how messy it makes the code.
If you want me to work on this, I can, just let me know.

icanberk:
Show page does say that this course in fact exists, it's public and the user can join it if s/he signs up for the service. When the instructor puts the external website link, we'll also display it on this page and maybe also add a course description to make this page even more useful even if the user is not logged in. I know we can show a flash message and redirect to homepage, but it's not the same thing. This is a user experience choice and I want it to be this way. I also don't think it makes the codebase uglier than it currently is but if you see a better way of displaying the same page with cleaner code, please help me.

icanberk:
Guys, if you don't mind, I'd like to merge this and do design changes based on the user feedback.

dlbucci:
Let me take a look at this some time this week and see what I think/can do.  We really don't need to rush to get this in.  If it's there by fall, we'll be good.

icanberk:
We could finish this in the Spring and test for summer classes. There is no such thing as "we don't need to rush". If someone took the time to write it, we should try to get it in as soon as possible

jez:
Ilter,

I admire your passion to add this feature to Autolab. In the long run, I think this will offer a significant benefit to users of Autolab.

However, the real issue here is whether this is the particular implementation of this feature we want to add to the codebase. Rereading the conversation above, it is clear that Dan has expressed some concerns about the design decisions taken in creating this feature as they relate to the overall quality of the codebase.

To reconcile these two things (a desire to land this feature, and the need for high quality code), I don't think that getting this change landed as soon as possible is the correct approach. Rather, I think we should start our conversation over, evaluate the specific points of the implementation that are contentious, and work towards a solution that everyone can agree to.

Thanks for spearheading this feature. @icanberk and @dlbucci, do you think you can work out the details from here?

icanberk:
@jez, I don't think @dlbucci has any more technical issues. I'd be happy to fix them if there are. We tried 3 different implementation for this feature and this seems to work best. I think his concern was more on the design side. (Should the visitors be able to see the course name, url and some metadata in a nicely formatted way if they're not signed in? I definitely think they should, even if it costs 30 lines of extra code)

In the meantime, I think both Dan and I are grown-up enough to debate things on our own, in fact we sort of enjoy this and I'll actually miss it :) I'd rather have you contribute to the discussion instead of writing long mediatory comments.

Trying to get a PR in 3 months after the submission is not really rushing. I'm just concerned that we're not having any progress here.

dlbucci:
@icanberk Actually, I think the fact that this fragments/duplicates our model loading path is a serious technical issue, and the main reason I'd rather not merge this in its current form.  I'd like to take one more look over this to see if there's a way to code this without having to do that.

icanberk:
@dlbucci Any solutions so far?

icanberk:
@dlbucci ping

I don't think we're fragmenting our model loading path. We're just creating a new way called `self.action_auth_no_course` and using it at one place. Still if you have a better solution, I'm totally okay with it.

icanberk:
I would like to push a new version of Autolab including this, Docker, the annotations and the small stuff we've been working on over the summer. I will be merging this on Friday if no one has better suggestions.

icanberk:
@jez and @ymzong, what do you guys think?

ymzong:
@icanberk I'd say we merge this now before conflicts go crazy -- we can always fix things later if necessary :smile_cat:

Issue #397, opened 2015-03-16
`importAsmtFromTar` method should use the new archive module


icanberk:
It's too complex as it is.

ymzong:
I'm on it :+1:

icanberk:
I think this will improve our GPA at least 0.1. Just saying :santa:

ymzong:
I'm on it!

remzr7:
Merged

PR #438, opened 2015-03-29
Refactor long functions in asmt,courses controller

ymzong:
This code mainly refactors long functions in `AsmtController` and `CoursesController`. 

Regarding #397: This PR only refactors the current code instead of switching to `Archive` library in `#importAsmtFromTar`. This is because the library would not make the code cleaner, and instead I would need to pre-process import tarball in order to use the library. I'd like to fix all `Archive` issues together in another PR.

### DIFF
--- a/app/controllers/assessments_controller.rb
+++ b/app/controllers/assessments_controller.rb
@@ -54,7 +54,6 @@ class AssessmentsController < ApplicationController
   # SVN
   autolab_require Rails.root.join("app", "controllers", "assessment", "SVN.rb")
   include AssessmentSVN
-
   action_auth_level :adminSVN, :instructor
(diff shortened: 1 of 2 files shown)


dlbucci:
Did you run rubocop on these files?  You don't have to clean the entire files, but it would help cleanup the functions you're refactoring.

ymzong:
@dlbucci No I didn't run rubocop on the files. How would I run it on individual functions?

Also, probably you can do `rubocop`ping in a separate PR?

dlbucci:
You can run it on the files, but only correct it in the functions you change.

ymzong:
Just `rubocopped`!

PR #443, opened 2015-03-30
Add Groups Links back to ass#show

dlbucci:
These got overwritten in a develop merge and we didn't notice for 6 weeks.


icanberk:
Pulled

PR #445, opened 2015-04-01
Rubocop Fixes

dlbucci:
Under 1000 issues again!


ymzong:
Thanks for removing `user_module_spec.rb`! :smile_cat:

PR #449, opened 2015-04-06
Renamed display_name helper to full_name

icanberk:
In the edit course page, :display_name field was being filled by the display_name helper method instead of the display_name field. Therefore, the display_name field would get updated with the new suffix every time the edit for mis submitted. Renaming the helper method to `full_name` fixes this problem. That was the best name I could come up with... display_name_with_sem would have been too log. but lmk if you have a better idea.

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -62,7 +62,7 @@ class Course < ActiveRecord::Base
     assessments.where("start_at < :now AND end_at > :now", now: now)
   end
 
-  def display_name
+  def full_name
     if self[:semester].to_s.size > 0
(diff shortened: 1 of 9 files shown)


app/models/course.rb
```diff
@@ -62,7 +62,7 @@ def current_assessments(now = DateTime.now)
     assessments.where("start_at < :now AND end_at > :now", now: now)
   end
 
-  def display_name
+  def full_name
```
dlbucci: Maybe check this with rubocop, but I think we should just refer to everything as an instance variable here (self[:semester] -> semester, etc).  I don't think we have code anywhere else that's doing anything like this.

PR #450, opened 2015-04-06
Autolab full round-trip test working!

ymzong:
How the round-trip test works:
- `Capybara` logs in user, goes to assessment, and submits a file;
- Verifies that the jobs page contains the job, and 15 seconds later the job completes;
- Since Travis CI is an isolated environment, callback from Tango cannot be directly issued -- now Capybara grabs the callback URL on the jobs page and manually sends a POST request with a feedback file we created;
- Within five seconds the new score should be on "Submission History" page.

The issue that blocked me for a week was to send the feedback file by multi-part POST request via Capybara. After reading countless number of pages, [this one](http://seejohncode.com/2012/04/29/quick-tip-testing-multipart-uploads-with-rspec/) saved my life. Thanks @seejohnrun!!
.

### DIFF
--- a/spec/features/autograding_roundtrip_spec.rb
+++ b/spec/features/autograding_roundtrip_spec.rb
@@ -1,4 +1,6 @@
 require "tempfile"
+require "uri"
+require "httparty"
 
 RSpec.describe "autograding", type: :feature do
   it "runs through successfully" do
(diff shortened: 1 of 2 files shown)


icanberk:
Cool!

Issue #289, opened 2015-02-17
Scoreboards need numbered rows


droh:
Somewhere in the transition to the new system, scoreboards lost their numbered rows (starting at 1 on the lefthand side). These are very helpful to instructors and should be reinstated.

jez:
After looking into this issue, some poor attempts at modularization are preventing this from being the simple change that I thought it would. It turns out that this will involve changing a method that can right now be overridden in an assessment's config file. Once we've cleaned up config files and have a way to warn instructors about changes to them (#227), this becomes a 10-line fix.

PR #454, opened 2015-04-11
Feature scoreboard model

dlbucci:
THIS IS NOT READY, DO NOT MERGE.

Just sending this to get a feel for what people think.  This will require changes to config files, but 213 is the only course that uses any of these functions.


app/models/scoreboard.rb
```diff
@@ -1,7 +1,10 @@
-class ScoreboardSetup < ActiveRecord::Base
```
icanberk: I'd have preferred `ScoreboardSetup`. When you say Scoreboard I expect something that saves the scores of the users etc.

dlbucci: This is more in line with the Autograder, and it makes more sense from a modeling perspective.  An Assessment has a Scoreboard or doesn't, and if it has a Scoreboard, then that scoreboard's settings will be in the Scoreboard model.  Plus, this is shorter, which is a plus in my book.

droh: We'll adapt 213 config files to whatever you think best. So don't let 213
stand in the way of improving the scoreboard code.

On Saturday, April 11, 2015, Daniel Bucci <notifications@github.com> wrote:

> In app/models/scoreboard.rb
> <https://github.com/autolab/Autolab/pull/454#discussion_r28196609>:
>
> > @@ -1,7 +1,10 @@
> > -class ScoreboardSetup < ActiveRecord::Base
>
> This is more in line with the Autograder, and it makes more sense from a
> modeling perspective. An Assessment has a Scoreboard or doesn't, and if it
> has a Scoreboard, then that scoreboard's settings will be in the Scoreboard
> model. Plus, this is shorter, which is a plus in my book.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/autolab/Autolab/pull/454/files#r28196609>.
>


-- 
Typed with fat fingers from my iPhone.

dlbucci: It won't.  I've already change the scoreboardHeader spec, so we'll need to update those calls, but only the 213 assignments and one other assignment use that, so we'll be fine.

dlbucci:
So I think this is about ready to merge.  People should pull and test though, this turned into a huge PR.

jez:
So I see you fixed #289: does this mean we've fixed all that we needed to regarding refactoring config files, or did you just happen to be able to fix that issue?

dlbucci:
It means we're gonna need to alter about 8 assessment config files, but besides those, we should be fine.

PR #456, opened 2015-04-11
Fixed some navbar links

### LINKED ISSUE #455: Allow an instructor/TA to see both gradebook & section Gradebook
Currently if you have instructor permissions and TA permissions (e.g. section K). You see a section K gradebook tab, but not the general gradebook tab as well - would it be possible to put both side by side?

dlbucci:
Fixes #455 
Changes/"cleans" some of the logic.  Instructors who are course assistants now get two gradebook links, one for the section gradebook, and one for the course gradebook.  Autolab admins now have a "Manage Autolab" link on every page.

Do we need to have a "Courses" link when the Autolab header goes to the same place?


icanberk:
We can get rid of the "courses"

PR #457, opened 2015-04-12
Develop

icanberk:
Let's plan to merge this afternoon?

It'd be great if we can merge the public classes feature. It's been sitting for too long.


dlbucci:
We shouldn't merge until the scoreboard stuff is ready.  I noticed a few autograder bugs that I fixed in that branch, and I think that they should not be put into production.

PR #469, opened 2015-04-18
Remove CMU omniauth keys from devise config

jez:
This removes the CMU-specific keys from the devise.rb.template.

We should probably regenerate these keys and change them in the production
deployment of Autolab.


jez:
Hmm, removing those lines broke the build. @ymzong, does the devise configuration affect the Travis build? If so, we would ideally expose those keys to Travis only (and not the whole open source repo).

dlbucci:
You can't instantiate User models without a valid Devise configuration, so we'll need to revise our build process.

ymzong:
Travis (actually Rails) uses Devise to create a new user, so a valid devise config is necessary. Currently what happens is that we do `cp config/initializers/devise.rb.template config/initializers/devise.rb` in Travis, so that Travis has a valid config.

And yes we should definitely remove the secret keys from the public repo (I don't know why we included it in the first place). The workaround for this is to have a separate private repo holding the secret files, pull those into Travis along with Autolab codebase, then run the tests (see `http://docs.travis-ci.com/user/private-dependencies/`).

jez:
Well we're in luck, we do have a private repo!

ymzong:
@jez I'm gonna take over this PR and modify `.travis.yml` on top of it, such that build can pass.

jez:
Alright, keep me posted.

ymzong:
Awesome, it works now @jez !

jez:
Nice!

PR #477, opened 2015-04-22
Develop

icanberk:


app/models/assessment.rb
```diff
@@ -108,11 +114,11 @@ def problem_id_to_name
   end
 
   def latest_submission_by(cud)
-    assessment_user_data.where(course_user_datum: cud).first.latest_submission
+    assessment_user_data.find_by(course_user_datum: cud).latest_submission
```
ymzong: #492 ?

ymzong: Possibly related to #492?

PR #480, opened 2015-04-24
Show Recipients of Admin Emails

dlbucci:
The @cuds getting selected looked like they were correct, but this gives us a sanity check.


jez:
I tested this and it doesn't look like it sorts the emails alphabetically. This would be a good feature to have if we're using the list of instructor emails as a way to detect duplicates.

dlbucci:
Good idea, I will add that.

PR #481, opened 2015-04-25
New footer with github and fb links

icanberk:


ymzong:
:+1:

app/views/layouts/application.html.erb
```diff
@@ -27,6 +27,15 @@
...
 
   <body class="autolab">
+    <div id="fb-root"></div>
+    <script>(function(d, s, id) {
+      var js, fjs = d.getElementsByTagName(s)[0];
+      if (d.getElementById(id)) return;
+      js = d.createElement(s); js.id = id;
+      js.src = "//connect.facebook.net/en_US/sdk.js#xfbml=1&version=v2.3&appId=259908887444831";
+      fjs.parentNode.insertBefore(js, fjs);
+    }(document, 'script', 'facebook-jssdk'));</script>
```
dlbucci: What is this code doing?

jez: Looks like it adds this:

screenshot 2015-04-25 21 52 31

dlbucci: Ah, okay, I guess that's what 'fb link' would probably mean...

jez: Since it's JavaScript, it should be in its own file though, not in application.html.erb

dlbucci: I'm actually okay with this, seeing as it's one less file to load.  I would put it at the bottom of the body tag, or if you're going to move it into a file, make it a file that contains code for every dependency (github and facebook and heap, etc.).

PR #485, opened 2015-04-27
Concat validation errors and show as flash warning

icanberk:


app/controllers/course_user_data_controller.rb
```diff
@@ -106,7 +106,8 @@ def update
         if @editCUD.save
           redirect_to(action: :show) && return
         else
-          redirect_to(action: :edit) && return
+          @editCUD.tweak ||= Tweak.new
+          render :edit
```
dlbucci: For the record, I don't really like that this doesn't redirect on a POST request, but for the time being, I think it's the best solution we have.

icanberk: Do you thin it is better to do `@cud.errors.full_messages` like we did in application controller?

dlbucci: Eh, it would be more consistent, but I'm not sure it's super important.

PR #489, opened 2015-04-28
Add support for deploying Autolab with Docker

jez:
This feature adds a number of files that are necessary to run Autolab using
Docker. In particular, many snippets have been added that make configuring
Autolab either easier to automate or easier to customize by the system
administrator.

Apart from config files, there are noteworthy changes:

- a devise view was updated to prevent crashes that came up during deployment
- a Rake task was added that automates the processor of granding a user root
  priviledges

All documentation for the features added in this commit can be found in the
wiki page entitled "Deploying Autolab with Docker".


jez:
I mostly want to get these changes in so that I can shift focus to making some small UI and consistency changes based off of my Docker deployment that simulate the process a new user would go through. Let me know if you have any questions. If you have a spare server lying around, it only takes about 30 minutes to get everything (minus SSL, which you have to wait a bit for) up and running!

config/environments/production.template.rb
```diff
@@ -71,10 +71,24 @@
...
+  #config.action_mailer.smtp_settings = {
+  #  address:              'smtp.mandrillapp.com',
+  #  port:                 25,
+  #  enable_starttls_auto: true,
+  #  authentication:       'login',
+  #  user_name:            'MANDRILL_USERNAME',
+  #  password:             'MANDRILL_PASSWORD',
+  #  domain:               'example.com',
+  #}
+
```
dlbucci: What's this stuff doing?  Is this related to Docker?

lib/tasks/admin.rake
```diff
@@ -0,0 +1,15 @@
+namespace :admin do
```
dlbucci: Is this getting used somewhere?  It's nice to have, but I don't see it getting called.

dlbucci:
This looks pretty good, I'm gonna download docker and see if I can get this running on my laptop, then I'll merge.

config/environments/production.template.rb
jez: I added some stock (optional) configuration to make it easier for people to add support for a custom mailer. You can see how to use it [here][1] and the [documentation I got it from][2], that @andrewwuan (I think) initially wrote.

[1]: https://github.com/autolab/Autolab/wiki/Deploying-Autolab-with-Docker#71-configure-email-with-mandrill
[2]: https://github.com/autolab/Autolab/wiki/Adding-email-support

lib/tasks/admin.rake
jez: This is a command line task that an admin can run from the command line whenever they want.

It's documented [here][1], and it aims to replace the instructions that were initially written [here][2] that required people to use `irb` to enter the code themselves.

[1]: https://github.com/autolab/Autolab/wiki/Deploying-Autolab-with-Docker#13-create-an-administrator-account
[2]: https://github.com/autolab/Autolab/wiki/Deploying-Autolab-on-Ubuntu#optional-setup-administrator

jez:
Sure. I didn't test it on anything but an Ubuntu VPS, so I'm not sure if it will work with boot2docker. Presumably it will go pretty smoothly.

jez:
Also, you'll probably have to use localhost and do weird things with ports if your doing it on your laptop. The configuration really heavily favors production-readiness, so getting it set up on a laptop might not actually be that easy.

lib/tasks/admin.rake
icanberk: Wouldn't it be better to have `rake autolab:promote_user` since we already have that name space for other autolab related tasks?

icanberk: Wouldn't it be better to have `rake autolab:promote_user` since we already have that name space for other autolab related tasks?

dlbucci: I sort of agree, but admin:promote_user is more descriptive in my opinion, and I think populate and depopulate are the only autolab tasks, so for something like this, which could be used in production, maybe a different namespace makes sense.

jez: I actually tried to make it in the Autolab namespace at first and it didn't work. I think it was because there's a lot of hacky stuff going on in it to get auto populate to work. 

Now though, I agree with Dan that it sounds more descriptive.

jez:
@dlbucci Have you had a chance to look at those?

dlbucci:
No, I have a final at 830am tomorrow.  I'll check it out then.

jez:
Oh I almost forgot: we'll need to transfer the ownership of [jez/autolab-docker](https://github.com/jez/autolab-docker) eventually.

Dan, if you've tried to run through the instructions already, you'll have to use my repo instead (I think wrote the wiki documentation assuming it had already been transferred).

jez:
bump

jez:
Hmm, this PR has gone a whole without any feedback. @icanberk or @dlbucci, would one of you like to run through it? I wrote up some [pretty thorough instructions][1] for how to install Autolab this way. They require minimal docker knowledge. If this is still the part of the PR that you're the most uncomfortable with, I'd recommend using [this guide][2] as a reference in addition to the instructions in the PR. 

Try it out; you might have fun with it XD

[1]: https://github.com/autolab/Autolab/wiki/Deploying-Autolab-with-Docker
[2]: http://odino.org/spring-cleaning-of-your-docker-containers/

ymzong:
In
https://github.com/autolab/Autolab/wiki/Deploying-Autolab-with-Docker#4-get-the-autolab-source,
where is `autolab-docker` that you mentioned?

jez:
It's under my user because I don't have permissions to create repos:

https://github.com/jez/autolab-docker

We'll definitely want to move this over to the Autolab organization at some
point.

icanberk:
@jez Sorry I'm super super busy this week. I can definitely go through these instruction next week and we can merge this. I've also created `autolab-docker` repo just now, in case you want to transfer it.

jez:
Alright, I pushed everything to the new repo, so you shouldn't have to use mine.

Is there a reason why it's private?

icanberk:
Small not: 1.3.2 (default) version of docker-compose gives me this error: https://github.com/docker/compose/issues/1694

It works fine when I updated to 1.3.3

icanberk:
I got this when tried to migrate the db.

```
Ilters-MacBook-Pro:autolab-prod ican$ docker-compose run --rm -e RAILS_ENV=production web rake db:migrate
Starting autolabprod_db_1...
Cannot start container 2ef8d589bdcce38d8d961a762cf934e0c0326c94d6b59827c05cf15f3e6b852f: Cannot link to a non running container: /autolabprod_db_1 AS /autolabprod_web_run_1/autolabprod_db_1

```

Looks to me like it's trying to connect to `autolabprod_web_run_1` but the name of my docker is actually `autolabprod_web_1`

jez:
@icanberk that's interesting. Did you already run `docker compose up -d`? Docker Compose should normally take care of the container names.

And actually, those names make sense: you _should_ have two separate containers, `autolabprod_web_1` and `autolabprod_web_run_`, one from running the `docker-compose up -d` command and one from running the `docker-compose run` command.

It looks like the actual issue is not the name of the containers but the fact that the db container is not running?

Also, what do the logs say?

jez:
@icanberk I can't reproduce this error. I think that you may have forgotten to run `docker-compose up -d` before trying to migrate.

What does `docker ps -a` say when you get the failed migration? Does running `docker-compose logs` give any information about why the `db_1` container wouldn't be running?

GjjvdBurg:
Hey, I've come across the problem @icanberk describes too. The problem is that docker tries to link before the `db` container is started successfully. What worked for me was to start the db container first using `docker-compose up db`, wait until all initialization is done, then stop it and run `docker-compose up -d` again.

jez:
Oh, interesting. Thanks for investigating this. I'll look into a way to
force this synchronization, but in the mean time I'll update the
documentation in the wiki with your workaround.

GjjvdBurg:
Thanks for looking into it. It seems to be a common problem I think, [this](https://github.com/docker/docker/issues/7445) proposal seems related.

jez:
Looks like docker/docker closed it as `wontfix` under the pretense that it was an "orchestration" problem. I wonder if the discussion has moved to docker-compose?

GjjvdBurg:
The docker-compose discussion is over here I think: https://github.com/docker/compose/issues/374

icanberk:
That did the trick. Thanks @GjjvdBurg! @jez I'm merging this, but let's don't forget to put a warning on Step 11 on the wiki.

PR #493, opened 2015-04-29
Fixed runMoss

dlbucci:
Assuming the latest_submission issue isn't an issue.
Don't merge this unless Jimmy confirms that it was an issue on his end.


ymzong:
Imma test it by EOD.

PR #510, opened 2015-05-05
Exception Notifications for no-view routes

### LINKED ISSUE #491: We should be notified if no-view routes fail
Specifically, local_submit, log_submit, and autograde_done should be sending us emails if they fail, which doesn't happen because we (correctly) catch the errors before the global render_error function is called.

dlbucci:
Fixes #491 
This mucks with handin and autograding code, so test it out.  I checked and regular handins seemed to work fine with autograded assessments using polling for the autograding.  We should test the autograde_done callback on staging.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -20,11 +20,8 @@ class ApplicationController < ActionController::Base
   before_action :update_persistent_announcements
   before_action :set_breadcrumbs
 
-  # this is where Error Handling is configured. this routes exceptions to
-  # the error handler in the HomeController, unless we're in development mode
-  #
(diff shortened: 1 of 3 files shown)


jez:
@dlbucci What's the state of this PR, does it work?

How can I test it?

dlbucci:
It should work/shouldn't have a noticeable effect, except we'll get error emails and messages.  You can probably test that handin and autograding still works by pulling this to staging and testing there.

dlbucci:
Can we also get this into v1.7.0?

dlbucci:
PING

nayak16:
Deprecated

PR #401, opened 2015-03-17
Add datepickers to the Announcements forms

jez:
Remaking this after fixing conflicts.

See #385 for further discussion.

@dlbucci: I'm pretty sure I fixed the initialization issue discussed there.

PR #514, opened 2015-05-10
Puts attachments in a more to a more visible place

icanberk:
Puts attachments in a more to a more visible place and with better design. Removes the need for a "Manage Attachments" page -> Simplifies the left menu for the instructors.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
-/attachments/
+/attachments/*
+!/attachments/.gitkeep
 log/
 tmp/cache/
 config/autogradeConfig.rb
--- a/app/views/assessments/show.html.erb
+++ b/app/views/assessments/show.html.erb
@@ -157,8 +157,9 @@
       <% end %>
       <% if @cud.instructor? then %>
         <li class="attachment add">
-          <a href="<%= new_course_assessment_attachment_path(@course, @assessment) %>" class="glyphicon glyphicon-plus" aria-hidden="true"></a>
-          <%= link_to 'Add '+@assessment.display_name+' Attachment', new_course_assessment_attachment_path(@course, @assessment) %>
+          <%= link_to new_course_assessment_attachment_path(@course, @assessment) do %>
+            <span class="glyphicon glyphicon-plus"></span><%= "Add #{@assessment.display_name} Attachment" %>
+          <% end %>
         </li>
       <% end %>
     </ul>
--- a/app/views/assessments/show.html.erb
+++ b/app/views/assessments/show.html.erb
@@ -38,7 +38,6 @@
         <li> <%= link_to "Reload config file", {:action => "reload" }, {:title=> "Reload the assessment config file (provided for backward compatibility with legacy assessments)" } %></li>
         <li> <%= link_to "Manage extensions", [@course, @assessment, :extensions], { title: "Give extensions to students" } %></li>
         <li> <%= link_to "Manage submissions", [@course, @assessment, :submissions], { title: "Create, view, export, and re-autograde submissions" } %></li>
-        <li> <%= link_to "Manage assessment attachments", [@course, @assessment, :attachments], { title: "Distribute files to your students" } %></li>
         <li> <%= link_to "View statistics", {:action => "statistics" }, {:title=> "View detailed stats for this assessment" } %></li>
         <li> <%= link_to "Bulk import grades", {:action => "bulkGrade" }, {:title=> "Upload scores or feedback for multiple students from a CSV file" } %></li>
         <li> <%= link_to "Bulk export grades", {:action => "bulkExport" }, {:title=> "Export grades (with sub-scores) \nfor all students to a CSV file" } %></li>
@@ -138,8 +137,8 @@
 
   <%# Display any attachments %>
   <% if @assessment.attachments.length > 0 %>
-    <h2><%= @assessment.display_name %> Files</h2>
-    <ul>
+    <h2>Handouts</h2>
+    <ul class="attachments">
       <%= render @assessment.attachments %>
 
       <% # this code is temporarily "commented out" because Dan wants to preserve the logic for future changes to _attachment.html.erb %>
@@ -157,6 +156,10 @@
         <% end %>
       <% end %>
     </ul>
+    <% if @cud.instructor? then %>
+      <%= link_to 'Add '+@assessment.display_name+' Attachment', new_course_assessment_attachment_path(@course, @assessment) %>
+    <% end %>
+
   <% end %>
 
   <% if @assessment.exam? %>
--- a/app/views/assessments/show.html.erb
+++ b/app/views/assessments/show.html.erb
@@ -136,7 +136,7 @@
   <% end %>
 
   <%# Display any attachments %>
-  <% if @assessment.attachments.length > 0 %>
+  <% if @assessment.attachments.length > 0 or @cud.instructor? %>
     <h2>Handouts</h2>
     <ul class="attachments">
       <%= render @assessment.attachments %>
--- a/app/views/assessments/show.html.erb
+++ b/app/views/assessments/show.html.erb
@@ -155,10 +155,15 @@
           <% end %>
         <% end %>
       <% end %>
+      <% if @cud.instructor? then %>
+        <li class="attachment add">
+          <a href="<%= new_course_assessment_attachment_path(@course, @assessment) %>" class="glyphicon glyphicon-plus" aria-hidden="true"></a>
+          <%= link_to 'Add '+@assessment.display_name+' Attachment', new_course_assessment_attachment_path(@course, @assessment) %>
+        </li>
+      <% end %>
     </ul>
-    <% if @cud.instructor? then %>
-      <%= link_to 'Add '+@assessment.display_name+' Attachment', new_course_assessment_attachment_path(@course, @assessment) %>
-    <% end %>
+
+
 
   <% end %>
 
--- a/app/views/attachments/_attachment.html.erb
+++ b/app/views/attachments/_attachment.html.erb
@@ -1,19 +1,29 @@
-<li>
+<li class="attachment">
+  <% if attachment.assessment.nil? then %>
+    <a href="<%= url_for [@course, attachment] %>" class="glyphicon glyphicon-folder-open" aria-hidden="true"></a>
+  <% else %>
+    <a href="<%= url_for [@course, attachment.assessment, attachment] %>" class="glyphicon glyphicon-folder-open" aria-hidden="true"></a>
+  <% end %>
+  
   <% if attachment.assessment.nil? then %>
     <%= link_to attachment.name, [@course, attachment] %>
   <% else %>
     <%= link_to attachment.name, [@course, attachment.assessment, attachment] %>
   <% end %>
+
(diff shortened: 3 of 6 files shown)


app/views/attachments/_attachment.html.erb
```diff
@@ -1,19 +1,29 @@
...
+    <a href="<%= url_for [@course, attachment.assessment, attachment] %>" class="glyphicon glyphicon-folder-open" aria-hidden="true"></a>
+  <% end %>
+  
   <% if attachment.assessment.nil? then %>
     <%= link_to attachment.name, [@course, attachment] %>
   <% else %>
     <%= link_to attachment.name, [@course, attachment.assessment, attachment] %>
   <% end %>
+
+  <% if !attachment.assessment.nil? and @assessment.nil? then %>
```
dlbucci: Use && instead of and.

app/views/attachments/_attachment.html.erb
```diff
@@ -1,19 +1,29 @@
...
+    (<%= attachment.assessment.display_name %>)
+  <% end %>
+
+  <span class="pull-right">
   <% if @cud.instructor? then %>
     <% if attachment.assessment.nil? then %>
-      ---
       <%= link_to 'Edit', [:edit, @course, attachment] %>
```
dlbucci: Why aren't there bootstrap classes for this links as well?

dlbucci:
If there are routes in the AttachmentsController that this removes the need for, we should remove that code.

icanberk:
I think attachments index page is still linked from some other places, so we still need the routes.

dlbucci:
It would be better to fix those links, as opposed to duplicating code/functionality.

jez:
@icanberk I can't seem to get these changes to show up. Am I looking on the right page? I'm on `http://localhost:3000/courses/AutoPopulated/assessments/homework0`, but I don't see anything new related to attachments on the page.

icanberk:
@jez Are you sure you have attachments on the page?

https://www.dropbox.com/s/vc1bfbmktmbo7qw/Screenshot%202015-06-13%2016.23.04.png?dl=0

dlbucci:
Why is there a folder icon next to each attachment?  Wouldn't a file icon be more accurate?

app/views/assessments/show.html.erb
```diff
@@ -157,6 +156,10 @@
         <% end %>
       <% end %>
     </ul>
+    <% if @cud.instructor? then %>
+      <%= link_to 'Add '+@assessment.display_name+' Attachment', new_course_assessment_attachment_path(@course, @assessment) %>
+    <% end %>
+
```
jez: I think that you want to display this regardless of whether `@assesment.attachments.length` is greater than 0.

jez:
That's a good point Dan. Even though people might be uploading tarballs as attachments, the folder icon makes me think I should be expanding the folder to see more files.

jez:
When I try to upload a file, I get this error:

```
No such file or directory @ rb_sysopen - /Users/jake/Dropbox/Documents/Programming/autolab/Autolab/attachments/resume.pdf
```

For context:
- `/Users/jake/Dropbox/Documents/Programming/autolab/Autolab` is where my Autolab repo is
- I tried uploading a file called `resume.pdf` from `/Users/jake/Dropbox/resume.pdf`

Maybe there's an issue with where the controller is looking for the file?

Here's the stack trace (Application Frames):

```
Attachment#file=
app/models/attachment.rb, line 24
AttachmentsController#update
app/controllers/attachments_controller.rb, line 53
AttachmentsController#create
app/controllers/attachments_controller.rb, line 29
```

jez:
While we're touching attachments, I think we should take the time to update `app/views/attachments/new.html.erb` to use the new FormBuilder that we've been cleaning up. It should be a pretty easy change, and you can see an example of how it works on one of my previous PRs:

- #401 Add datepickers to the Announcements forms, specifically
  - This line <https://github.com/autolab/Autolab/pull/401/files#diff-b76ac1fddc1583275323589933fc65d9R4> and
  - This file <https://github.com/autolab/Autolab/pull/401/files#diff-7b8be66719056f3df229b112add26a48R1>

app/views/assessments/show.html.erb
dlbucci: Agreed.

dlbucci:
@jez make an 'assessments' folder at the top level of your local repo.

jez:
Alright that worked (assuming you made a typo and meant 'attachments' folder).

- Is there a way to automate the process of creating that folder?
- How does that work if multiple classes (or even one class with multiple assessments) have conflicting filenames?

jez:
I think the "No Submissions yet!" should be on a separate line.

screenshot 2015-06-13 18 58 05

It might also make sense to have the "Add ... Attachment" link be a button.

dlbucci:
We could add a check for the folder before an attachment is created, but adding a .gitkeep or .gitignore file might work just as well for making sure the folder is there. That would be less clutter in the code.

Conflicting filenames just get numbered, I think. There's definitely some kind of conflict resolution.

icanberk:
You right about the icon and the add button thing. Will fix that today.

icanberk:
https://www.dropbox.com/s/26wtblqk7whsp32/Screenshot%202015-06-14%2012.45.29.png?dl=0

jez:
Oh, there's also this one: 
screenshot 2015-06-14 12 31 14.

Either one looks fine to me.

icanberk:
I still like the other one better but up to you. Feel free to change it.

jez:
Ok sure.

What do you think about this patch that cleans up the code a bit?

```diff
diff --git a/app/views/assessments/show.html.erb b/app/views/assessments/show.html.erb
index 8f14033..94e6292 100755
--- a/app/views/assessments/show.html.erb
+++ b/app/views/assessments/show.html.erb
@@ -157,8 +157,9 @@
       <% end %>
       <% if @cud.instructor? then %>
         <li class="attachment add">
-          <a href="<%= new_course_assessment_attachment_path(@course, @assessment) %>" class="glyphicon glyphicon-plus" aria-hidden="true"></a>
-          <%= link_to 'Add '+@assessment.display_name+' Attachment', new_course_assessment_attachment_path(@course, @assessment) %>
+          <%= link_to new_course_assessment_attachment_path(@course, @assessment) do %>
+            <span class="glyphicon glyphicon-plus"></span><%= "Add #{@assessment.display_name} Attachment" %>
+          <% end %>
         </li>
       <% end %>
     </ul>
diff --git a/app/views/attachments/_attachment.html.erb b/app/views/attachments/_attachment.html.erb
index f0e6e31..ddd22ed 100755
--- a/app/views/attachments/_attachment.html.erb
+++ b/app/views/attachments/_attachment.html.erb
@@ -1,14 +1,12 @@
 <li class="attachment">
   <% if attachment.assessment.nil? then %>
-    <a href="<%= url_for [@course, attachment] %>" class="glyphicon glyphicon-list-alt" aria-hidden="true"></a>
-  <% else %>
-    <a href="<%= url_for [@course, attachment.assessment, attachment] %>" class="glyphicon glyphicon-list-alt" aria-hidden="true"></a>
-  <% end %>
-  
-  <% if attachment.assessment.nil? then %>
-    <%= link_to attachment.name, [@course, attachment] %>
+    <%= link_to [@course, attachment] do %>
+      <span class="glyphicon glyphicon-list-alt"></span><%= attachment.name %>
+    <% end %>
   <% else %>
-    <%= link_to attachment.name, [@course, attachment.assessment, attachment] %>
+    <%= link_to [@course, attachment.assessment, attachment] do %>
+      <span class="glyphicon glyphicon-list-alt"></span><%= attachment.name %>
+    <% end %>
   <% end %>
 
   <% if !attachment.assessment.nil? && @assessment.nil? then %>
```

The biggest thing that It changes is that it gets rid of redundant `link_to`s so that when you hover, both the icon and text change.

icanberk:
Looks good to me :+1:

app/views/assessments/show.html.erb
```diff
@@ -38,7 +38,6 @@
         <li> <%= link_to "Reload config file", {:action => "reload" }, {:title=> "Reload the assessment config file (provided for backward compatibility with legacy assessments)" } %></li>
         <li> <%= link_to "Manage extensions", [@course, @assessment, :extensions], { title: "Give extensions to students" } %></li>
         <li> <%= link_to "Manage submissions", [@course, @assessment, :submissions], { title: "Create, view, export, and re-autograde submissions" } %></li>
-        <li> <%= link_to "Manage assessment attachments", [@course, @assessment, :attachments], { title: "Distribute files to your students" } %></li>
```
jez: Since we've removed this link, there's no way to add an attachment if there isn't one there yet.

app/views/assessments/show.html.erb
```diff
@@ -156,7 +155,17 @@
           <% end %>
         <% end %>
       <% end %>
+      <% if @cud.instructor? then %>
+        <li class="attachment add">
+          <%= link_to new_course_assessment_attachment_path(@course, @assessment) do %>
+            <span class="glyphicon glyphicon-plus"></span><%= "Add #{@assessment.display_name} Attachment" %>
+          <% end %>
+        </li>
+      <% end %>
```
jez: We need to move this logic outside of the outer `if` block to expose the link even when there aren't attachments.

jez:
@dlbucci Is this what you'd imagine the gitkeep thing looks like? I've never done this before.

jez:
@icanberk If you address those last two inline comments I made, this will be ready to :ship:

app/views/assessments/show.html.erb
```diff
@@ -138,8 +137,8 @@
 
   <%# Display any attachments %>
   <% if @assessment.attachments.length > 0 %>
```
icanberk: How about adding `or @cud.instructor?` here?

jez: Then students would see the 'Add ... Attachment' link if there were more than 0 attachments.

icanberk: We still have the other `<% if @cud.instructor? then %>` below.

jez: Ok, sure. I think that works.

.gitignore
```diff
@@ -1,4 +1,5 @@
-/attachments/
+/attachments/*
+!/attachments/.gitkeep
```
dlbucci: Actually, would placing a .gitignore file that only contained '*' achieve the same effect here?

jez: I was using [this StackOverflow](http://stackoverflow.com/a/19005115) article as an example, and it discouraged that.

jez:
Merged from the command line.

icanberk:
:pray:

.gitignore
dlbucci: I disagree with that answer, based on this answer: http://stackoverflow.com/questions/7229885/what-are-the-differences-between-gitignore-and-gitkeep
Basically, .gitkeep is a convention thing, but doesn't really mean anything, whereas .gitignore has significance.  I think the .gitignore might be cleaner, but it's not too important.

jez: Oh, interesting. I just merged it into develop. Do you want me to touch it up, or should we just keep it as is?

Issue #180, opened 2015-01-24
Condense Edit Assessment Tabs into Single Page


jez:
Rather than having the various parts of an assessment spread across four distinct pages (linked to with active/inactive tabs), place all the constituent parts required to edit an assessment on the same page.

Navigation between the logical units can be accomplished using something similar to a scroll spy with in-page anchors. For inspiration, see the Piazza "Manage Class" interface.

Z1MM32M4N:
Originally #157, point (3).

icanberk:
Hi @droh , What do you think about this?

I feel like the page might become too intimidating to the instructors if we merge into one. They are pretty used to the tabs right now.

jez:
I think the tabs work pretty well, and the interface got a nice redesign with the new FormBuilder, so I'm closing this for now.

icanberk:
I'm contemplating on putting those tabs to left side like menu items, but not an urgency.

dlbucci:
It's probably more important to have the tabs function on the same page, more than anything (i.e. clicking a tab shouldn't take you to a new form/make you lose your data).

jez:
We actually have an issue for that specifically that I've been meaning to
get around to.

jez:
After discussing this in the team meeting today, we agreed to reopen this today to
- reduce complexity in the codebase
- simplify the process for ensuring that all the properties get saved to the correct places

/cc @dlbucci @droh

remzr7:
Closed as irrelevant

PR #515, opened 2015-05-10
Move autograder + scoreboard settings into Assessment Settings

icanberk:
My goal here is to simplify the instructor menu for assessments.

1) It renames the menu options to distinguish between "Settings", properties that needs to be set once, and "Admin", managing ongoing processes like groups and SVN. 
2) Moves the Actions to the top of the menu. "Grade submissions" is by far the most clicked option in the menu, so it's going to be on the top.
3) Moves some of the settings into Assessment settings page. This is also a step towards Issue #180. I'm going to handle that next.


app/views/assessments/_edit_basic.html.erb
```diff
@@ -20,8 +20,7 @@
 <h4>Modules Used</h4>
 <div class="form-group">
   <% if @assessment.has_autograder? %>
-    <%= link_to "Edit Autograder Settings", [:edit, @course, @assessment, :autograder], class: "btn primary" %>
-    <%= link_to "Delete Autograder", [@course, @assessment, :autograder], method: :delete, class: "btn danger",
+    <%= link_to "Remove Autograder", [@course, @assessment, :autograder], method: :delete, class: "btn danger",
```
dlbucci: If you're gonna change this from "Delete" to "Remove", you should do the same on the actual autograder form as well. Same for the Scoreboard.

app/views/assessments/show.html.erb
```diff
@@ -28,37 +28,25 @@
...
         <% if @assessment.has_groups? %>
-          <li><%= link_to "Group Settings", [@course, @assessment, :groups], title: "Manage Group Settings" %></li>
+        <li> <%= link_to "Admin Groups", [@course, @assessment, :groups], title: "Manage Group Settings" %></li>
         <% end %>
-
-        <% if @assessment.has_scoreboard? %>
-          <li><%= link_to "Scoreboard Settings", [:edit, @course, @assessment, :scoreboard], title: "Configure the appearance of the scoreboard" %></li>
-        <% end %>
-
+        <li> <%= link_to "Assessment Settings", {:action => "edit" }, {:title=> "View and modify the properties for this assessment, including its problems" } %></li>
```
dlbucci: Why do you insist on using the ugly old object syntax?

app/views/assessments/show.html.erb
```diff
@@ -28,37 +28,25 @@
...
         <li> <%= link_to "Manage extensions", [@course, @assessment, :extensions], { title: "Give extensions to students" } %></li>
         <li> <%= link_to "Manage submissions", [@course, @assessment, :submissions], { title: "Create, view, export, and re-autograde submissions" } %></li>
         <li> <%= link_to "Manage assessment attachments", [@course, @assessment, :attachments], { title: "Distribute files to your students" } %></li>
         <li> <%= link_to "View statistics", {:action => "statistics" }, {:title=> "View detailed stats for this assessment" } %></li>
         <li> <%= link_to "Bulk import grades", {:action => "bulkGrade" }, {:title=> "Upload scores or feedback for multiple students from a CSV file" } %></li>
         <li> <%= link_to "Bulk export grades", {:action => "bulkExport" }, {:title=> "Export grades (with sub-scores) \nfor all students to a CSV file" } %></li>
-
         <% if @assessment.has_svn then %>
-          <li><%= link_to "SVN Settings", [:admin_svn, @course, @assessment], title: "Manage the SVN repositories" %></li>
+        <li> <%= link_to "Admin SVN", [:admin_svn, @course, @assessment], title: "Manage the SVN repositories" %></li>
```
dlbucci: I guess I see why this isn't quite a "Settings" page, but I don't really like the use of "Admin" as a verb either.  Maybe "Manage SVN" would be better?  Same for groups.

dlbucci:
At the last meeting, we discussed the tabs on this page and decided that we didn't really like them and wanted to get rid of them/make them a single page.  This sort of goes against that, and I'm not sure I like it (it's actually a step away from solving #180, not a step towards fixing it).

Once you have multiple "modules" enabled, you can have up to 8 tabs on one page.  Also, the SVN and Groups tabs don't stay on the assessments#edit page, and take you to the current pages for SVN and Group management (not sure if that's a work in progress or not).
From a RESTful stand point, it makes more sense for each of these pages to be in a different controller with a different route, and the way we handle the routing is already sort of a hack.

icanberk:
This is what it looks like:

https://www.dropbox.com/s/a4bqbg44kj06mt6/Screenshot%202015-06-14%2013.34.41.png?dl=0
https://www.dropbox.com/s/ykl4v3xnqfrxosr/Screenshot%202015-06-14%2013.35.20.png?dl=0

Closes #180

dlbucci:
So the menu is always displaying on the left and the links just direct you to anchors, correct?

icanberk:
yessir

jez:
@icanberk I'm going to take a look at this after I get back from lunch. Just a heads up, though: it looks like #514 Puts attachments in a more visible place introduced merge conflicts on this branch.

icanberk:
Should be good now

jez:
Instructors will be confused that we renamed "Edit Assessment" to "Assessment Settings".

We should make a blog post of things we've changed (attachments, assessment settings, etc.) and send it in an email to instructors. That can be done independently of this PR.

icanberk:
I agree. We can also include PDF annotations in that email. We won't be merging this to master for a while.

app/views/assessments/edit.html.erb
```diff
@@ -36,7 +53,164 @@
...
+
+<%= f.text_field :display_name,
+  help_text: "Name that will be displayed on the course home page. E.g.,
+  'Malloc Lab'" %>
+
+<div class="form-group">
+  <%= f.label :category_name, { :class=>"control-label" } %>
+  <%= f.collection_select :category_name, @course.assessment_categories, :to_s, :to_s, {selected: @assessment.category_name}, {class: "form-control"} %>
+  <p>or</p>
+  <%= text_field_tag :new_category, nil, placeholder: "New Category" %>
```
jez: `text_field_tag` should use `f.text_field` instead.

icanberk: `new_category` is not a field, so I can't seem to use f.text_field. Do you know how to get around it?

app/controllers/scoreboards_controller.rb
```diff
@@ -123,8 +123,10 @@ def show
 
   action_auth_level :edit, :instructor
   def edit
+    params[:active_tab] ||= "scoreboard"
     # Set the @column_summary instance variable for the view
     @column_summary = emitColSpec(@scoreboard.colspec)
+    redirect_to(edit_course_assessment_path(@course, @assessment)+"/#{params[:active_tab]}") && return
```
jez: Do we have do to something similar on the `update` routes as well?

app/views/assessments/_edit_basic.html.erb
jez: :+1:

jez:
Didn't our flash messages used to be more substantial?

screenshot 2015-06-14 15 55 08

jez:
This actually looks pretty good. @icanberk I think you still have some of @dlbucci's comments, and I just left some inline comments. I'll take another look when you're ready.

app/views/assessments/edit.html.erb
jez: Oh, dang, I don't. If you can't figure out how to get it working, we can just push it off to some other time, or you can hard code the classes that the FormBuilder would generate.

icanberk:
@jez All addressed.

PR #516, opened 2015-05-10
SQLite Compatibility

dlbucci:
This shouldn't introduce any functionality differences to anyone using MySQL, instead the site should be able to use SQLite, which should make setup easier.


jez:
What's the best way to test this? I see you've added some sample code in `database.yml.template`. I guess since we'd only be using SQLite in development that it's not entirely crucial we catch all the bugs now?

dlbucci:
Yeah, the changes to the queries are more important than the SQLite stuff, but as long as the pages that have had queries get changed still work, this should be good.  I'd test that using SQLite on your machine works as well.

config/database.yml.template
```diff
@@ -7,6 +8,13 @@ development:
   socket: /var/run/mysqld/mysqld.sock
   host: localhost
 
+# SQLite Configuration
+#development:
+#  adapter: sqlite3
+#  database: db/butts.sqlite3
```
jez: That's maybe not the best name for a sqlite database file.

jez:
After

- changing my `config/database.yml` file
- running `rake db:create`
- running `rake db:migrate`, I get this error:

`SQLite3::SQLException: Cannot add a NOT NULL column with default value NULL: ALTER TABLE "assessments" ADD "grading_deadline" datetime NOT NULL`

It looks like it fails on this migration:

````
= 20130330220843 AddGradingDeadlineToAssessment: migrating ===================
-- add_column(:assessments, :grading_deadline, :datetime, {:null=>false})
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:
```

jez:
There are five years worth of migrations in that folder ಠ_ಠ

jez:
@dlbucci on your machine, does it work for this migration?

dlbucci:
That won't/doesn't need to work anywere. Use db:reset instead.

jez:
Ok. I'll make a note about that in the README.

Out of curiosity, what was your process for identifying places that would need to be changed to provide SQLite compatibility?

jez:
Merged from the command line.

I tested the routes that looked like they had changed, and they seem to be correct. Since this is just for development purposes anyways, we can test it better by incorporating in the changes now and testing more down the road.

dlbucci:
Is there a reason you have been merging from the command line?
Also, I basically checked for places where we use SQL, and made sure it worked there.  The adapters should handle the rest.

jez:
I've been merging from the command line so that we can squash and rebase our PRs.

It leads to a much nicer commit log history, which makes it easier to revert commits if necessary as well as discern what changes have been made since you last pulled.

Dropping this link for reference: [How to Rebase a Pull Request][1].

[1]: https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request

PR #528, opened 2015-06-16
Fix flash message for confirmation email sent

### LINKED ISSUE #478: "Need to sign in or sign up before continuing"
Steps to reproduce:

- Sign up for account
- Press "Sign up"

Observed behavior:

- Redirected to sign up page with error message "Need to sign in or sign up before continuing"

Expected behavior:

- Redirected to sign up page with error message "Check your email to confirm your account"

jez:
When devise sends an email with a validation/confirmation link after a
new user registers using the email/password "Sign up" option, it used to
say "You must be sign in or sign up to continue"

It showed this message because we were redirecting to "/", which
requires that the user is authorized. Now, it redirects to the sign in
page so that the devise flash message discussing what to do next doesn't
get clobbered by a chain of redirects.

Most code taken from the documentation at

<https://github.com/plataformatec/devise/wiki/How-To:-Redirect-to-a-specific-page-on-successful-sign-up-%28registration%29>

Fixes #478

__Please don't use the GitHub merge button to merge this commit. Instead,
follow the instructions here:__

<https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request>

to ensure that the commit is rebased against `develop` first, and then `git merge --no-ff` into develop.

### DIFF
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -1,0 +1,10 @@
+class RegistrationsController < Devise::RegistrationsController
+  protected
+
+  # After hitting the 'Sign up' page, redirect users here so that they see the
+  # confirmation flash message ("check your email...")
+  def after_inactive_sign_up_path_for(resource)
+    # /auth/users/sign_in
+    new_user_session_path
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,7 +1,8 @@
 Autolab3::Application.routes.draw do
   root "courses#index"
 
-  devise_for :users, controllers: { omniauth_callbacks: "users/omniauth_callbacks" },
+  devise_for :users, controllers: { omniauth_callbacks: "users/omniauth_callbacks",
+                                    registrations:      "registrations" },
                      path_prefix: "auth"
 
   get "contact", to: "home#contact"


app/controllers/registrations_controller.rb
```diff
@@ -0,0 +1,10 @@
+class RegistrationsController < Devise::RegistrationsController
+  protected
+
+  # After hitting the 'Sign up' page, redirect users here so that they see the
+  # confirmation flash message ("check your email...")
+  def after_inactive_sign_up_path_for(resource)
+    # /auth/users/sign_in
+    new_user_session_path
+  end
+end
```
icanberk: (@dlbucci parody comment :D )

I don't know what these 10 lines do. Our codebase is already bad enough. I'm not sure if adding this meaningless stuff just for a flash message is a good ida.

icanberk:
That was pretty painful. Is it really worth rebasing for every merge?

jez:
I think it is. Our commit history is a nightmare. It makes it hard for
people to keep track of what we've been doing lately if they want to see,
and also it makes things easier when we need to quickly revert something
that breaks Autolab.

It is pretty tricky but I think it'll be pretty painless in time.

jez:
And actually there's a much easier way of getting the same effect:

```shell
$ git checkout develop
$ git merge --squash feature-branch
$ git commit
```

It "just works" and you don't have to do any `merge-base`, `rebase -i`, etc.

icanberk:
That looks easy enough. You should write it on the white board in the room
or something.

PR #530, opened 2015-07-05
Annotations for PDF submission

icanberk:
This is the initial implementation for a system that will be used by theory classes.

1) Let's us display PDFs on a page using PDF.js by mozilla
2) Let's the instructors add annotations on a page, similar to the code annotations
3) Let's the students view this feedback.
4) Let's both students and instructors download an annotated version of the submission as PDF (using Prawn library)

Javascript implementation is a little messy since I built it upon the annotation system that we use for code annotations. Planning to switch to a good and simple JS library like Backbone in the future.


db/schema.rb
```diff
@@ -150,6 +152,9 @@
     t.integer  "version_penalty_id",            limit: 4
     t.datetime "cgdub_dependencies_updated_at"
     t.text     "gb_message",                    limit: 65535
+    t.boolean  "public",                        limit: 1,     default: false
```
icanberk: going to fix these... left overs from my other branch.

icanberk:
Going to pull this into staging soon, so you guys can play around.

icanberk:
Demo is here: https://sawshark.ics.cs.cmu.edu/courses/15122-s15/assessments/induction/submissions/68741/view

ymzong:
I absolutely love this!!! :smiley:

Gemfile
```diff
@@ -98,5 +98,11 @@ group :development do
   gem 'rubocop', require: false
 end
 
+# for PDF annotations
+# This is an outdate version however support for
+# templating has been dropped in the future versions
+# and it is crucial for us
+gem 'prawn', '0.13.0'
```
jez: What does "templating" give us?

Did you consider any alternative, maintained libraries? (i.e., why was this one the best?)

app/assets/javascripts/annotations.js
```diff
@@ -236,13 +248,106 @@ $(function() {
...
+      value: "Save",
+      class: "btn primary small"
+    });
+    var cancelButton = elt("input", {
+      style: "margin-left: 4px;",
+      type: "button",
+      value: "Cancel",
+      class: "btn small"
+    });
+    var hr = elt("hr");
```
jez: I would really love to switch over all or a significant portion of the rendering to React and use the Rails app really as a RESTish server. We'd be able to avoid doing weird stuff like this as well as be able to treat Autolab more as a snappy, single-page app.

app/views/submissions/viewPDF.html.erb
```diff
@@ -81,5 +113,30 @@
...
 <div class="row">
+  <div class="col-md-2 annotationPane">
+    <% if not (@submission.assessment.exam or @submission.assessment.course.exam_in_progress) %>
+      <% if @preview_mode then %>
+      <a style="width:100%;" href="?" title="Download Submission" class="btn"> 
+        <span class="glyphicon glyphicon-pencil pull-left" aria-hidden="true"></span>
+        Edit
+      </a>
+      <% else %>
+      <a style="width:100%;" href="?preview=true" title="Download Submission" class="btn">
```
jez: Should `title` be "Download Submission" on all of these things?

jez:
Some things from an interface perspective that are not clear:

- clicking on the PDF allows you to start composing a comment
- what part of the document the comment refers to
  - ideally, we'd be able to specify a range (select some text, or select a rectangle) that the comment applies to, as well as comment on the document as a whole.

jez:
I didn't realize that the "Preview" button became an "Edit" button after I'd clicked on it. Perhaps we should make the button change color, and maybe call it something like "Toggle comments"

There is no way to view the document without any comments on it (only the Preview or the Edit views are what I can find).

jez:
I'm also not sure how I feel about the comments covering up as much of the PDF as they do.

Some ways I've seen other people get around this:

- Color My Graph used a library that only showed the content of the comment when it was selected. At other times, it just drew a box to refer to where the comment was placed. It [looked like this][1] and worked rather well.
- Dropbox is working on a feature that shows a small circle where you initially clicked on the document that represents where the comment was left. It then
  - Shows a list of all comments in a side bar
  - Lets you use this circle to drag the comment around
  - Lets you click this circle to hide the comment (it's still visible in the sidebar)

I think that being able to comment on a range, move comments around, toggle comments on/off, and see all comments in a sidebar would all be great features to implement next.

[1]: http://documents.nytimes.com/charles-darwin-on-the-origin-of-species#document/p185/a8

jez:
All in all, nice work! I think this could easily be the most impactful feature we release for Autolab next semester.

icanberk:
We went over this with @droh too. There needs to be a lot of UI changes for this to be usable by the instructors. Hopefully another PR coming this weekend.

As for the framework thing, it's way too much work to switch to whole thing to React, but I definitely think we can use something on this page and maybe slowly switch parts of the app over time.

PR #531, opened 2015-07-05
Spruced up landing page styles

### LINKED ISSUE #453: Editing course settings appends an extra "(semester)" field to the course name
Each time an instructor updates the Course Settings page, Autolab appends an extra "(semester)" field to the course display name. For example, after updating the 213 course settings 4 times, the Display Name for 213 is now:  

15213: Introduction to Computer Systems (f15)(f15)(f15)(f15)(f15)

Can someone make a hot fix for this? It looks really weird.

jez:
When a new user creates an account, now they see some helpful "Getting
Started" information, instead of the "You aren't in any classes"
nonsense.

Fixes #453.

### DIFF
--- a/app/views/home/_signUp.html.erb
+++ b/app/views/home/_signUp.html.erb
@@ -1,6 +1,0 @@
-<h2>Interested in Autolab?</h2>
-
-<p>Are you an instructor who is interested in autograding, or a TA who wants a
-better way to handle submissions? Autolab might be the tool you've been looking
-for. You can read more about Autolab's purpose on our
-<a href="http://greatwhite.ics.cs.cmu.edu/">public page</a>.</p>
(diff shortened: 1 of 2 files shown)


icanberk:
I love this.

Do you think it'd make sense if we put a `if administrator` clause around the For Admins part?

jez:
So actually the first reason why I thought we could use this page was because I was setting up Autolab for Docker ( #489 ) and I didn't know how to make myself an admin.

Also, I think it could also pique people's curiosity in Autolab regardless; maybe after seeing the admin stuff they'll take a peek at the docs or the source.

icanberk:
I see. Cool.

icanberk:
Closed by 57c8acd4c646b9d52c75e1864844e002e0122cbe

Issue #524, opened 2015-06-14
Refactor the attachments#new view to use our FormBuilder


jez:
The form in `app/views/attachments/new.html.erb` should use the new FormBuilder (`FormBuilderWithDateTimePicker`) that we've been cleaning up. It should be a pretty straightforward change, and you can see an example of how it works on one of my previous PRs:

- #401 Add datepickers to the Announcements forms, specifically
  - This line <https://github.com/autolab/Autolab/pull/401/files#diff-b76ac1fddc1583275323589933fc65d9R4> and
  - This file <https://github.com/autolab/Autolab/pull/401/files#diff-7b8be66719056f3df229b112add26a48R1>

Right now, it looks pretty ugly:

![screenshot 2015-06-14 13 40 13](https://cloud.githubusercontent.com/assets/5544532/8150353/f9bad028-129a-11e5-96fe-a861ba1c195c.png)

in comparison with our other forms:

![screenshot 2015-06-14 13 41 15](https://cloud.githubusercontent.com/assets/5544532/8150355/10cdc888-129b-11e5-88c3-9fc3344c3d79.png)

remzr7:
@nayak16 shouldn't this be closed? The form is using FormBuilderWithDateTimePicker

ymzong:
@remzr7 Yup that's right -- we forgot to close the issue after the PR was merged. Thanks for the heads up! :smiley_cat:

PR #536, opened 2015-07-19
Use form builder in attachment views

cjford:
Updated create/edit attachments views to use FormBuilderWithDateTimeInput 

See issue #524

### DIFF
--- a/app/views/attachments/new.html.erb
+++ b/app/views/attachments/new.html.erb
@@ -1,23 +1,8 @@
-<% path = (@is_assessment)? course_assessment_attachments_path(@course, @assessment) : course_attachments_path(@course) %>
-<%= form_for :attachment, url: path, method: :create, html: {multipart: true} do |f| %>
-  <%= f.error_messages %>
-  <table class="verticalTable">
-    <tr>
-      <th>Name</th>
-      <td><%= f.text_field :name %></td>
-    </tr>
-    <tr>
-      <th>Released</th>
-      <td><%= f.check_box :released %></td>
-    </tr>
-    <tr>
-      <th>File</th>
-      <td><%= f.file_field :file %></td>
-    </tr>
-    <tr>
-      <td colspan=2>
-        <%= f.submit("Upload") %>
-      </td>
-    </tr>
-  </table>
-<% end %>
+<% path =
+  if @is_assessment
+    course_assessment_attachments_path(@course, @assessment)
+  else
+     course_attachments_path(@course)
+  end %>
+
+<%= render partial: "form", locals: {path: path} %>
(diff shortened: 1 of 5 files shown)


dlbucci:
This looks good, but the one thing I can recommend is to put the form fields into a partial template, as shown here: http://guides.rubyonrails.org/getting_started.html#rendering-a-partial-form

I think if you use form_for with a new instance that hasn't been saved yet (`@attachment = Attachment.new`, or what have you), rails will be smart enough to use the Create route instead of the Update route.  There are some other places in the code that do this as well.

jez:
+1 to @dlbucci's comments. Before merging this in Dan, make sure that it gets rebased appropriately on top of @cjford's last commit to develop:

screen shot 2015-07-19 at 5 10 27 pm

cjford:
@dlbucci Do you mean to use the same form partial for both the new and edit views, or pull them out into two separate partials? It would be nice to use one, but they are slightly different right now - edit has a field to manually set the mime type (which I assume is set automatically in new based on the uploaded file) Any way to unify those?

Also, would you prefer sticking the whole form (including form_for tag) into a partial like they do in that guide, or just the fields themselves like I'm seeing some places in the existing code? (e.g. _announcementFields.html.erb)

dlbucci:
I was thinking of moving them into a single partial view to reduce repetition, but I didn't notice that the fields were slightly different.  You could still do that though by checking if the model your rendering is new or not (`@attachment.new_record?`).

cjford:
Updated commit so new/edit use a single partial

app/views/attachments/new.html.erb
```diff
@@ -1,23 +1,7 @@
...
-        <%= f.submit("Upload") %>
-      </td>
-    </tr>
-  </table>
-<% end %>
+<% path = if @is_assessment
+            course_assessment_attachments_path(@course, @assessment)
+          else
+            course_attachments_path(@course)
+          end %>
```
jez: Style nit: nest the `if` statement under the assignment to avoid large rectangles of whitespace.

```
val =
  if true
    "foo"
  else
    "bar"
  end
```

jez:
I added a single style comment.

@dlbucci: unless you have any outstanding comments, I can merge this.

icanberk:
Looks good to me

jez:
@cjford Have you had a chance to look touch up the style?

cjford:
@jez Apologies for the delay - updated with your feedback.

jez:
Thanks, I'll test this out locally and merge soon enough.

jez:
Merged via 35eac1f

Issue #537, opened 2015-07-20
Refactor the courses#new view to use our FormBuilder


jez:
See #524 for more details.

This view is particularly important because creating a course is one of the first ways users will interact with our product, so it's important to be polished.

It look like this right now:

![screenshot 2015-07-20 14 03 59](https://cloud.githubusercontent.com/assets/5544532/8787355/826463c6-2ee8-11e5-9c13-3ebc4f532381.png)

jez:
We should also make sure that we copy the help text from the "Course Settings" page.

jez:
Closed by #538

PR #538, opened 2015-07-22
Update course#new form to use form builder

cjford:
Fix for #537 

Since instructor_email isn't an attribute of the course, the field can't go through the form builder. So I just wrote out the Bootstrap classes, not sure if there's any better way of doing that...

Also it didn't seem worth putting the form in a partial since it's not used anywhere else, correct me if I'm wrong though.

### DIFF
--- a/app/views/courses/new.html.erb
+++ b/app/views/courses/new.html.erb
@@ -1,24 +1,21 @@
 <h2>Create New Course</h2>
 
-<%= form_for :newCourse, url: courses_path do |f| %>
-
+<%= form_for :newCourse, url: courses_path, builder: FormBuilderWithDateTimeInput do |f| %>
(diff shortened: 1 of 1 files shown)


dlbucci:
I think it's fine to add the bootstrap manually, we do that elsewhere in the codebase, but I would just use HTML for anything that isn't the text field.  we haven't used content_tags anywhere else.

jez:
@cjford: If you could update this to use the help text from the "Course settings" page, especially as it relates to the `:name` field I could merge this. This would address the issue that was raised in #550, where it was unclear that `:name` has to be a URL-able string.

cjford:
Added help text to name & semester fields

jez:
Cool, thanks! I'll merge this in.

PR #539, opened 2015-07-22
Exception Notifier 4.1 has been released

cg2v:
it is no longer necessary to use git version to get slack support


dlbucci:
Is this something that needs to skip develop, or can we merge there and not have to merge master into develop later?

ymzong:
@dlbucci: I'd say it's a better idea to pull to `develop` in this case, but since this is the first time for @cg2v  (and I assume he's also busy), how about we allow it for this time?

@cg2v: Thanks for the PR! Just FYI, in general we pull non-urgent changes to `develop` first, which is periodically flushed to `master`. :smile_cat:

PR #540, opened 2015-07-27
SVN now affects submissions

dlbucci:
I will test this more when it's not 3:18am, but Autolab will now do SVN validations and attempt to checkout work instead of just pretending that an SVN submission is a normal submission.

More work needs to be done, specifically for groups, which I'm 80% certain won't work quite the way users would expect them to (currently, they checkout every users own repository, vs sharing the repo of the submitting user).  I'll fix that at some point.

### DIFF
--- a/app/controllers/assessment/handin.rb
+++ b/app/controllers/assessment/handin.rb
@@ -24,7 +24,11 @@ module AssessmentHandin
 
     # save the submissions
     begin
-      submissions = saveHandin(params[:submission])
+      if @assessment.has_svn
+        submissions = svn_create_sub(@cud) { |sub| saveHandin(sub) }
+      else
+        submissions = saveHandin(params[:submission])
+      end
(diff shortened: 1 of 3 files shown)


jez:
@dlbucci I don't know enough about how this part of Autolab works to test this at all.

dlbucci:
We should probably merge this in before v1.7.0, or with v1.7.0

icanberk:
@dlbucci how can we test this? I have no idea either :/

dlbucci:
Write an assessment config module with a `checkoutWork(fromDir, toDir)` that somehow moves files in the `fromDir` to the `toDir`.  Click checkout and make sure the files in the `fromDir` wind up in the handin tar.

dlbucci:
PING

remzr7:
Removing SVN and hence closing issue

PR #542, opened 2015-07-28
Update README.md

bdwolfe:
Added the step about editing config/initializers/devise.rb.template


dlbucci:
Just a note: our steps are numbered incorrectly, but will still render correctly because markdown allows for that.  We should correctly order them at some point.

jez:
So now that we have three different setup instructions (development, production+Ubuntu, production+Docker), I'd really like to link to the wiki instead of having setup instructions in the README. We can certainly make this link stand out a bunch, but I think that moving people into the Wiki is a good idea, because that's where our documentation will be eventually.

I know that I was in the minority the last time I expressed interest to do this, but I still think that it's a good idea, particularly now that the Docker setup has landed. Thoughts?

PR #551, opened 2015-08-10
Allow for different site footers on home pages

cg2v:
Display messages describing other Carnegie Mellon autolab sites on both the courses index page and the no_user page. 

Use the applications hostname to decide what messages to display, since autolab.andrew.cmu.edu will need to include references to autolab.cs.cmu.edu

This change intentionally targets master, since (a) autolab.andrew.cmu.edu will not be running code based on develop this fall, and (b) develop has reworked the no_user page significantly.

### DIFF
--- a/app/views/home/_topannounce.html.erb
+++ b/app/views/home/_topannounce.html.erb
@@ -1,0 +1,26 @@
+<% if @mysite == "autolab.cs.cmu.edu" %>
+<h1>Older Courses</h1>
+
+<p>Courses on Autolab prior to Spring 2015 are viewable at oldautolab.cs.cmu.edu.</p>
+
+<%= link_to 'Old Autolab Courses', 'https://oldautolab.cs.cmu.edu', class: 'btn main' %>
+<% end %>
+
+<% if ["autolab.andrew.cmu.edu", "autolab-dev.andrew.cmu.edu"].include?(@mysite)  %>
+<h1>Other Courses</h1>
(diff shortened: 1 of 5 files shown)


app/views/home/_topannounce.html.erb
```diff
@@ -0,0 +1,26 @@
+<% if @mysite == "autolab.cs.cmu.edu" %>
```
jez: Do we maybe want to have this file be `app/view/home/_topannounce.html.erb.template` and have this file be gitignored? Otherwise, people will get merge conflicts if they try to update this file and we release an update.

jez:
It looks like you're working off the `master` branch. We do all of our development on the `develop` branch and use `master` exclusively for releases and hotfixes.

I wouldn't normally raise a concern about this, except that I'm pretty sure some of the files (like home/no_user.html.erb) are ahead in `develop`. To avoid merge conflicts, it'd be great if you could rebase and resubmit the patch against `develop`.

cg2v:
As my comment said, I intentionally submitted this against master because of the no-user change (where should this content even go on the new no-user page, which has much more content than the old one?). What I didn't say is that intended it as an initial proposal, and that the developers should pick a solution, implement it, and close this request.

jez:
Oh, that's my fault; I jumped hastily to the diff. Thanks for this! I'll take it over from here.

jez:
See #574

PR #552, opened 2015-08-12
fix first_name instead of email in confirmation email, forgot password

yesyayen:
replaced @resource.first_name with @resource.email


app/views/devise/mailer/confirmation_instructions.html.erb
```diff
@@ -1,4 +1,4 @@
-<p>Welcome <%= @email %>!</p>
+<p>Welcome <%= @resource.first_name %>!</p>
```
jez: We've had some issue in the past that bit us because not every user had a display name: can you change this to `@resource.first_name || @resource.email`?

nayak16:
Merged 00222c1ca47f8c0194d35e8feb945bc8fe3302cd

PR #554, opened 2015-08-16
Enhance annotations

icanberk:
Upon @droh's feedback, I've made some improvements to the PDF annotations that improves the UX of this feature. Although there is still a lot of room for improvement, I think it feels much better now.

- Makes annotations resizable, and adds a border around them
- Makes annotations draggable
- Removes the score, and problem from the chrome and adds it inside the comment box. This makes the - UI look less cluttered and also includes the score/problem information on the rendered PDF.
- Makes edit form larger so it's easier to edit longer comments

### DIFF
--- a/app/assets/javascripts/annotations.js
+++ b/app/assets/javascripts/annotations.js
@@ -19,12 +19,13 @@ $("#highlightLongLines").click(function() {
 });
 
 var initializeAnnotationsForCode = function() {
-  
+  window.annotationMode = "Code";
+
   var block = document.getElementById('code-block');
   hljs.highlightBlock(block);
 
   // annotationsByLine: { 'lineNumber': [annotations_array ]}
-  var annotationsByLine = {};
+  annotationsByLine = {};
   _.each(annotations, function(annotationObj, ind) {
     var lineInd = annotationObj.line
     if (!annotationsByLine[lineInd]) {
@@ -438,8 +439,9 @@ var initializeAnnotationsForCode = function() {
       type: "text",
       name: "comment",
       placeholder: "Comments Here",
-      maxlength: "255"
-    }, commentStr);
+      maxlength: "255",
+      value: commentStr
+    });
 
     if (annotationMode === "PDF") {
       var commentInput = elt("textarea", {
--- a/app/assets/javascripts/annotations.js
+++ b/app/assets/javascripts/annotations.js
@@ -579,6 +579,9 @@ var initializeAnnotationsForPDF = function() {
 
     $(annotationEl).css({ "left": xCord + "px",  "top" : yCord + "px", "position" : "absolute" });
     $("#page-canvas-wrapper-"+pageInd).append(annotationEl);
+    console.log("making draggable")
+    $(annotationEl).draggable();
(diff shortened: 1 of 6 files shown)


jez:
Another trigger happy merge lol

app/assets/javascripts/annotations.js
```diff
@@ -19,12 +19,13 @@ $("#highlightLongLines").click(function() {
 });
 
 var initializeAnnotationsForCode = function() {
-  
+  window.annotationMode = "Code";
```
jez: It'd probably be good to have a master list/enum of all the constants that `annotationMode` can take on declared at the top of the file.

app/assets/javascripts/annotations.js
```diff
@@ -163,6 +164,72 @@ var initializeAnnotationsForCode = function() {
...
+    }
+
+    var body = elt("div", {
+      class: "body"
+    }, commentStr);
+
+    var box = elt("div", {
+      class: "ann-box",
+      id: "ann-box-" + annObj.id
+    }, header, body, score);
```
jez: Instead of reconstructing this box every time we need to accept user input, would it be easier to just have a `div` always in the page that is only visible when someone has clicked? It might make the JavaScript code look cleaner. We could populate the values as needed by manually setting the innerHTML.

app/assets/javascripts/annotations.js
```diff
@@ -163,6 +164,72 @@ var initializeAnnotationsForCode = function() {
...
+
+    $(del).on("click", function(e) {
+      $.ajax({
+        url: deletePath(annObj),
```
jez: Where is `deletePath` defined?

jez: Oh, I guess maybe it's from a previous PR. Disregard.

app/assets/javascripts/annotations.js
```diff
@@ -364,6 +442,17 @@ var initializeAnnotationsForCode = function() {
       maxlength: "255",
       value: commentStr
     });
+
+    if (annotationMode === "PDF") {
+      var commentInput = elt("textarea", {
+        class: "col-md-12 comment",
+        type: "text",
+        name: "comment",
+        placeholder: "Comments Here",
```
jez: A more inviting placeholder text couldn't hurt :P

app/controllers/submissions_controller.rb
```diff
@@ -199,16 +202,30 @@ def download
 
           position = annotation.coordinate.split(',')
           page  = position[2].to_i
+
+          width = position[3].to_f * pdf.bounds.width
```
jez: Would you mind assigning these array values to variables, or at least write a comment mapping out what values they contain.

app/views/submissions/viewPDF.html.erb
```diff
@@ -2,6 +2,8 @@
 
 <% content_for :stylesheets do %>
 <%= stylesheet_link_tag "annotations" %>
+<%= external_stylesheet_link_tag "jquery-ui", "1.11.4" %>
```
jez: Do we not already include jquery-ui?

PR #562, opened 2015-08-23
Migrating nicknames with non-ASCII characters

icanberk:
This stops the migration since we have started asserting that the nicknames are non-ascii with this version. I'm not good at regexp, so I'm open to improvements but this worked on production and everything is functional now.

### DIFF
--- a/db/migrate/20150403173410_add_has_joined_to_cud.rb
+++ b/db/migrate/20150403173410_add_has_joined_to_cud.rb
@@ -10,6 +10,12 @@ class AddHasJoinedToCud < ActiveRecord::Migration
 
     CourseUserDatum.find_each do |cud|
       cud.has_joined = true
+
+      if !cud.nickname.nil? then
+        # if there is a non-ascii nickname, we migrate it too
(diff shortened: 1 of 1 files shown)


dlbucci:
http://ruby-doc.org/core-2.2.0/String.html#method-i-ascii_only-3F

jez:
@dlbucci: That just returns a boolean; don't we need to convert the contents of the string in this migration?

dlbucci:
Didn't notice the latter half, so this is probably fine.

jez:
Alright @icanberk. I'll leave merging and deploying this to you. I don't want to mess with the prod environment right now.

jez:
Let's try to move away from using the GitHub merge feature so we can get a more linear Git history, which makes it easier to revert things in a pinch.

icanberk:
Sorry I realized that after merging. Did you write a note somewhere about how to merge properly?

jez:
Yeah, I have it in CONTRIBUTING.md locally, but I was going to wait for `master` to settle down before opening up a PR.

PR #563, opened 2015-08-24
Fix the problems with adding Tweak on CUD + auto join if admin

icanberk:


jez:
What's the general idea here? (Also, is this a WIP?)

icanberk:
Creating a blank Tweak was causing an error and we were never saving the cud. Now we create a cud and automatically confirm the join request for an admin. (Will write something longer about how this feature works soon)

Should be merged.

jez:
Go ahead and do it. Make sure to

- do a fast-forward merge (using the command line)
- push the update to `develop`

Then you can deploy to prod as necessary.

dlbucci:
I don't understand why we always save the newCUD?

jez:
Closed in favor of reverting public courses because it fixes other errors. We can come back to this with a version of public courses that's more well-tested.

PR #564, opened 2015-08-24
Updated CONTRIBUTING to reference new Git workflow

jez:
Check this one out, there are a few simple workflow changes here. This
should make it easier to revert changes quickly when users report that
something's broken that didn't used to be.

/cc @team

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,23 +3,34 @@
 
 ## Reporting Bugs
 
-1. Always update to the most recent master release; the bug may already be resolved.
+1. Always update to the most recent master release; the bug may already be
+   resolved.
 
-2. Search for similar issues on the [Issue Tracker](https://github.com/autolab/Autolab/issues); it may already be an identified problem.
+2. Search for similar issues on the [Issue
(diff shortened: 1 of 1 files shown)


jez:
Also (just throwing it out there because we're addressing process-type stuff now): How do people feel about the `develop`/`master` split? Is it working? Should we tweak or drop it?

dlbucci:
I'm not certain I like the `develop/master` split anymore.  Code that gets into develop might be tested a little more than code going straight to master, but only if other people do development from develop after the code has been merged, which doesn't happen.  It seems like we merge a bunch of stuff into develop, then immediately merge develop into master with each release, which sort of defeats the point.

That's also the other problem: the develop system tends to put a few big changes into production at the same time, which greatly increases the error probability.  I'd say we should switch to using just master and test our pull requests more diligently (i.e. test them at all).

jez:
I'm in favor of this as well. We can still mark specific commits as stable releases, but we will no whether things are broken much sooner. We'll also be able to iterate faster.

However, I think we should still be diligent about deploying `master` to prod immediately after merging in a PR. The `accepted` label along with the command-line-only merges should encourage this behavior.

I'll keep this PR open for a while longer so people see it and can still discuss it.

PR #566, opened 2015-08-24
Add some uniqueness checks to models

cg2v:
Several models should have uniqueness constraints. Missing uniqueness
constraints can trigger mysql errors when the database has unique keys
for some columns.


cg2v:
Forgot to switch to develop

PR #567, opened 2015-08-24
Fix tests (for sqlite?).

cg2v:
find_user was returning Nil even if the database had been populated, since
0 and false are not the same thing.

Also fix the home#no_user test to look for text that exists there now


jez:
Closed via b9713f2

PR #568, opened 2015-08-24
Add some uniqueness checks to models

cg2v:
Several models should have uniqueness constraints. Missing uniqueness
constraints can trigger mysql errors when the database has unique keys
for some columns.

### DIFF
--- a/app/models/problem.rb
+++ b/app/models/problem.rb
@@ -10,7 +10,7 @@ class Problem < ActiveRecord::Base
   belongs_to :assessment, touch: true
   has_many :annotations
 
-  validates :name, presence: true, uniqueness: {scope: :assessment_id}
+  validates :name, presence: true, uniqueness: { scope: :assessment_id }
   validates_associated :assessment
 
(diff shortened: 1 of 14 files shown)


app/models/problem.rb
```diff
@@ -10,7 +10,7 @@ class Problem < ActiveRecord::Base
   belongs_to :assessment, touch: true
   has_many :annotations
 
-  validates :name, presence: true
+  validates :name, presence: true, uniqueness: {scope: :assessment_id}
```
dlbucci: Can you check these files with rubocop?  I think validates with a :uniqueness argument is the preferred way to add this check.  If so, I think the other files should be edited to use that method instead, just to keep the code slightly more up to date with conventions.

cg2v:
The only thing it didn't like was the (lack of) spaces on the line you quoted

dlbucci:
Agreed, it looks better with the spaces :)

nayak16:
Develop branch deleted. Will add uniqueness checks in new branch

PR #571, opened 2015-08-25
Added a root_path

dlbucci:
Didn't test, but this should shut the errors up.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -16,6 +16,7 @@ Autolab3::Application.routes.draw do
 
   end
 
+  root "courses#index"
 
   get "contact", to: "home#contact"
 


jez:
How can we be sure this won't break something? Do we know where the errors are coming from that rely on `root_path` being set?

It looks like this was the diff that public course instances introduced. If we merge this diff, `root "course#index"` will be in there twice. Is that what we want?

```diff
diff --git b/config/routes.rb a/config/routes.rb
index 0aedc69..26b2adf 100755
--- b/config/routes.rb
+++ a/config/routes.rb
@@ -1,10 +1,22 @@
 Autolab3::Application.routes.draw do
-  root "courses#index"
 
   devise_for :users, controllers: { omniauth_callbacks: "users/omniauth_callbacks",
                                     registrations:      "registrations" },
                      path_prefix: "auth"
 
+  devise_scope :user do
+
+    authenticated :user do
+      root 'courses#index', as: :authenticated_root
+    end
+
+    unauthenticated do
+      root 'devise/sessions#new', as: :unauthenticated_root
+    end
+
+  end
+
+
   get "contact", to: "home#contact"
 
   namespace :home do
@@ -22,6 +34,7 @@ Autolab3::Application.routes.draw do
   end
 
   resources :courses, param: :name do
+
     resources :schedulers
     resources :jobs, only: :index do
       get "getjob", on: :member
@@ -122,6 +135,7 @@ Autolab3::Application.routes.draw do
     end
 
     resources :course_user_data do
+
       resource :gradebook, only: :show do
         get "bulkRelease"
         get "csv"
@@ -132,6 +146,7 @@ Autolab3::Application.routes.draw do
       end
 
       member do
+        post "confirm"
         get "destroyConfirm"
         match "sudo", via: [:get, :post]
         get "unsudo"
@@ -142,6 +157,7 @@ Autolab3::Application.routes.draw do
       get "bulkRelease"
       get "downloadRoster"
       match "email", via: [:get, :post]
+      post "join"
       get "manage"
       get "moss"
       get "reload"
```

dlbucci:
Will that even matter? I don't think multiple root declarations will break the routes file, and that will only happen if we revert public courses.

jez:
Closed in favor of reverting public courses because it fixes other errors. We can come back to this with a version of public courses that's more well-tested.

PR #574, opened 2015-08-25
Configurable announcement partial for '/'

jez:
This lets people configure what shows up on the courses index at a
per-deployment level. This solves what we need for both
autolab.cs.cmu.edu as well as autolab.andrew.cmu.edu in a way that's
flexible enough to be used by others.

__Summary__:

- Based off of @cg2v's original PR (#551)
  - Thanks for the initial work!

- Rename topannounce partial to .template, add to gitignore

- Remove `@mysite` variable
  - Instead of having the `@mysite` variable keep track of what should
    be displayed, we'll just always show the _topannounce message, and
    write a different one on every host where it's necessary.

    If the template doesn't exist, we don't render a message on that host.

- Remove social buttons from _footer

  - We can move these to the CMU-specific _topannounce file so that
    people can opt into having these on their site. Also, it restricts
    it to the courses page, so it's less intrusive on the user
    experience.

__TODO (after landing)__:

- [ ] Remember to add this to CMU _topannounce after landing:

  ```html
  <h1>Think Autolab's awesome?</h1>
  <p>Help spread the word!</p>
  <div>
    <a class="github-button" href="https://github.com/autolab/Autolab" data-count-href="/autolab/Autolab/stargazers" data-count-api="/repos/autolab/Autolab#stargazers_count" data-count-aria-label="# stargazers on GitHub" aria-label="Star autolab/Autolab on GitHub">Github</a>
    <br>
    <div class="fb-like" data-href="https://www.facebook.com/autolabproject" data-width="300" data-layout="standard" data-action="like" data-show-faces="false" data-share="false"></div>
  </div>
  ```

- [ ] Update documentation in wiki showing how to use _topannounce

### DIFF
--- a/app/views/courses/index.html.erb
+++ b/app/views/courses/index.html.erb
@@ -31,8 +31,4 @@
   <% end %>
 <% end %>
 
-<h1>Older Courses</h1>
-
-<p>Courses on Autolab prior to Spring 2015 are viewable at oldautolab.cs.cmu.edu.</p>
-
-<%= link_to 'Old Autolab Courses', 'https://oldautolab.cs.cmu.edu', class: 'btn main' %>
+<%= render partial: "home/topannounce" rescue nil %>
--- a/app/views/home/_topannounce.html.erb.template
+++ b/app/views/home/_topannounce.html.erb.template
@@ -1,0 +1,20 @@
+<%#
+  Use this file as an opportunity to broadcast a permanent message on
+  the root '/' path at the bottom of the page. This is normally the page
+  with a list of all courses.
+
+  For example, at CMU we use it to provide students and instructors with
+  a convenient link to an older deployment of Autolab that exists solely
+  for archival purposes. Uncomment the below, rename this file to drop
+  the `.template` suffix, and nagivate to Autolab in your browser to see
+  the effect.
+%>
+
+<%# REMOVE THIS LINE TO UNCOMMENT
+
+<h1>Older Courses</h1>
+<p>Courses on Autolab prior to Spring 2015 are viewable at oldautolab.cs.cmu.edu.</p>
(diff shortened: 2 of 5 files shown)


app/views/courses/index.html.erb
```diff
@@ -31,8 +31,4 @@
   <% end %>
 <% end %>
 
-<h1>Older Courses</h1>
-
-<p>Courses on Autolab prior to Spring 2015 are viewable at oldautolab.cs.cmu.edu.</p>
-
-<%= link_to 'Old Autolab Courses', 'https://oldautolab.cs.cmu.edu', class: 'btn main' %>
+<%= render :partial => "home/topannounce" rescue nil %>
```
dlbucci: Can this call (and the later one) use the new hash symbol syntax? (`partial: "home/topannounce"`)

app/views/home/_topannounce.html.erb.template
```diff
@@ -0,0 +1,16 @@
+<%#
+  Use this file as an opportunity to broadcast a permanent message on
+  the root '/' path at the bottom of the page. This is normally the page
+  with a list of all courses.
+
+  For example, at CMU we use it to provide students and instructors with
+  a convenient link to an older deployment of Autolab that exists solely
+  for archival purposes.
+%>
```
dlbucci: I think this explanation should also say that you need to rename the file for this to work.  Also, I don't know if our older courses html should be included in this file if we're trying to remove CMU-specific stuff.  If you want to leave it as an example, maybe leave a comment to explain that it's an example/comment the code out.

dlbucci:
Aside from my two comments, this looks good and can be merged.

jez:
Cool, I updated it. Let me know if anything else stands out.

dlbucci:
This looks good.  You can merge, or I can merge it in about 9 hours.

PR #534, opened 2015-07-09
Make quiz feature not suck

icanberk:
@droh will be using this in production for the first assignment of 2013. As the first step, I re-enabled the button and fixed the quiz creation issue so he can start working on the assignment and we get feedback. 
I will be improving the UI/UX of the page over the weekend. Any contributions would be welcome.

PR #304, opened 2015-02-20
Fix quiz + Assessment Builder

icanberk:

PR #575, opened 2015-08-27
Revert quiz feature

jez:
Reverts quiz feature entirely, with the exception of the changes that
were made to the models--any quiz-related change to models or migrations
remains.

Reverts #47
Reverts #304
Reverts #534


s-koch:
Hi Autolab Team,
we are interested in the quiz feature and consider to implement this ourselves. Could you maybe explain why this feature was finally removed from the system? Were there any special difficulties that made an implementation too difficult/bulky? Can we start out with what is/was already there or would you recommend to start from scratch?

Thanks

ymzong:
Hey @s-koch, we eventually decided to revert the feature because it was acting pretty flaky. Implementing a complete quiz feature in Autolab is non-trivial (especially if it's to be designed carefully, i.e. with nice UI, time limit, etc.), and I would recommend the following workaround:

You can create an assessment for the quiz, add the quiz questions as an attachment (handout), and ask students to submit the answers to the quiz in a plain text file, like the following:

```
12
A
0x00c0ffee
```

You then may first write an autograder that validates the student's submission (to make sure that their submissions can be parsed), and after the quiz deadline release a final autograder that actually grades students' responses.

Does the above sound reasonable to you?

s-koch:
Hi @ymzong, thank you for the extensive answer and the workaround suggestion. One of the features we need to have for the quiz is the random permutation of questions for each student. I guess this can't be easily done with the assignment workaround?

Issue #576, opened 2015-08-29
Very inconsistent behaviour when setting assessment due time


GjjvdBurg:
Hey guys,

I'm experiencing some very inconsistent behaviour when setting the assessment times. I'm not an expert in Ruby so I haven't been able to find the relevant code which could be responsible, but I think that the locale time may also have an influence.

Let's say I have an assessment which currently has a due time of 7pm, in the main assessment window:

![screenshot - 290815 - 19 58 21](https://cloud.githubusercontent.com/assets/3286696/9563607/1e83764c-4e89-11e5-9239-806a5f8b881c.png)

Now, if I check the "Edit Assessment" Page, the due time I see there is 5pm:

![screenshot - 290815 - 19 58 45](https://cloud.githubusercontent.com/assets/3286696/9563611/3673c676-4e89-11e5-838d-8777de2d2ea5.png)

This is a 2 hour difference which could be caused by the difference with UTC, which is currently 2 hours behind my local time. If I try to set the due time at 9am as desired using this screen:

![screenshot - 290815 - 19 59 21](https://cloud.githubusercontent.com/assets/3286696/9563620/7bdc5ca0-4e89-11e5-82b8-5a8934031f70.png)

As soon as I press "Save", the due time changes to 1pm:

![screenshot - 290815 - 20 00 37](https://cloud.githubusercontent.com/assets/3286696/9563622/8e60a782-4e89-11e5-9c1a-9bb1790598fa.png)

And back in the main assessment window the time is shown as 3pm:

![screenshot - 290815 - 20 00 58](https://cloud.githubusercontent.com/assets/3286696/9563623/99fd537e-4e89-11e5-9787-e40669a85325.png)

So, in order to have a displayed due time of 9am, I have to set the time to 3am in the configuration, and hope that the system will use the displayed due time instead of the one that will show up in the configuration after pressing Save (which is 7am).

Any help is appreciated since we'll start using Autolab here on Monday :)

dlbucci:
We've never had Autolab deployed in more than one time zone, so I wouldn't be surprised if it has some timezone inconsistency issues.  I can try to take a look at this sometime this weekend, but the relevant code should be in `app/controllers/assessments_controller.rb` and `app/models/assessment.rb`.

GjjvdBurg:
Thanks for the quick response, would be great if you could take a look at it. If you need some debugging output or something let me know.

dlbucci:
[...]

PR #577, opened 2015-08-30
Add Time Zone to Datepickers

dlbucci:
With this change, if the user's browser's time zone does not match the servers timezone, there shouldn't be skew in the date values the user sends to the server, so #576 should disappear.  The one downside is that the time zone value is now always present in the input box, which is a bit ugly, but is worth it.


dlbucci:
For the record, this appears to have fixed #576, so we can close that when this lands.

jez:
:+1:

jez:
This should now be live.

PR #585, opened 2015-09-12
Remove AUTOCONFIG_COURSE_DIR

### LINKED ISSUE #583: extract_tar_for_moss breaks later handins
extract_tar_for_moss chdir's to the temporary directory (even though I don't think it needs to - destination files are built up as absolute paths by File.join), which breaks any later uses of relative paths, such as autograding handins

Edit: I guess this only happens when AUTOCONFIG_COURSE_DIR is a relative path, so maybe it's not a big deal. the chdir should still be removed if it's not needed

dlbucci:
Fixes #583

### DIFF
--- a/app/controllers/assessment/autograde.rb
+++ b/app/controllers/assessment/autograde.rb
@@ -183,7 +183,7 @@ module AssessmentAutograde
   #
   def tango_upload(course, assessment, submission, existing_files)
     # first, figure out which files need to get sent
-    ass_dir = File.join(AUTOCONFIG_COURSE_DIR, course.name, assessment.name)
+    ass_dir = assessment.folder_path
     begin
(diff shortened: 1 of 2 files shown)


jez:
Since this makes changes to a config file, we should probably prefix the commit with something like `[config]` so we remember to make a note of it in the release changelog.

Are there any other ways this could mess up things in prod? Do we mention `AUTOCONFIG_COURSE_DIR` in the docs anywhere?

dlbucci:
This shouldn't mess up prod on restart.  We should delete the line from the config file, but it won't break anything by being there, and we may need to update our travis stuff.  A search of the repo indicates that these three places are the only places that string appears in our code base.

ymzong:
@dlbucci: It looks great to me -- feel free to merge it!

jez:
Merged via d3804ab

PR #592, opened 2015-09-17
Update how we parse tango.info response

ymzong:
Supports the fix at autolab/Tango#96.
The response changes from a list of `var_name=value` to a JSON dictionary.


dlbucci:
LGTM

ymzong:
Reproduced #613

PR #594, opened 2015-09-21
Adds the ability to display PDF files residing in tar/tgz submissions

icanberk:
This corrects the fileLink we use to get the PDF file when we want to display a PDF file on the browser.

- Tested with one level deep tar submission.
- Tested that normal .pdf submissions still work as expected.

### DIFF
--- a/app/views/submissions/viewPDF.html.erb
+++ b/app/views/submissions/viewPDF.html.erb
@@ -31,9 +31,9 @@
     var annotations = <%=raw @annotations.to_json %>;
     var fileNameStr = "<%= @filename %>";
     <% if @preview_mode then %>
-    var fileLink    = "<%= download_course_assessment_submission_path(@course, @assessment, @submission, :annotated => true) %>";
+    var fileLink    = "<%= url_for [:download, @course, @assessment, @submission, header_position: params[:header_position], annotated: true] %>";
     <% else %>
-    var fileLink    = "<%= download_course_assessment_submission_path(@course, @assessment, @submission) %>";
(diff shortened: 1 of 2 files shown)


app/views/submissions/viewPDF.html.erb
```diff
@@ -31,9 +31,9 @@
     var annotations = <%=raw @annotations.to_json %>;
     var fileNameStr = "<%= @filename %>";
     <% if @preview_mode then %>
-    var fileLink    = "<%= download_course_assessment_submission_path(@course, @assessment, @submission, :annotated => true) %>";
+    var fileLink    = "<%= download_course_assessment_submission_path(@course, @assessment, @submission, :header_position =>params[:header_position], :annotated => true) %>";
     <% else %>
-    var fileLink    = "<%= download_course_assessment_submission_path(@course, @assessment, @submission) %>";
+    var fileLink    = "<%= download_course_assessment_submission_path(@course, @assessment, @submission, :header_position =>params[:header_position]) %>";
```
dlbucci: I would use the new symbol syntax here.  Also, does this fix the downloading submission errors?  If not, I would hide the download link until that is fixed.

dlbucci:
So was the missing problem name the issue here? That wasn't really what the error messages were pointing to, but I could see that being the case.  If you update the hash syntax (`:blank =>` becomes `blank:`), this can be merged.

icanberk:
Problem was that we didn't pass the "header_position"(I didn't even knew what this meant) variable while generating the fileLink JS variable. It gets the right link to display as well as the right link to download now.

PR #596, opened 2015-09-24
Ensure that volume directories are app-writable

### LINKED ISSUE #560: Permission issues when following documentation
If I follow the directions at https://github.com/autolab/Autolab/wiki/Deploying-Autolab-with-Docker

When I create a new course I get the following error:

```
--- !ruby/exception:Errno::EACCES
message: Permission denied @ dir_s_mkdir - /home/app/webapp/courses/123
```

If I manually correct the problem, I then get an error about /home/app/webapp/assessmentConfig.  If I `chown -Rv app:app /home/app/webapp`, the problem goes away.

### LINKED ISSUE #579: Dockerfile does not set permissions for tmp/cache
I receive the following error when visiting most pages:

`--- !ruby/exception:Errno::EACCES
message: Permission denied @ dir_s_mkdir - /home/app/webapp/tmp/cache/206`

jez:
These are all the folders mounted as volumes that need to be able to
writable by the app.

This is linked with autolab/autolab-docker#2 which ensures that these
folders are always created and as such __must be merged after that PR
merges__.

Fixes #560. Fixes #579.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,6 +36,13 @@ RUN bundle install
 # Add the rails app
 ADD . /home/app/webapp
 
+# Fixup permissions on courses folder
+RUN chown -R app:app /home/app/webapp/courses \
+    /home/app/webapp/assessmentConfig \
+    /home/app/webapp/courseConfig \
+    /home/app/webapp/gradebooks \
+    /home/app/webapp/tmp
+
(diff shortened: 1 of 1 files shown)


jez:
This doesn't actually accomplish what I thought it did.

PR #601, opened 2015-10-10
Bump Ruby version in README

AEgan:
The app uses 2.2.0 but the README tells developers to install 2.1.2

Also removes trailing whitespace


ymzong:
LGTM, thanks for the PR! :smile:

PR #607, opened 2015-11-03
Fix nested function call when setting up course

cg2v:
Don't say
FileUtils.mkdir_p (File.join Rails.root, "foo"), since that's
confusing style. Instead put parens around the File.join arguments:
FileUtils.mkdir_p File.join(Rails.root, "foo")

This fixes some rubocup warnings:
app/models/course.rb:42:22: W: (...) interpreted as grouped expression.
    FileUtils.mkdir_p (File.join Rails.root, "assessmentConfig")

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -33,15 +33,15 @@ class Course < ActiveRecord::Base
     course_dir = File.join Rails.root, "courses", name
     FileUtils.mkdir_p course_dir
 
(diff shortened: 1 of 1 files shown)


dlbucci:
These could be further simplified from `File.join(Rails.root, "foo")` to `Rails.root.join("foo")`.  I've been doing that wherever I've seen it, so if you want to update to that, we can merge this.

PR #608, opened 2015-12-10
Fixes for Annotation Alignment Issues

icanberk:
- Make list items use position absolute
- Expand the line number area to accomodated 4 digit numbers


ymzong:
LGTM! :smile_cat:

PR #619, opened 2016-02-04
Send bulk email to past and current instructors

### LINKED ISSUE #599: Need to capability to send bulk email to all Autolab instructors, past and present
According to yzong, the current bulk email sends email to instructors in current classes.  For the CMU autolab, we'll need to be able to email all Autolab instructors past and present, even if they're not using Autolab in the current semester.

nayak16:
Fix #599

### DIFF
--- a/app/controllers/admins_controller.rb
+++ b/app/controllers/admins_controller.rb
@@ -9,7 +9,7 @@ class AdminsController < ApplicationController
   action_auth_level :email_instructors, :administrator
   def email_instructors
     @cuds = CourseUserDatum.select(:user_id).distinct.joins(:course).joins(:user)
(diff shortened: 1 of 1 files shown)


ymzong:
LGTM! Just one comment: Do you think sometimes it is necessary to _only_ send mail to current instructors, or is it fine to always send emails to all instructors?

If you think it's fine to always email all instructors, feel free to merge it!

nayak16:
I think it makes sense to send to all instructors. For example, for 15-150, the professors switch every semester. So it would be advantageous to keep the professor not teaching in the loop about Autolab changes.

ymzong:
:+1:

PR #621, opened 2016-02-05
Hide submit file button if assessment handin disabled

### LINKED ISSUE #548: Disabling handins has no impact on assessment page
If handins are disabled in an assessment, Autolab still displays the "Submit File" button.

nayak16:
Fixes #548 

This is the only place this change should be made correct? Just making sure I'm not missing anything.


ymzong:
Yup that's the only spot, LGTM!

Nitpicking: http://stackoverflow.com/questions/27429711/can-someone-explain-not-vs-in-ruby

nayak16:
Merged 4211457d49018cd1f107f083bc8adeb10f68bdb6

PR #623, opened 2016-02-11
Don't return HTML errors to non-html requests

cg2v:
API requests (e.g. annotations) should not try to display
donkey kong when they fail, since that just triggers another
error


cg2v:
untested

zyx-billy:
merged into master with #735

PR #624, opened 2016-02-11
Don't return deleted annotation

cg2v:
When an annotation is deleted, don't return the deleted object, just
an HTTP 204 response


cg2v:
I don't know if this actually fails, but it seems wrong to return the object we just destroyed.

Issue #631, opened 2016-02-22
File re-submission fails after pressing back button


nayak16:
Steps to recreate bug:
1. Submit a file for an assessment
2. Press back button on browser
3. Attempt to resubmit same file
4. Re-submission fails unless a page refresh happens

My guess is that the problem is a caching issue with the file name since submitting another file works.
Best place to start would be `app/controllers/assessment/handin.rb`

ymzong:
I'm guessing that it's more like a front-end issue. The submission button has a protection mechanism that prevents anyone from submitting a same file twice. That's why when you submit some file (say `a.tar`), BACK, then submit the same file again, the browser knows it's the same file and prevents it.

nayak16:
Is this still an issue though? Or something that cannot be prevented?

ymzong:
Checking to see if #641 fixes this.

ymzong:
Fixed by #641.

PR #641, opened 2016-03-16
Clear the submit field after submitting a file

WChargin:
Summary:
This is good practice for submit forms in general,
because it makes the Back button work properly
(and re-selections in general).
The Autolab-specific workflow is as follows.
Frequently, I submit a file,
then immediately remember some minor changes
(as one does).
So I let it submit, then hit "Back,"
then open the submit file dialog again.
But now the file is already selected,
so if I select it again no change event is fired.
The solution is to just clear the value.

Test Plan:
**I have not tested this change to the CoffeeScript file.**
What I did do was to enter the line that I added
into the Chrome JS console,
then click the Submit File button and submit a file,
then click Back and submit the same file again.
This worked.
So the JS is fine;
the question is whether I integrated it properly with Autolab.


ymzong:
Fixes #631. Thanks @WChargin! :+1:

PR #643, opened 2016-03-16
Faster Gradesheet (on-demand feedback and deferred rendering)

### LINKED ISSUE #635: Grade All Submissions page should load Autograder output on-demand
Currently, the source of the "Grade All Submissions" page contains the Autograder output for every student's submission, repeated (# of problems) times. This causes the page to load quite slowly, sometimes taking over a minute for large classes or when the autograder feedback is long.

In order to fix this, the autograder output should be removed from the page source, and clicking on a score should cause an AJAX request to retrieve the autograder output.

zyx-billy:
Aims to resolve #635.

1. On-demand feedback
Feedback is no longer embedded in the gradesheet html. Individual feedback is fetched from server automatically when user clicks on a score.

2. Deferred rendering
To make the page load faster, only the first 40 rows will be displayed initially. Whenever the user scrolls near the bottom, more rows are appended to the table. The non-visible rows are not even inserted into the DOM until it is needed for display.

An added bonus of deferring rendering is the initial html can be even smaller. Previously, all raw data was generated into the table as html, which means the same structure is being repeated over and over for each row, only differing in a few values. Now the data is stored in JS arrays, so we don't need to repeat the html structure for each row (It doesn't sound much but it turns out to be a lot => 77% size decrease for each row).

Overall Performance:
Performing this on a regular 122-size class decreased the size of the initial page from ~13MB to <500KB, which is much more realistic. The js for loading the table also finishes in an instant.

### DIFF
--- a/app/assets/javascripts/gradesheet.js.erb
+++ b/app/assets/javascripts/gradesheet.js.erb
@@ -185,10 +185,9 @@ jQuery(function() {
 
     // always called when editor's closed, the hub for which is current_editor.reset()
     function _close_current_editor() {
-
+        $(current_editor).parent().removeClass("focus");
         close_current_popover();
         current_editor = null;
-
     }
 
     function close_current_editor() {
@@ -224,6 +223,10 @@ jQuery(function() {
         setTimeout(f, 0); 
     }
 
+    /* opens the score editor and feedback popover for a particular score
+     *
+     * @param {element} editable The div.editable element clicked on by the user
+     */
     function open_editor(editable) {
         // if we're already open, we're done
         if (current_editor == editable) return;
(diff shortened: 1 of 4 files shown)


ymzong:
Let's do another demo this Thursday!

zyx-billy:
Sure!

zyx-billy:
Just updated the pr. There's now a default sorting based on emails, and the rows don't jump around anymore when scrolling.

ymzong:
Cool! Let's merge this in the next meeting.
Gradesheet is really important, and we want to avoid any possible glitches since Computing Services will deploy a new version in two weeks (including this PR).

Issue #632, opened 2016-02-22
Assessment name shouldn't start with number


ymzong:
According to a bug report, an instructor was able to create an assessment whose name begins with a number (e.g. `15000assignmentone`). This is unsupported as a Ruby object name, and we need to block this on our front-end.

Relevant page: `https://AUTOLAB_HOST/courses/COURSE_NAME/assessments/new`

andersonleite:
Hi @ymzong 
How about adding a validation in the assesment.rb model?

`validates_format_of :name, :with => /^[a-zA-Z]+$/, :multiline => true`

This way we can block server receiving assessment starting with numbers and will make use of the actual notifications:

The result message in the frontend would be like that:

> Error saving 15000assignmentone: Validation failed: Name is invalid

Let me know if I can submit this PR.

ymzong:
Yep @andersonleite, that's exactly what I was thinking about. Go ahead and submit your PR!

remzr7:
@ymzong have you looked at this commit?

ymzong:
Yeah, it's already been pulled to master as cd88a445

PR #645, opened 2016-03-22
Assessment name shouldn't start with number #632

andersonleite:
Added regex validation in the assesment.rb model.

### DIFF
--- a/app/models/assessment.rb
+++ b/app/models/assessment.rb
@@ -21,6 +21,7 @@ class Assessment < ActiveRecord::Base
 
   # Validations
   validates_uniqueness_of :name, scope: :course_id
+  validates_format_of :name, :with => /^[a-zA-Z]+$/, :multiline => true
   validates_length_of :display_name, minimum: 1
   validate :verify_dates_order
   validate :handin_directory_and_filename_or_disable_handins, if: :active?
--- a/app/models/assessment.rb
+++ b/app/models/assessment.rb
@@ -21,7 +21,7 @@ class Assessment < ActiveRecord::Base
 
   # Validations
   validates_uniqueness_of :name, scope: :course_id
-  validates_format_of :name, :with => /^[a-zA-Z]+$/, :multiline => true
+  validates :name, format: { with: /\A[^0-9].*/, message: "can't have leading numeral" }
   validates_length_of :display_name, minimum: 1
   validate :verify_dates_order
   validate :handin_directory_and_filename_or_disable_handins, if: :active?


dlbucci:
I think the current rails best practice is to use `validates`, rather than the special `validate_*` methods, as seen here: https://github.com/autolab/Autolab/pull/568/files

Also, why is `multiline: true`?

andersonleite:
Hi @dlbucci 
So, how about going with this?

`validates :name, :format => {:with => /^[a-zA-Z]+$/, :multiline => true}`

Not sure why, but both with raise an error on rake autolab:update.

> bundle exec rake autolab:populate
> Creating Course AutoPopulated and config file
> Creating Assessments
> Creating Problems
> rake aborted!
> ActiveRecord::RecordNotSaved: You cannot call create unless the parent is saved

PS: :multiline - Set to true if your regular expression contains anchors that match the beginning or end of lines as opposed to the beginning or end of the string. These anchors are ^ and $.
http://api.rubyonrails.org/classes/ActiveModel/Validations/HelperMethods.html

-

dlbucci:
Just realized this regex can is incorrect.  A course/assessment name can have numerals, they just can't begin with numerals.  The line we want is here: https://github.com/autolab/Autolab/blob/master/app/models/course.rb#L11
Also, I don't think we need or want newlines/multiple lines in assessment names.  The reason for the leading numeral restriction is that the assessment name is used to create a ruby module, and ruby module names cannot have newlines AFAIK.

andersonleite:
@dlbucci 
Added validation line you suggested but travis build still falling. Any thoughts?

dlbucci:
We've had issues before with forked repo pull requests not being authorized for certain steps of the build process, which causes builds to fail for non-team members.  This might be that, but another member could probably confirm.

zyx-billy:
Yep. It seems the build failed when trying to access an environment variable. If I understood this correctly, we have [encrypted variables](https://docs.travis-ci.com/user/environment-variables/#Encrypted-Variables) in our .travis.yml, which is unavailable to forked repos.

This does seem a bit inconvenient though, @ymzong, do you think it's possible to have a different build process for external pr, without having to use the github private key?

ymzong:
@zyx-billy

Good point -- I'll patch the test script such that when the encrypted var is unavailable, the test is skipped instead of errors.

zyx-billy:
Sounds Cool! :+1:

andersonleite:
Thanks guys.
@ymzong 
Let me know if I need to do something more to close this PR.

ymzong:
Thanks for the contribution @andersonleite!

PR #651, opened 2016-04-06
Gradesheet save on change only

### LINKED ISSUE #446: Gradesheet graders should only change if the feedback/score changes.
Currently if a grader clicks a score, does nothing and clicks away, the score will get saved with that user being the grader, which is incorrect.

zyx-billy:
Aims to resolve #446.

A really simple enhancement:
Users who click on a score but don't edit anything are no longer saved as the grader. 
(As a side effect, there's now a local cache of the fetched feedback, so subsequent clicks on the same score no longer go through the server).

Possible next steps (not included in the upcoming deployment) would be to include the feature discussed in last week's meeting:
Edits are saved locally until user clicks save button (i.e. do not automatically save to server when user clicks away) to prevent any accidental edits being saved.

### DIFF
--- a/app/assets/javascripts/gradesheet.js.erb
+++ b/app/assets/javascripts/gradesheet.js.erb
@@ -364,7 +364,7 @@ jQuery(function() {
                 console.log("submitting changes to score");
                 return true;
               }
-              _close_current_editor();
+              close_current_editor();
               return false;
           },
           submitdata: function(value, settings) {
@@ -454,6 +454,9 @@ jQuery(function() {
 
         if (released_changed || old_feedback != curr_feedback) {
           console.log('submitting changes to feedback/released-state');
+
+          // reset released-state changed flag
+          jQuery('.score_details input.released', editor).data('changed', false);
 
           var released = jQuery('.score_details input.released', editor).is(':checked');
           jQuery.ajax("quickSetScoreDetails", {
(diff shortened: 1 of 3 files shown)


nayak16:
Do you think caching via cookies/sessions would be better than a javascript dictionary? Not saying it's wrong but just wanted to ask if you considered both options.

ymzong:
@nayak16 
I believe caching with sessions is usually done by storing the "drafts" on the server (e.g. creating new posts in Piazza), and this would involve creating a table for drafts in our database, which can complicate things really fast. @zyx-billy what do you think?

zyx-billy:
Since the items that are cached are mapped with two keys (submission-id and problem-id), and there are multiple things that's stored per item (namely score number and feedback text), I just thought it'd make storing and accessing easier using regular JS objects, because with cookies you have only one key and you can only store strings. 
Not sure about the "drafts" though, but sessions normally involve the server so it's probably not that useful.

zyx-billy:
I'll be at the office before 3 tomorrow, so I can show you two how it performs.

nayak16:
This seems like a temporary fix good enough for this release that we can move away from in the future if other pages involved caching?

zyx-billy:
Sure if other pages involve caching I think we should definitely find a more systematic way of caching, but... for now this is a pretty efficient fix IMO (the reason it exists is we needed to check for diff and potentially store local edits in the future, the caching part is really just a side effect)

nayak16:
I agree. Looks good!

PR #655, opened 2016-04-12
remove redundant decodeURI call when showing annotations

### LINKED ISSUE #653: Annotations Weirdness with '%'
As in the issue recently reported to us, when a TA's annotation includes the percentage sign, the annotation functions on student submission page are all unresponsive.

I'm guessing that it has to do with `annotations.js` and Javascript string escaping. Any Javascript wizard wanna help out?

zyx-billy:
Fixes #653 

decodeURI function call is not needed since original data is not encoded.


ymzong:
Thanks!!

PR #663, opened 2016-04-29
Make search experience consistent

### LINKED ISSUE #660: Search Inconsistencies
Currently, different pages have different searching mechanisms.
Below is a summary thanks to Prof. Kosbie:
- [x] - You are now unable to search in Grade Submissions by Andrew ID - only by first or last name. There is no indication that this is the case, leading many CAs to ask me why their students disappeared from Autolab.
- [x] - However, you still have to search by Andrew ID in Manage Submissions view. 
- [x] - Double however, once you go into Create Submission in Manage Submissions, you can only search by first and last name again. 
- [x] - And then in Manage Extensions, you don't search by Andrew ID by typing really fast, but you can only search by first and last name now.

Ideally all pages should be able to search by both email and name.

zyx-billy:
Aims to fix #660.

All search boxes can now search for _any part of a user's first name, last name, or email_.

Affected pages and example:
Grade Submissions (gradesheet)
gradesheet
Manage Submissions (now shows the user's name in the table like the gradesheet)
manage_extensions
Create Submission
create_submission
Manage Extensions
manage_extension

### DIFF
--- a/app/controllers/extensions_controller.rb
+++ b/app/controllers/extensions_controller.rb
@@ -11,7 +11,7 @@ class ExtensionsController < ApplicationController
     @extensions = @assessment.extensions.includes(:course_user_datum)
     @users = {}
     @course.course_user_data.each do |cud|
-      @users[cud.email] = cud.id
+      @users[cud.full_name_with_email] = cud.id
     end
     @new_extension = @assessment.extensions.new
   end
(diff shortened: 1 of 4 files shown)


zyx-billy:
Rebased onto new master. Tests now pass.

nayak16:
@zyx-billy Isn't this already part of `master` and `Autolab2.0`

zyx-billy:
@nayak16 I don't think this one was ever put into production, so it's not currently on master or Autolab2.0 (not even on develop).

cg2v:
it looks like item 1 (grade submissions) from #660 is fixed in production, but nothing else is.

PR #670, opened 2016-05-31
Trap and report scheduler exceptions

cg2v:
fork with a block does not seem to propagate exceptions in the subprocess
to rescue blocks outside the fork block, so exceptions inside scheduler
modules were not being logged. Trap them inside the fork block, also trap
ScriptError exceptions, and send the subprocess exceptions to
ExceptionNotifier.

This is tested, except for the metadata added to the ExceptionNotifier call


zyx-billy:
merged into master with #735

PR #673, opened 2016-06-07
Disconnect from database before running Moss

cg2v:
Since moss can run for an extended period of time, it is possible
for database connections to time out while it is running. ActiveRecord
does not check database connections for liveness whenever they are used,
only when they are checked out or back in to the connection pool.

Make activerecord return all active connections to the pool before running
moss, so any post-moss database access will use hopefully-live connections.


zyx-billy:
merged into master with #735

Issue #675, opened 2016-07-11
bug with importing existing assessment from tarball (and a fix)


msherr:
Hi,

We're setting up Autolab at Georgetown.  During this process, we couldn't get "Option 3: Import an existing assessment from tarball." to work from the InstallAssessments screen.  There appears to be a bug in https://github.com/autolab/Autolab/blob/master/app/controllers/assessments_controller.rb#L675.

Specifically, tarballs include a "/" at the end of directory names.  So it will never be the case that 

`if entry.directory? && pathname.count("/") == 0`

As a workaround, I added the line:

`pathname.chomp!("/")`

which removes trailing "/".  I admit that I've never actually programmed in Ruby, and this probably isn't the best fix.  In particular, it will cause problems for non-directories that have "/" as the last character in the filename (a pretty strange edgecase).

Adding in the chomp! statement seems to now allow Option 3 to work, which was important for our deployment here at Georgetown.

Again, I should re-emphasize that there's probably a much better way to fix this.  

Thanks,
Micah
CS Faculty at Georgetown

Here's the patch:

```
diff --git a/app/controllers/assessments_controller.rb b/app/controllers/assessments_controller.rb
index f298e39..d160418 100755
--- a/app/controllers/assessments_controller.rb
+++ b/app/controllers/assessments_controller.rb
@@ -670,6 +670,7 @@ private
     asmt_yml_exists = false
     tar_extract.each do |entry|
       pathname = entry.full_name
+      pathname.chomp!("/")
       next if pathname.start_with? "."
       # nested directories are okay
       if entry.directory? && pathname.count("/") == 0
```

pcbouman-eur:
In Rotterdam we also had issues with this exact problem. The problem doesn't occur when you export an assessment from Autolab via the webinterface (you get a tar that you can import), but as soon as you extract the tar and use the tar command to create a new archive, Autolab refuses the new file.
As a workaround, I took the "Assessment export" code from Autolab and put it in a ruby script that I could use instead of the regular tar command to create archives, and the results can be imported by Autolab. A fix for this issue would be very convenient for us as well.

zyx-billy:
Fixed by #683.

PR #683, opened 2016-09-02
fix trailing slash issue with importing assessment from tarball

zyx-billy:
For #675.

The logic with the original code seems fine. The only problem was it was assuming directories in a tarball don't end with a slash. We just need to remove any trailing slashes for directories so that it aligns with the assumption.

(This assumption guarantees that directories containing no slashes are root level directories. An alternative solution would be updating the export assessment code so it appends a slash to every directory (just like how the cmdline tool tar would), and then assuming directories that contain exactly one slash are root level directories, but that isn't robust either. So it seems the most flexible way is to just preprocess the directory name so we don't assume anything about trailing slashes)

### DIFF
--- a/app/controllers/assessments_controller.rb
+++ b/app/controllers/assessments_controller.rb
@@ -671,10 +671,11 @@ private
     tar_extract.each do |entry|
       pathname = entry.full_name
       next if pathname.start_with? "."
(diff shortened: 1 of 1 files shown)


ymzong:
Lgtm!

PR #697, opened 2016-10-02
Updated CMU Deployment merge into master

nayak16:
Merge the working commits the CMU deployment of Autolab into the master branch of Autolab


zyx-billy:
Note to self/team: revert commits involving CMU specific info after merge

cg2v:
why do you want to resolve the differences this way? merge/update the underlying prs, and I will rebase my tree to match

zyx-billy:
@cg2v So this isn't actually trying to resolve the differences between master and the deployed version. We are currently retiring the develop branch in favor of just using master. Instead of merging develop right into master, which would pull in untested changes from a long time ago that were never used, we discovered it would be safer and easier to just merge the current deployed version since it was based on master, and also included the essential changes made on develop. (of course, if we ever find out we need anything specific from develop, we could always pick them into master again).

nayak16:
@cg2v We also agreed that merging this branch would be better than cherry picking each PR

nayak16:
Merged as part of #699

PR #699, opened 2016-10-05
Updated and rebased CMU Deployment merge into master

cg2v:
Merge the working commits the CMU deployment of Autolab into the master branch of Autolab (use this instead of #697 

This excludes the local deployment patches, which would spare you from needing to revert them after.
A rebased tree that includes all the patches is cg2v/new-deploy-fall-2016 which has an identical tree to cg2v/deploy-fall-2016-mid


nayak16:
Same here @cg2v can you rebase with autolab:master and then push back to this PR.

PR #726, opened 2016-11-10
Adapt to new roster format

cg2v:
try to parse new roster format.

I converted the column mapping to something table driven.
It's not tested yet, so don't merge at this time.

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -27,7 +27,8 @@ class CoursesController < ApplicationController
     redirect_to course_assessments_url(@course)
   end
 
-  NEW_ROSTER_COLUMNS = 29
+  ROSTER_COLUMNS_S15 = 29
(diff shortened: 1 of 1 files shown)


app/controllers/courses_controller.rb
```diff
@@ -577,48 +578,58 @@ def detectAndConvertRoster(roster)
...
+    else
+      offset = 0
+    end
+    numRows = parsedRoster.length - offset
+    convertedRoster = Array.new(numRows) { Array.new(11) }
+
+    if (Rails.env == "production")
+       domain="andrew.cmu.edu"
+    else
+       domain="foo.bar"
```
nayak16: Can we change this to something like "develop.autolab.edu". Just to make it clearer. Been wanting to change this for a bit.

nayak16:
@cg2v Can you rebase with master so the build passes. I just fixed Travis. Once that's done, I can merge this

PR #731, opened 2016-11-27
Material Course List and Assessment List

zyx-billy:
New course and assessment index pages using Materialize:

Courses page:
material_course_page
Uses a different color (deep-orange for now) for courses where the user is a CA/Instructor so its easier to distinguish. The second card action also changes from viewing the user's gradebook to viewing the section's gradebook.

Assessments page:
material_assessment_list
Unreleased assessments will have the unreleased badge.

(We can definitely do a lot with these badges).

### DIFF
--- a/app/views/assessments/index.html.erb
+++ b/app/views/assessments/index.html.erb
@@ -17,7 +17,10 @@
   <h3 class="section-title"><span class="white">Assignments</span></h3>
 
   <div class="row">
-    <% @course.assessment_categories.each do |cat| %>
+    <% @course.assessment_categories.each_with_index do |cat, idx| %>
+      <% if idx % 3 == 0 %>
+        <div class="clearfix hide-on-small-only"></div>
+      <% end %>
       <% asmts = @course.assessments_with_category(cat, @cud.student?) %>
       <% if asmts.any? %>
         <div class="col s12 m4">
--- a/app/views/assessments/index.html.erb
+++ b/app/views/assessments/index.html.erb
@@ -6,46 +6,52 @@
 <% end %>
 
 <% if @is_instructor %>
-<h2>Instructor Actions</h2>
+<div class="section">
+  <h3 class="section-title"><span class="white">Instructor Actions</span></h4>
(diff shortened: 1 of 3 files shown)


nayak16:
@zyx-billy can you rebase with Autolab2.0 and push to this branch again to make sure build passes

zyx-billy:
@nayak16 All tests pass 👌

nayak16:
Awesome. How does this do on small window sizes? Can you just post a screenshot for the record of mobile size? @zyx-billy

app/views/assessments/index.html.erb
```diff
@@ -6,46 +6,55 @@
...
+            </div>
+
+            <div class="collection red darken-4" style="border-bottom: 0">
+              <% asmts.each do |asmt| %>
+                <%= link_to(course_assessment_url(@course, asmt),
+                            {:class => "collection-item grey-text text-darken-4"}) do %>
+                    <%= asmt.display_name %>
+                    
+                    <% if !asmt.released? %>
+                      <span class="new badge blue darken-4" data-badge-caption="unreleased"></span>
```
nayak16: Can this badge link to the page where you can release the assessment? So the person won't have to go the normal way, but go straight there. I think it's the edit handin page `/courses/AutoPopulated/assessments/homework0/edit/handin`

nayak16:
Also personally I don't think the deep orange is the best. Maybe experiment with a border? We'll look at it more at the meeting tomorrow.

zyx-billy:
screen-shot-2016-11-29-at-16 58 15
Right now they are set to flow into one column on small screens. 

Yeah, I just kind of picked a color... We should definitely consider more options.

app/views/assessments/index.html.erb
zyx-billy: Great idea! It wouldn't be too hard to add that.

nayak16:
screen shot 2016-12-03 at 6 49 14 pm

On mobile the badge overlaps the title. Maybe stacking them on small screens is better? This is prob a small change (making col s9 and s3).

PR #734, opened 2016-11-29
Add mysql gem

nayak16:
Build scripts and dependancies need MySQL. Adding back to ensure Travis builds.


nayak16:
@devanshk tagging so you see this

PR #735, opened 2016-11-30
CMU Production features and commits

nayak16:
Clone of PR #699. Rebased with master so Travis build passes


nayak16:
@zyx-billy Can you take a look quickly? Want a second 👁

zyx-billy:
looks pretty good, same as the previous one

PR #743, opened 2016-12-31
Add materialize datepicker and timepickers

nayak16:
Version 1.0 Suggestions welcome. This took a surprising amount of time (pun intended) to do

screen shot 2016-12-30 at 7 58 24 pm
screen shot 2016-12-30 at 7 58 12 pm

PR #746, opened 2017-01-03
Add datetimepicker and styles

nayak16:
Found this lightweight datetimepicker. Better validation and easier to use than the Materialize Date Picker with less custom code. No more bugs as #743 had


zyx-billy:
this one looks great!

PR #751, opened 2017-01-04
Attachments#show should use an absolute path

cg2v:
Otherwise if the working directory changes (e.g. due to moss),
it will always fail


nayak16:
Looks good. Curious if you encountered another case where the wd changes other than Moss?

Issue #546, opened 2015-08-04
Instructions for installing new assessment


droh:
The instructions on the page installing an assessment from a previous instance (option 2) will be very confusing to CMU professors because it gives an absolute local pathname (/opt/...) for the instructors to copy their files to. It would be better to remove the absolute pathname and simply refer to the course directory.

TheodorJ:
Merged #759

PR #752, opened 2017-01-04
Use Rails.root.join everywhere

cg2v:
Never use File.join(Rails.root...), only Rails.root.join(...)

also add some parenthesis around some path computations to
improve readability.

I assumed that this would be requested after reviewing #751. If you don't think it is necessary, close it.


nayak16:
Looks good 👍

nayak16:
This fixes #546 and covers #759

cg2v:
This does not fix the linked issues. It only changes how the path is computed, not what is displayed. If you want to display a relative path or resolve the link to the real absolute path, that must be done separately

Issue #710, opened 2016-10-12
Version Penalty does not account for deleted submissions


zyx-billy:
As reported, after an earlier submission was deleted by course staff, the total number of submissions is not updated when considering version penalty.

zyx-billy:
The code for giving penalty (in submission.rb) only compares the version number (which simply increments for each new submission) with the version penalty threshold. To account for deleted submissions, this entire logic would need to be fixed.

TheodorJ:
Merged #753

PR #753, opened 2017-01-05
Fix Issue #710: Version Penalty does not account for deleted submissions

TheodorJ:
Fix Issue #710: When calculating version penalty, the version_over_threshold_by method in submission.rb now only accounts for the number of submissions that exist rather than the version number of the current submission. The new version numbers of new submissions will still be dependent upon deleted submissions, but deleted submissions will not count toward the version penalty.

### DIFF
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -234,13 +234,18 @@ class Submission < ActiveRecord::Base
     final_score_opts o
   end
 
+  # NOTE: threshold  is no longer calculated using submission version,
+  # but now using the number of submissions. This way, deleted submissions will
+  # not be accounted for in the version penalty. 
   def version_over_threshold_by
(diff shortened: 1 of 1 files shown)


nayak16:
Looks good but seems like you included your previous commit in this branch too. This happens when you create a new branch from your old branch and not `master`. But it's okay in this case since it's a small PR, but just keep it in mind.

nayak16:
This still contains commits from another branch. Try rebasing this branch with an updated `master` branch and `git push -f origin tjjohans_edits` (force push)

TheodorJ:
There we go, I figured out how to remove all of the previous commits. Only "Fix Issue #710" is there now, so the branch should be ready to merge.

nayak16:
👍

PR #759, opened 2017-01-12
Fix Issue #546: Absolute Directory

TheodorJ:
Basically, I removed "Rails.root" from File.join(Rails.root, "courses", @course.name). The only change is that the root directory is no longer displayed.


nayak16:
#752 Solves this already and more (sorry @TheodorJ 🙃 )

nayak16:
Apologies, this is solving a different problem

nayak16:
This can be merged!

Issue #763, opened 2017-01-18
UsersController#create with existent email throws ruby error


TheodorJ:
When creating a new user that has an email that is already in use, Autolab crashes with a ruby error. This error should be caught and displayed in front of the user.

TheodorJ:
Merged #765

PR #765, opened 2017-01-21
Fix Issue #763: Duplicate email error not caught

TheodorJ:
I added a begin/rescue clause that flashes "Email #{@user.email} already in use" if the email already exists.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -99,6 +99,9 @@ class UsersController < ApplicationController
     save_worked = false
     begin
       save_worked = @user.save
+      if !save_worked
+        flash[:error] = "User creation failed"
+      end
     rescue => error
       error_message = error.message
       if error_message.include? "Duplicate entry" and error_message.include? "@"
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -102,17 +102,17 @@ class UsersController < ApplicationController
     rescue => error
       error_message = error.message
       if error_message.include? "Duplicate entry" and error_message.include? "@"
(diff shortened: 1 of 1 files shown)


app/controllers/users_controller.rb
```diff
@@ -96,13 +96,23 @@ def create
...
+        flash[:error] = error_message
+      end
+    end
+    if save_worked
       @user.send_reset_password_instructions
       flash[:success] = "User creation success"
       redirect_to(users_path) && return
     else
-      flash[:error] = "User creation failed"
+      #flash[:error] = "User creation failed"
```
nayak16: Why is this commented out? Also, let's change this to "Sorry, could not create user. Please try again later."

app/controllers/users_controller.rb
```diff
@@ -96,13 +96,23 @@ def create
...
     @user.skip_confirmation!
-
-    if @user.save
+    save_worked = false
+    begin
+      save_worked = @user.save
+    rescue => error
+      error_message = error.message
+      if error_message.include? "Duplicate entry" and error_message.include? "@"
+        flash[:error] = "Email #{@user.email} already in use"
```
nayak16: Make this error message "User with email #{@user.email} already exists"

app/controllers/users_controller.rb
```diff
@@ -96,13 +96,23 @@ def create
...
+      else
+        flash[:error] = error_message
```
nayak16: I don't think it's good to surface raw MySQL errors to the user. Instead here just set `save_worked = False` so that the "Could not create user" message will be shown.

nayak16:
Looks good to merge

PR #767, opened 2017-01-25
Entire blocks as links

### LINKED ISSUE #756: entire "blocks" in materialize should be links
Since there are no buttons or links on the new course list index page, AND the whole block has a shadow animation, the whole block, and not just the text, should be a link to the course.

zyx-billy:
Aims to fix #756.

This fixes the original issue with individual assessment links on the assessment index page. Instead of only the text being clickable, now the entire collection-item block is clickable and acts as a link. The sub-link for the badges still works (to allow for this, a JS onclick approach was used, since nested a tags are illegal).

### DIFF
--- a/app/views/assessments/index.html.erb
+++ b/app/views/assessments/index.html.erb
@@ -1,3 +1,17 @@
+<% content_for :javascripts do %>
+  <script>
+    jQuery(function() {
+      /* a way to mimic nested anchor tags */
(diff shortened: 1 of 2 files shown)


zyx-billy:
the second commit makes the entire red block on the course cards (on the course index page) clickable.

nayak16:
LGTM 👍

PR #768, opened 2017-01-25
Non exist course 404 error

### LINKED ISSUE #625: Course does not exist returns 500
If a course doesn't exist, wouldn't 404 be better as opposed to internal server error?

jbman223:
Fixes #625. Courses that don't exist will produce a 404 instead of a 500.


nayak16:
Looks good. I'll merge this in

Issue #757, opened 2017-01-10
validation error when saving course_user_datum does not explain the problem


cg2v:
if CourseUserDatum#create fails due to a validation, it just says "The user with email #{email} could not be created". It does not display the results of validation (as http://guides.rubyonrails.org/active_record_validations.html#displaying-validation-errors-in-views would)

TheodorJ:
How did/do you deliberately cause validation errors when creating new users? I need to test out a few solutions.

nayak16:
Try an email that already exists @TheodorJ

TheodorJ:
@nayak16 I tried that and it threw a MySQL "Duplicate entry 'user0_autopopulated@foo.bar' for key 'index_users_on_email'" error. This happened when user.save was called in user_controller.rb, so I'll try moving the validation error clause into the same function so that it will check for validation errors before SQL errors like this one happen.

cg2v:
The place I (and some faculty) were seeing it was missing first name and/or last name

TheodorJ:
@cg2v That's odd. I tried creating users with missing first and last name fields, and instead of throwing a validation error, it simply flashes "User creation failed" and lists which field is missing. If this is the only case, I could add a note that says which fields are required, which would likely solve the issue.

cg2v:
What I get is "The user with email something@andrew.cmu.edu could not be created"
That is, unless the username is already in autolab due to having logged in or being in another class, in which case the name is autofilled from the existing user record.

nayak16:
Merged 3d32d55

PR #769, opened 2017-01-28
Fix Issue #757: CUD creation now explains problem

TheodorJ:
Basically, we know that when CUD creation fails, it simply displays the message "The user with email #{email} could not be created". As it turns out, there is a code snippet in app/views/course_user_data/new.html.erb that displays all errors that the object contains, but because @newCUD.user is equal to nil when the user is created, no errors are actually stored anywhere (at least, as far as I know). So instead of printing the errors, I added a code snippet in course_user_data_controller.rb that flashes "All required fields must be filled" if the email, first, or last name fields are empty (which caused the error in the first place) if any of these fields are empty. The generic error message "The user with email #{email} could not be created" is still included just in case there are any errors that are not caused by empty fields.

In addition to these changes, I also added some html code that indicates required fields.

### DIFF
--- a/app/assets/stylesheets/style.css.scss
+++ b/app/assets/stylesheets/style.css.scss
@@ -386,6 +386,11 @@ ul.gray-box.small > li + li {
   font-size:10px;
 }
 
+/*This is the one that we're (I'm) using for required-field indicators.*/
+.redText {
+  color: red;
+}
+
 /**
 * Displayed on the home page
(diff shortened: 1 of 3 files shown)


nayak16:
Minor fix and then good to merge

TheodorJ:
Alright, I removed all of the comments I made. Should be good to go

Issue #729, opened 2016-11-22
Display times in user's timezone


dnivra:
Currently, the times being displayed(eg: deadline for an assessment) are based on the time zone configured in `config/application.rb`. It seems only UTC times are stored in the database. Perhaps, users can configure their preferred time zone in their profile and the settings in `config/application.rb` could be used as fallback in case users don't set their preferences.

zyx-billy:
Instead of making it a user setting, all times will be displayed in the user's browser time zone. See PR #775.

PR #775, opened 2017-02-01
Display browser local time everywhere

zyx-billy:
Currently times displayed on Autolab are in the time zone of the server clock (and sometimes in the time zone specified in the config file, and sometimes even in UTC). #729 brought into attention that we should instead use browser local time to prevent confusion or the need for manual conversion.

New standardized procedure:
All instances of displayed time are wrapped in a 'moment-date-time' class, which will be processed once on the client side into the browser local time (a feature of moment.js). The raw time in the html will still be in the time zone specified in the rails config file, so even if js is blocked or fails, the time will still be understandable.

A list of all instances of time updated:
- Submit file page (assessments:show)
- Assessment submission history page (assessments:history)
- Manage submissions page (submissions:index)
- Recent jobs page (jobs:index)
- Job details page (jobs:getjob)
- Scoreboard page (scoreboard:index)

If there's any other page with a displayed time that's missing please comment.

### DIFF
--- a/app/assets/javascripts/app-level.js
+++ b/app/assets/javascripts/app-level.js
@@ -5,9 +5,9 @@
 $(document).ready(function() {
 	$(".moment-date-time").each(function(ind) {
 		$el = $(this);
-		var format = $el.data("format")  || "MMMM Do YYYY, HH:mm:ss";
+		var format = $el.data("format")  || "YYYY-MM-DD HH:mm:ss";
 		var unformattedDate = $el.html();
-		var formattedDate = moment(unformattedDate, "YYYY-MM-DD hh:mm ZZ").format(format);
+		var formattedDate = moment(unformattedDate, "YYYY-MM-DD hh:mm:ss ZZ").format(format);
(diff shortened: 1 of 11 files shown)


app/assets/javascripts/app-level.js
```diff
@@ -5,9 +5,9 @@
 $(document).ready(function() {
 	$(".moment-date-time").each(function(ind) {
 		$el = $(this);
-		var format = $el.data("format")  || "MMMM Do YYYY, h:mm a";
+		var format = $el.data("format")  || "YYYY-MM-DD HH:mm:ss";
```
nayak16: Why was seconds added back in?

zyx-billy: That's the default format for those that do not specify a format, which means timestamps like assessment submission time, job completion time, etc., and I thought it would be useful to have seconds in there to be precise (and I also changed it into a standard format that's easy to sort and manipulate I guess). For human friendly times like assessment due times, it won't display seconds.

PR #776, opened 2017-02-08
UB Blue

MikeDrewitt:
Other than all of the blue stuff. We fixed a ton of CSS bugs. We also built a feature that allows for assignment creators to upload an HTML form that can be used in place of uploading a submission file and rather fill in a form for an assignment. Moss has been made 1000x easier for professors to use who don't actually know how to use moss. We also fixed a bug that effects admins while not in development mode that caused a 500 error. 

Theres a ton more that's been merged in but it's a lot to list.


zyx-billy:
Thanks for the PR! Merged into Autolab 2.0 with #782.

PR #778, opened 2017-02-11
UB changes with Autolab red theme

nayak16:
Copy of #776 with Autolab red theme instead


nayak16:
Closing in favor of #781

PR #780, opened 2017-02-22
upgrade to latest datatables to fix error with gradesheet

zyx-billy:
Apparently the gradesheet page no longer works now because of a problem with our old version of JQuery Datatables library. Upgrading it to the latest version restores functionality.


zyx-billy:
Apparently already fixed by UB changes (#776 / #778)

PR #786, opened 2017-02-26
Deprecate SVN

### LINKED ISSUE #785: Deprecate SVN
We need to remove all mentions of SVN from the Front-End. 

Backend code will be removed later.

nayak16:
Fixes #785

### DIFF
--- a/app/views/assessments/_edit_basic.html.erb
+++ b/app/views/assessments/_edit_basic.html.erb
@@ -87,12 +87,6 @@
       <% end %>
     </span>
   </li>
-  <li class="collection-item">
-    SVN
-    <span class="secondary-content">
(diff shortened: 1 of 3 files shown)


nayak16:
@zyx-billy Do you know of any other places SVN is mentioned?

zyx-billy:
I think there's an option to view SVN settings in the assessment#show view.

zyx-billy:
Just did a find, there's no more SVN links. But if they had the link they could still access it. Do you think we should just remove the routes to the SVN settings page as well?

nayak16:
Let's leave them in for now. We would get less blowback if SVN routes still work but they can no longer edit settings than if we removed them all together. Plus, it's not too urgent

zyx-billy:
👌

PR #787, opened 2017-02-26
Remove 'Delete User' Button on Edit CourseUserData page and change SignUp page HTML

TheodorJ:
Destroying a student removes all data permanently, which is not good. Instead, instructors should drop students so that their data persists and they can be re-added to the course if necessary.


TheodorJ:
screen shot 2017-02-26 at 4 36 07 pm

nayak16:
Looks god

PR #788, opened 2017-02-26
Fix destroy course id reference

zyx-billy:
Destroy course tried to reference the "id" param, which doesn't exist at all. Replaced it with the @course instance variable set by "set_course".

For some reason, set_course is skipped if the authorization level is set to administrator, which doesn't make sense, but also didn't cause any problems since course#destroy is the only method with admin authorization that needs the course set. This update removes the skip, and adds it to the other admin methods that need skipping.


nayak16:
We find more and more strange code every day....but this fix looks good

PR #792, opened 2017-02-28
Fix Gradesheet Numeric Column Sorting

### LINKED ISSUE #791: gradesheet sorting is lexical, not numeric
This was reported by a CMU faculty in F16, and worked around in the computing services deployment:

When I click on Grade Submissions, then click on a field (say, "Autoscore") to sort by that field, the sorting is not numeric, but seemingly alphabetic.
and
...TA noticed that this works properly if we click on "Total"

My workaround is clearly non-optimal: I added
if (type == "sort" && score != "" && score != "-") return Number(score);
right above the return in the render function in _gradesheet.html. I assume it would be better to store the values as numbers from the beginning, but I don't have enough javascript experience to know if there would be bad side effects from that.

zyx-billy:
fixes #791.

There was already a set of custom sorting functions defined for the 'numeric columns', however the actual numeric columns were not registered with the datatables plugin because the count was wrong. Fixing the num_cols count solves the problem.

e.g.
gradesheet_numeric_columns_sor

### DIFF
--- a/app/assets/javascripts/gradesheet.js.erb
+++ b/app/assets/javascripts/gradesheet.js.erb
@@ -7,7 +7,7 @@ jQuery(function() {
     var current_editor = undefined;
     var current_popover = undefined;
 
-    var num_cols = jQuery("table#grades > thead > tr:first > th").length
+    var num_cols = hclasses.length;
 
     numeric_columns = []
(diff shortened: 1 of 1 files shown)


nayak16:
Thanks for the quick fix. Looks good 👍

PR #800, opened 2017-03-01
Fix checkbox grade release

jbman223:
Fixes invisible checkbox when choosing whether or not to release a grade. This issues was caused by how materiel styles check boxes. A check box MUST have a label associated in order to appear (in materialize), however this checkbox had no label and was invisible.


jbman223:
Accidentally based off of the wrong branch, disregard this for #801.

PR #805, opened 2017-03-05
Fix Manage Submissions Page functionality

zyx-billy:
Adds labels as required by materialize so the checkboxes in the table shows up.

### DIFF
--- a/app/views/submissions/index.html.erb
+++ b/app/views/submissions/index.html.erb
@@ -2,6 +2,13 @@
 
 <% content_for :stylesheets do %>
   <%= stylesheet_link_tag "datatable.adapter" %>
+
+  <style>
+    table [type="checkbox"] + label {
+      width: 18px;
+      padding: 0;
+    }
+  </style>
 <% end %>
 
 <% content_for :javascripts do %>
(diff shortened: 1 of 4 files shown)


nayak16:
Review pls @zyx-billy

app/views/submissions/index.html.erb
```diff
@@ -2,58 +2,74 @@
...
+      z-index: 5;
+    }
+  </style>
 <% end %>
 
 <% content_for :javascripts do %>
   <%= external_javascript_include_tag "lodash", "3.3.1" %>
   <%= javascript_include_tag "sorttable" %>
-  <%= external_javascript_include_tag "jquery.dataTables", "1.10.5" %>
+  <%= external_javascript_include_tag "jquery.dataTables", "1.10.9" %>
```
nayak16: Main problem

zyx-billy:
functionality looks good. I removed the select-all checkbox in the table head for now since it's not doing anything (people still have the 'regrade all' button after all). If we want to implement that we'll need to have it loop through and select every single submission, which might be hard since the table is paginated.

PR #810, opened 2017-03-20
Use entities instead of inline utf8 in view

cg2v:
The footer includes a raw middle dot · in the HTML. Use the
&amp;middot; entity instead


cg2v:
I was looking at this because I've gotten some exceptions of the form "incompatible character encodings: ASCII-8BIT and UTF-8 app/views/layouts/application.html.erb:68:"
application.html.erb:68 is where the footer is loaded. It's possible the unencoded content would still trigger problems later.

You might also want to make this an actual list and style it using something like

.footer li {
display: inline;
list-style-type: none;
}

.footer li + li:before {
    content: " &middot; ";
}

This is all untested.

PR #813, opened 2017-03-23
set z-index for annotation forms

### LINKED ISSUE #812: Grading window appears in layer below document
When grading a document that is multiple pages long, the grading window appears in the layer above the page clicked on (as intended) but in the layer below the following page. This hides the Save and Cancel buttons, making it impossible to save the grade.

zyx-billy:
Fixes #812.

The z-index of annotation forms were not set, causing them to be behind the next page.
(This was not introduced by 2.0 though. Tested with an earlier version and the same problem occurred.)

pdf_annotations_box_zindex


nayak16:
Looks good 👍

Issue #816, opened 2017-03-27
Moss result still displays "Failure" when success


nayak16:
As reported by David Kosbie, Moss result still shows a failed message even though the link to the results is displayed

nayak16:
Fixed with #817

PR #817, opened 2017-03-27
Move moss status fetching to after moss command

zyx-billy:
fix for #816.

The command that gets the exit status used to be before the command that actually runs moss, which meant it was getting the exit status of the previous child.


nayak16:
Looks good

PR #833, opened 2017-03-30
Gradesheet enable sort by tweak

### LINKED ISSUE #827: Cannot sort by tweak on gradesheet
On the Gradesheet page, the table cannot be sorted by the "tweak" column.

zyx-billy:
Fixes #827.

The gradesheet sorting function for numeric-html columns was not working for the tweak column. Updated it to be simpler and work better with NaN. Now it is able to sort the tweak column.


nayak16:
Awesome this works great 👍

PR #835, opened 2017-03-31
Refactor handin and autograde helpers

zyx-billy:
As part of the API, I had to refactor functions used for handin validation and autograding processing and error handling. Previous versions of these functions assumed the user would be viewing it on the web, and displayed error messages directly as "flash" messages.

This branch separates these functions from the assessment controller and into their own helper modules. For handin validation, the function returns one of a hash that represents either "success" or some problem. The caller then cases on the return value.
For autograding submission, functions no longer return all sorts of negative numbers. Instead, AutogradeErrors are raised and expected to be caught by the initial caller of sendJob. Each error contains a short message, a hash representing the category of error (in case the caller wants to case on it and display alternate error messages), and an additional message for generating verbose error messages.

No other program logic was intended to be changed. Accessing these functions through the normal web interface should work exactly the same as before. If anything seems odd please report. For now it passes the hello.c test, but that's just a success case, may need to test more kinds of failure cases.


app/controllers/assessment/autograde.rb
```diff
@@ -71,12 +74,16 @@ def regradeBatch
...
-          failed_jobs += 1
-          failed_list += "#{@submission.filename}: autograding error.<br>"
+        begin
+          autogradeSubmissions(@course, @assessment, [submission])
+        rescue AssessmentAutogradeHelper::AutogradeError => e
+          if e.error_code == :missing_autograding_props
+            # no autograding properties for this assessment
+            redirect_to([@course, @assessment, :submissions]) && return
+          else # autograding failed
+            failed_jobs += 1
```
nayak16: Do we even need `failed_jobs`? We can just check the length of `failed_list`

app/controllers/assessment/autograde.rb
```diff
@@ -71,12 +74,16 @@ def regradeBatch
...
+            failed_list += "#{@submission.filename}: autograding error.<br>"
```
nayak16: Will this function only be called when rendering HTML? If not, we want to get rid of the `<br>` in case this function is called via API

app/controllers/assessment/autograde.rb
zyx-billy: Good point! Not sure why it was there... I'll get rid of it.

app/controllers/assessment/autograde.rb
zyx-billy: Right now this is only called in the web. I'd imagine if we add this functionality to the API we wouldn't need to be this verbose with error messages. What we could do is write a function that loops through a list of submissions, regrade them, and return a list of failed ones. This way both the API and the HTML controllers would just need to call this helper and format responses.

nayak16: Okay makes sense. But ideally we would want both API and web code to call the`AssessmentAutograde` module functions. Then in their individual caller code, do whatever error handling is specific to that platform. This + the helper introduces 3 layers of indirection. Does that make sense? We can discuss more tomorrow

zyx-billy: Haha yeah let's discuss this in person tomorrow. 
(Hashtag: [FundamentalTheoremOfSoftwareEngineering](https://en.wikipedia.org/wiki/Fundamental_theorem_of_software_engineering))

zyx-billy:
@nayak16 I think it'd be a good idea to pull #709 and #711, or integrate them into this one. Otherwise this would totally break the two prs, and they seem rather useful.

nayak16:
Agreed would be useful to have

zyx-billy:
merged into api_v1

PR #837, opened 2017-04-05
Remove waves classes

jbman223:
Fixes issues where some buttons require two clicks. Removes wave effect from materialize framework.


nayak16:
Looks good

PR #838, opened 2017-04-06
set_course should return after rendering 404

cg2v:
After set_course renders the 404 page inline, it needs to immediately return
since any future exception will try to redirect or render triggering a double
render error


cg2v:
it might be nice to have context sensitive error pages "This course is not registered" or some such.

PR #840, opened 2017-04-11
Add script to change times on input

### LINKED ISSUE #826: Autofill last-date-to-submit and grading-deadline when setting due date
The last-date-to-submit and grading-deadline fields no longer automatically update when a due date is set. It used to be that these would update to be the same as the due date unless they had already been set to be later than that, which was very convenient.

nayak16:
Fixes #826

### DIFF
--- a/app/assets/javascripts/init_handin_datetimepickers.js
+++ b/app/assets/javascripts/init_handin_datetimepickers.js
@@ -37,5 +37,3 @@ $(document).ready(function() {
 
 })
 
-function validateDates(e) {
-}
--- a/app/assets/javascripts/init_handin_datetimepickers.js
+++ b/app/assets/javascripts/init_handin_datetimepickers.js
@@ -29,8 +29,6 @@ $(document).ready(function() {
         end_at_pickr.setDate(cur_date, true);
       }
 
-      // Reinstall handler
-      flatpickr_inst.set('onClose', [onCloseHandler])
   }
   /* Add custom onClose handler for due at date picker */
   var due_at_pickr = createDatePicker('#assessment_due_at', {onClose : onCloseHandler})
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/init_handin_datetimepickers.js
```diff
@@ -0,0 +1,39 @@
...
+      if (grading_deadline_pickr.selectedDates[0].getTime() < cur_date.getTime()) {
+        grading_deadline_pickr.setDate(cur_date, true);
+      }
+
+      if (end_at_pickr.selectedDates[0].getTime() < cur_date.getTime()) {
+        end_at_pickr.setDate(cur_date, true);
+      }
+
+      // Reinstall handler
+      flatpickr_inst.set('onClose', [onCloseHandler])
```
zyx-billy: Everything looks good to me. Though I don't think we need to reinstall handler each time? I tried without this line and it still worked multiple times.

zyx-billy:
Things worked perfectly. Do you think we should add a similar handler to watch the opposite direction? i.e. when end_at decreases to before due_at, we update due_at. And similarly when start_at increases too much we update everything else.

app/assets/javascripts/init_handin_datetimepickers.js
nayak16: What browser did you try it on? I tried on Chrome without this and it wouldn't work multiple times

nayak16:
That would be good but I'm trying to figure out a more generalizable approach rather than hard coding all the restrictions inside the handler.

app/assets/javascripts/init_handin_datetimepickers.js
zyx-billy: I tried it on Firefox yesterday and I just tried on Chrome, both worked for me. Perhaps try adding a log statement in the handler and see if the handler was called?

PR #841, opened 2017-04-11
Remove max score check from gradesheet

### LINKED ISSUE #836: Leaving Max Score for a Problem blank breaks entering scores in gradesheet
Leaving max score blank no longer allows inputting a grade that's higher than the problem's max score, because max score defaults to 0. 

This check should be bypassed when entering scores in the gradesheet

nayak16:
Fixes #836

### DIFF
--- a/app/controllers/assessment/grading.rb
+++ b/app/controllers/assessment/grading.rb
@@ -232,14 +232,6 @@ public
     sub_id = params[:submission_id].to_i
     prob_id = params[:problem_id].to_i
 
-    if params[:score].present?
-      maxScore = Problem.find(prob_id).max_score.to_f
-      if params[:score].to_f > maxScore
-       render :status => 422, :text => "bad data"
-       return
-     end 
-   end
-
     # find existing score for this problem, if there's one
     # otherwise, create it
     score = Score.find_or_initialize_by_submission_id_and_problem_id(sub_id, prob_id)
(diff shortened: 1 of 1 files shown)


zyx-billy:
Looks good to me! Basically reverted to what it was before 2.0 I guess.
One weird thing though: We didn't used to set score during `quickSetScoreDetails` ([this line](https://github.com/autolab/Autolab/blob/master/app/controllers/assessment/grading.rb#L295) was added in 2.0). There's a [comment](https://github.com/autolab/Autolab/blob/master/app/assets/javascripts/gradesheet.js.erb#L470) mentioning this in the gradesheet js. Do you know why it was set this way? It seems to work perfectly fine to me.

nayak16:
Hmm that is strange. I left the line in `quickSetScoreDetails` and it was still giving `422` error. Somewhere `quickSetScoreDetails` is still being called.

zyx-billy:
oh what I meant was before 2.0 we weren't using the score parameter in quickSetScoreDetails. In 2.0, we started using the score parameter, and also added in the max score check. It was right to remove that check in quickSetScoreDetails, but I'm not sure if we should also keep [the line](https://github.com/autolab/Autolab/blob/master/app/controllers/assessment/grading.rb#L295) that uses the score parameter.

nayak16:
Yeah that line is just superfluous actually

PR #849, opened 2017-04-12
Update PDF annotation button style

### LINKED ISSUE #814: No text on buttons on viewPDF page
The text on the buttons overflowed to the next line.
(There might seem to be space, but material buttons come with a generous side padding, which applies to all our buttons actually. It only caused a problem here because we're limiting the width of the button here)

pdf_annotation_sidebar_button_no_text

jbman223:
Fixes #814.


nayak16:
Looks good

Issue #821, opened 2017-03-29
Submission Searches by first name but sorts by Andrew ID


TheodorJ:
As per kkleiven's email:

"Searching for a student when creating a new submission does not work very well. You can only type in the student's first name, but the list is sorted by andrewid. So if I am searching for a student named 'Chris' whose andrewid starts with an s, but there is another Chris whose andrewid starts with a c, the search tool will bring me to the c's and I have to just scroll through the entire roster to find the student I was looking for."

PR #853, opened 2017-04-15
Fix student search bar

TheodorJ:
As per issue #821, the student search bar on the "Create Submission" page searches by first name, but sorts the list by andrewid. This fix replaces that choosing method with a search bar that searches students by either andrewid or first name.


nayak16:
Looks good

PR #867, opened 2017-04-27
Freeze flatpickr version

nayak16:
Emergency fix to fix the following UI issue when date input clicked. A new flatpickr update broke our integration


nayak16:
screen shot 2017-04-26 at 11 46 37 pm

PR #871, opened 2017-05-03
Due date fix

### LINKED ISSUE #859: Problems with due dates
There are some issues with the dates on the Edit Assessment Handin page:
- No check for setting "Due at" before "Start at"
- No check for setting "End at" before "Due at"
- No check for setting "Grading deadline" before "Due at"

- Any change to a date should preserve the following invariant:
"Start at" <= "Due at" <= "End at" <= "Grading deadline"

jbman223:
Fixes #859.

### DIFF
--- a/app/controllers/assessments_controller.rb
+++ b/app/controllers/assessments_controller.rb
@@ -541,7 +541,7 @@ class AssessmentsController < ApplicationController
 
       redirect_to(tab_index) && return
     rescue ActiveRecord::RecordInvalid => invalid
-      flash[:error] = invalid.message
+      flash[:error] = invalid.message.sub! "Validation failed: ", ""
 
       redirect_to(tab_index) && return
     end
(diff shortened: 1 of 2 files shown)


jbman223:
screen shot 2017-05-03 at 6 34 25 pm

devanshk:
Mostly fixes #859, but two issues:
1. There's no validation for Grading Deadline
2. The error messages are really verbose and robotic

jbman223:
@devanshk Updated language, and added missing validation.

devanshk:
Users enter in dates from now into the future so it makes more sense to think of the dates as 'after' rather than 'before', right?

Let's flip around the check and assume the date that's before is always right. Instead of `End at must be before the grading deadline` we'd have something like `Grading deadline must be after end date`

jbman223:
@devanshk Good call. I switched the order of the messages and it makes a lot more sense now.

PR #877, opened 2017-05-19
Hotfix: Email field is no longer disabled when adding user

devanshk:
This is a hotfix that fixes a bug that stops someone from adding users to a course


cg2v:
It doesn't really make sense to have these fields be modifiable when editing an existing CUD
The new.html javascript should enable them, or you should pass a parameter to control whether they are disabled.

PR #878, opened 2017-06-13
Log and notify when handin exceptions occur

cg2v:
When exceptions occur processing handins or autograde output, send the
exception to ExceptionNotifier and the course log. Otherwise, the only
notification in some cases goes to the student, who gets a bland
"submission failed" flash that they probably ignore.


cg2v:
This probably over-notifies. In general, I think the system needs some way of sending notification to course staff about problems in instructor modules or with parsing grader output.

PR #879, opened 2017-06-27
Regression: Added the 'Delete User' button back in

### LINKED ISSUE #876: Regression: No way to delete a student from a course
Regression: We no longer have a way to delete a user from a course roster (note that this is different from dropping).

TheodorJ:
Fixes #876


devanshk:
Clicking the 'Delete User' button sends me to a 404 page. Any thoughts on why this is happening?
screen shot 2017-06-26 at 9 09 11 pm

Click Delete User and get sent to ->

screen shot 2017-06-26 at 9 08 36 pm

TheodorJ:
Oops, that's because I forgot to add confirmDelete.html back in. I'll fix that as soon as I can

TheodorJ:
There. The 404 error shouldn't occur anymore

devanshk:
Love the delete confirmation screen, but a few more things to fix.

1. Once we delete the user, we should redirect back to the course page, since that's where we came from.
2. We should throw up a flash notice letting the admin know that the user was deleted. This would look something like:

`flash[:success] = "<name> account successfully deleted"`

oliverli:
Closed due to staleness

Issue #881, opened 2017-07-20
Error when adding multiple annotations to one submission for the same problem


devanshk:
Creating one annotation works fine but trying to create another or editing an existing annotation fails with "Failed to save annotation!!!".

To replicate:
1. Go to Lab -> Grade Submissions -> List archive -> Click on source file
(*.c).
2. Click on the '+' icon next to a line.
3. Add a comment, non-zero grade and select 'problem0' from the drop-down menu.
4. Save.
5. Repeat steps 3 and 4.

StevenJamesMoore:
It appears this may have been hot-fixed now, but we had another report (INC0285764 in ServiceNow) of a user having this same problem.

Has the hotfix been pushed to dev? I was able to replicate this on https://autolab-dev.andrew.cmu.edu just now.

AmirAlavi:
I'm also having this issue when grading assignments. I need to annotate two issues with the same problem, but can't add the second one.

devanshk:
The hotfix for multiple annotations is ready and we're finishing some testing before we pull to production.

devanshk:
Resolved

PR #882, opened 2017-07-20
Hotfix to re-enable multiple annotations for the same problem and submission

devanshk:
Addresses issue #881.

Currently includes a regression that decouples annotations and the gradebook - still determining whether annotations should change gradebook scores.


oliverli:
Closed due to staleness

PR #883, opened 2017-07-25
Bind new submission CUD to correct field

cg2v:
Bind the CUD id to the correct form element in Submissions#new, so it is
sent to the server on submit. Otherwise no CUD would be sent and no
submission could be created.


damianhxy (approved):
LGTM - Create new submission now works

PR #895, opened 2017-09-08
Allow empty score dictionaries.

ryanakca:
The semantics of autoresults seem to be: for every problem key in the
scores dictionary, update the recorded problem score with the new value.
Consequently, problems that don't appear in the dictionary don't get
their recorded scores updated.

To be fully consistent with this semantics, it *should* be possible to
have an empty score dictionary, i.e., one of the form
    {"scores": {}},
that updates none of the recorded scores. This is useful, e.g., when
your autograder needs to abort due to an error and doesn't want to
update anything.

### DIFF
--- a/app/controllers/assessment/autograde.rb
+++ b/app/controllers/assessment/autograde.rb
@@ -467,7 +467,6 @@ module AssessmentAutograde
       else
         scores = parseAutoresult(autoresult, true)
       end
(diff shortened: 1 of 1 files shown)


damianhxy:
It seems that there is no need to explicitly make this change -- given an empty score dictionary, the error will be rescued and none of the existing scores will change, due to the line `next unless score.new_record?`.

PR #903, opened 2017-09-19
Hotfix annotations

devanshk:
Hotfix that fully addresses issue #881 
(Creating a new branch here and PR instead of reverting #882 to preserve the git history.)

Decouples the annotations from the gradebook scores to allow for multiple annotations. Annotations no longer set the score for any given problem, but it's more important to allow a TA to add multiple annotations. Re-linking annotations and scores is being developed in the branch from #882.


devanshk:
Closed as we work towards Annotations 2.0 instead of implementing an interim hotfix.

PR #910, opened 2017-10-12
Fix issue 667

### LINKED ISSUE #667: handin requires javascript but it's not obvious
The autolab UI generally appears to work even if javascript is disabled or blocked by noscript, and this can create confusion if someone tries to use a feature (handin, annotations, gradebook) that depends on javascript while it's disabled or blocked.

I think it would be appropriate to either
- allow handins to work without javascript by not hiding the file picker button in the base html, but have the same script that registers the event handlers hide it, or
- include some <noscript> html near the submit button explaining that submissions don't work without javascipt.

jbman223:
Fixes #667 by adding an error flash if the user has javascript disabled.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -6,7 +6,7 @@
     <%= favicon_link_tag 'favicon.ico?v=1' %>
 
     <%= stylesheet_link_tag 'application' %>
-    <!-- Flatpickr Datepicker CSS (lightweight file) -->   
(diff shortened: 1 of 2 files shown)


devanshk:
Confirmed working on the home screen, but why is this only on the home screen? If a user follows a link to somewhere else on Autolab, shouldn't they see the alert there too?

jbman223:
Added the snippet in the proper view so that it is now visible on all pages.

PR #929, opened 2017-10-28
Create assessment config files during autopopulation

### LINKED ISSUE #639: Auto-populate script does not create assessment config files
A while ago I remember some of us were confused by the error: `no implicit conversion of Pathname into String` after following the Autolab setup script. Now I have found the reason why -- the auto-populate script does not generate assessment config files, and thus app/controllers/assessment/autograde.rb:515` fails when trying to load the non-existent file.

We should modify the auto-populate script such that it generates those files properly.

zyx-billy:
Fixes #639.

The autopopulate script does not generate assessment config files, which causes the annoying `no implicit conversion of Pathname into String` error when accessing assessments after setup. This fix creates the required config files during autopopulation.

### DIFF
--- a/lib/tasks/autolab.rake
+++ b/lib/tasks/autolab.rake
@@ -60,14 +60,17 @@ namespace :autolab do
           a.handin_filename = "handin.c"
           a.course_id = course.id
 
-          assessment_dir = File.join(course_dir, a.name)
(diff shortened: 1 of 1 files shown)


zyx-billy:
Great. Going to rebase this onto the current develop and then merge.

Issue #880, opened 2017-06-27
Wonky form inputs on extensions page


devanshk:
On the extensions page for an assessment, the input text fields exhibit wonky visual behavior and aren't uniform with the rest of the input fields. To recreate, go to 

http://localhost:3000/courses/AutoPopulated/assessments/lab5/extensions

and click 'Start typing student name or email'. The inputs should instead look like the input fields at 

http://localhost:3000/courses/AutoPopulated/assessments/lab5/edit

They should also have hints that are more symmetrical with those on the edit page.

<img width="1440" alt="screen shot 2017-06-26 at 9 31 51 pm" src="https://user-images.githubusercontent.com/1356687/27567049-550704b8-5ab7-11e7-8706-6248d942df7c.png">

PR #937, opened 2017-11-10
Form fixes jacob

jbman223:
Fixes issue #880. This is mostly stylistic changes to form fields to make them look nicer & be well aligned on pages.


jbman223:
I have fixed the issues with the embedded quiz form, as well as removed the -15 margin that is seen on autocomplete forms.

PR #945, opened 2017-12-03
Merging Updates to Annotations

TheodorJ:


jbman223:
Issues with PDF annotations:

- After I add a comment, when I make an edit the edit saves but does not show up until I refresh the page.
- Sometimes, I cannot see the save and cancel buttons on edit, I have to expand the interface first.
- The minimize button is also a trash can so I can't tell which is actually the minimize and which is the trash.

Code Submission:

- Must refresh page to see edited annotations.
- The comment when just posted (appears as a black line) is kind of hard to notice/read


Other than that, looks good.
Tested in Chrome.

PR #946, opened 2017-12-05
Autolab v2.1.0

devanshk:
Preparing this PR with a number of bug fixes from the Develop branch


bin/setup.sh
```diff
@@ -187,12 +187,12 @@ autolab_setup() {
     cp $AUTOLAB_PATH/config/database.yml.template $AUTOLAB_PATH/config/database.yml
     sed -i "s/<username>/$USER/g" $AUTOLAB_PATH/config/database.yml
 
+    cp $AUTOLAB_PATH/config/school.yml.template $AUTOLAB_PATH/config/school.yml
```
adriene-orange: Looks like this was a regression. Will need to fix before it is merged.

PR #947, opened 2017-12-05
decoupled annotations and gradebook again

devanshk:
^^


TheodorJ:
Looks good. I'll mark it as reviewed

cg2v:
Wouldn't it make sense to disable the score entry box? (not remove it, since there's data there for old assignments, but set the control to disabled so people don't try to put stuff in there.)

devanshk:
That info's still useful so we can display total point changes for each problem. Even if the scores don't directly affect the gradebook, we need them to generate the summary.

Issue #953, opened 2018-01-18
Assessment Creation Date Fields Always Use EST, but Display as Local After Creation


StevenJamesMoore:
**Issue**:
When creating an assessment, the times picked for due/open/etc. are in EST. Once it's been created, these times are then shown as local. For instance, when someone creates an assessment in Doha, they select for it to be due at 9 am (since it uses EST for creation) which translates to 5 pm Doha time.

**Ticket**:
INC0327646

devanshk:
Addressed by #954

PR #954, opened 2018-01-21
Added automatic setting of the timezone based on the browsers timezone

devanshk:
Addresses #953


devanshk:
Great! Thanks for the thorough test.

Issue #905, opened 2017-09-29
Improve Gradesheet interface


ryanakca:
When grading submissions via the Gradesheet interface, one can only enter one score score at a time, and you must wait for the change to be committed before moving onto the next problem. This causes (in my opinion, unbearably) large delays for those manually entering grades for many problems. Moreover, the feedback box gets in the way of entering grades in the next problem's cell. There is also no way to select the next cell for data entry, apart from clicking on it.

From a usability point of view, it would be much nicer if graders could simply tab between cells (or if there were key bindings to move between cells), and if the grade transmission/saving could happen in the background. This would save us a considerable amount of time (even half a second per problem, times a 8 problems, times 100 students, adds up to nearly seven minutes of waiting).

victorhuangwq:
I wonder if this issue is addressed (in a different manner) with the new speedgrader upgrade. https://github.com/autolab/Autolab/releases/tag/v2.4.0

PR #961, opened 2018-02-14
905 improve gradesheet interface

rahuang:
Graders/TA's can tab through all the grading cells and save their progress instead of click on each cell individually.

https://github.com/autolab/Autolab/issues/905


devanshk:
Hm when I press tab after entering a score and a feedback, the styling looks wonky and the next grade's feedback box appears, then disappears. Pressing tab again changes the focus to a footer link.

Tested in Safari and Chrome.

screen shot 2018-02-18 at 12 06 40 pm

Are you able to replicate this locally?

PR #963, opened 2018-02-27
Fix logger

zyx-billy:
Right now we use COURSE_LOGGER and ASSESSMENT_LOGGER for logging.

Oddly enough, they are defined in app/models/course_logger.rb. The only reason for them being in models seems to be that models are autoloaded by rails, so the loggers are accessible everywhere. This hack worked until we got the new ASSESSMENT_LOGGER, which no longer autoloads because it's name doesn't match the file name (course_logger.rb). This means if ASSESSMENT_LOGGER is ever referenced before a COURSE_LOGGER, we'll get an error.

This fix moves the logger into lib/autoload/, which is now set to be autoloaded by default. It also condenses the two loggers into one, logging to either the course log file or the assessment log file depending on if an assessment is set. Usage is exactly the same as before, except it's now called AUTOLAB_LOGGER.


cg2v:
ASSESSMENT_LOGGER is only used for autograder results and log submit. That seems to make sense for assessments that have updaters. Do you really want the updaters to have to deal with everything?

zyx-billy:
I see. So by "updaters", do you mean code that parses autoresult strings and assigns the initial scores to student submissions? So were the per-assessment log files intended only for autoresult-related updates from these updaters? I guess that would make sense. If so, we could maintain the current use case by creating a separate method, e.g. log_autoresult, in AUTOLAB_LOGGER. It would also help distinguish it from a naming perspective since not all assessment-related actions are logged to the "assessment_logger".

cg2v:
The assessment log seems to exist only for courses that process log submissions or postprocess autograde output. 213 and its analogs call such processing scripts updaters when they are invoked by the autolab scheduler, and I assumed that usage was known.

PR #975, opened 2018-04-20
Do not fail on receiving non-ascii from Tango

thotypous:
Currently, Autolab fails if it receives unicode (or any non-ascii characters) from Tango, e.g.

```
An Encoding::UndefinedConversionError occurred in assessments#autograde_done:

  "\xC3" from ASCII-8BIT to UTF-8
  app/controllers/assessment/autograde.rb:458:in `write'
```

This is easily fixed by opening the file for writing in binary mode. This way, data is saved exactly as received from Tango. I tested it with a feedback message encoded in UTF-8 and everything else works as expected.

### DIFF
--- a/app/controllers/assessment/autograde.rb
+++ b/app/controllers/assessment/autograde.rb
@@ -454,7 +454,7 @@ module AssessmentAutograde
 
       feedback_file = File.join(ass_dir, @assessment.handin_directory, filename)
       COURSE_LOGGER.log("Looking for Feedbackfile:" + feedback_file)
-      File.open(feedback_file, "w") do |f|
+      File.open(feedback_file, "wb") do |f|
         f.write(feedback)
       end
     end


thotypous:
Travis CI tests appear to be failing even before the changed code path is called. Is that normal?

devanshk:
It appears to be a quirk where any non-contributor PR fails the CI test. @zyx-billy I believe you found the reason for this a while back, can you give more insight on that?

In any case, we'll take a look at this manually and test it out.

zyx-billy:
Yes, right now we use encrypted environment variables (for Tango access) in our test cases. These [aren't visible to forked repos](https://docs.travis-ci.com/user/pull-requests#Pull-Requests-and-Security-Restrictions), so travis builds will fail. We really should change this in the future so everything's more self-contained, but right now what we do is we'll go through your PR manually.

thotypous:
Nice, thanks!

fanpu:
Will be resolved with https://github.com/autolab/Autolab/pull/1384

PR #980, opened 2018-04-25
Added the ability to toggle full screen mode when annotating submissions

ZackC:
I'm a TA for 15213 at CMU. In the class I annotate the submissions in the grading pane. The bootstrap layout wastes a lot of useful space, which makes grading in split screen mode difficult (I would like to be able to open the grading rubric in one screen and the code in the other). Try to use the file annotation page in half screen mode on a 13 inch monitor and you will see the that the annotation page wastes a lot of useful space.

I often manually make the changes specified in the code to improve my grading experience and thought it would be helpful for other graders.

*Note* - I have not tested this commit. I have not set up a test version of Autolab and so I am not confident that it will work correctly. I mainly wanted to provide some starter code for the change that the project maintainers can easily adjust to implement the requested feature.

Also, I think I forgot to include the javascript file on the page.  Can you fix that when you review the commit? Thanks!

### DIFF
--- a/app/assets/javascripts/fullscreenToggle.js
+++ b/app/assets/javascripts/fullscreenToggle.js
@@ -1,0 +1,23 @@
+toggleFullscreen () = {
+    var cont = document.getElementById("pageBody");
+    if(cont.classList.contains('container')){
+      cont.classList.remove('container');
+      cont.classList.add('container-fluid');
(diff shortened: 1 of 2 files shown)


ZackC:
smallannotationwindow

jbman223:
This change should be bundled into a full redesign of the code annotation page.

PR #981, opened 2018-04-25
Statistics fix for excluding excused students.

jbman223:
Fixes a bug where statistics include users who are excused from an assignment.


devanshk:
Looks good, merging

PR #986, opened 2018-05-02
API v1 hotfix (fix regrade functions)

zyx-billy:
Fixes for bugs in api_v1

- fix error handling in the regrade functions


zyx-billy:
Yes this should be good to merge. There's no other bug reports regarding this so this should be it for now.

PR #1009, opened 2018-10-17
Added Submission Time to grade export csv

devanshk:
Summary:
Adds the submission time for each grade when you bulk export.

Test Plan:
Visited Assignment page and clicked 'Bulk Export Grades'
screen shot 2018-10-17 at 1 01 08 pm

Before:
screen shot 2018-10-17 at 12 55 26 pm

After:
screen shot 2018-10-17 at 12 55 44 pm

In the case that user doesn't have a latest submission, leaves their submission time column blank:
screen shot 2018-10-17 at 12 58 33 pm

### DIFF
--- a/app/helpers/assessment_helper.rb
+++ b/app/helpers/assessment_helper.rb
@@ -47,7 +47,7 @@ module AssessmentHelper
   def gradesheet_csv(asmt, as_seen_by)
     CSV.generate do |csv|
       # title row with the column names:
-      title = ["Email:"]
+      title = ["Submission Time:","Email:"]
       asmt.problems.each { |problem| title << "#{problem.name}:" }
       title << "Total:"
       csv << title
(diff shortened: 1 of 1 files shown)


app/helpers/assessment_helper.rb
```diff
@@ -69,8 +69,10 @@ def csv_row_for(asmt, cud, as_seen_by)
     aud = AssessmentUserDatum.get asmt.id, cud.id
     throw "csv_row_for: no AUD for (#{asmt.id}, #{cud.id})" unless aud
 
-    # create csv row with user email (first column)
-    row = [cud.user.email]
+    # create csv row with latest submission time and user email (first and second columns)
+    submission = aud.latest_submission
+    submission_time = submission.nil? ? nil : submission.created_at.in_time_zone.to_s
```
devanshk: This feels a bit verbose, any way to simplify the logic here?

PR #1016, opened 2018-11-29
Drag and drop

jbman223:
Adds drag and drop functionality to the hand in page. Makes a nicer user experience going from uploading files to submitting them.

screen shot 2018-11-29 at 6 17 22 pm

screen shot 2018-11-29 at 6 17 37 pm

screen shot 2018-11-29 at 6 17 46 pm


devanshk:
Looks good. A few minor tweaks - we should make the color different so it's more indicative that the user can take an action here and have the icon change based on what type of file was submitted, but those are smaller tweaks we can do in another pr.

PR #1021, opened 2019-01-22
Use flash correctly when listing bad assessments

cg2v:
When scanning potential assessment directories and identifying those
with missing YAML files, store the messages in a list and only set
flash[:error] once. That way, both nil + string and empty flashes are
avoided.

### DIFF
--- a/app/controllers/assessments_controller.rb
+++ b/app/controllers/assessments_controller.rb
@@ -88,12 +88,13 @@ class AssessmentsController < ApplicationController
   def installAssessment
     ass_dir = Rails.root.join("courses", @course.name)
     @unused_config_files = []
+    invalid_assessments=[]
     Dir.foreach(ass_dir) do |filename|
       # skip if not directory in folder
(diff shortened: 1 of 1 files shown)


cg2v:
Untested

cg2v:
Alternatively:
```
--- a/app/controllers/assessments_controller.rb
+++ b/app/controllers/assessments_controller.rb
@@ -93,6 +93,7 @@ class AssessmentsController < ApplicationController
       next if !File.directory?(File.join(ass_dir, filename)) or filename == ".." or filename == "."
       # assessment's yaml file must exist
       if !File.exist?(File.join(ass_dir, filename, "#{filename}.yml"))
+        flash[:error] ||= ""
         flash[:error] += "Yml does not exist: " + filename +"     -     "
         next
       end
```

damianhxy:
Was fixed in #1023

damianhxy:
Was fixed in #1023

PR #1026, opened 2019-03-11
Add notify callback url after submit

ronzalo:
When an submission is created through the API, Autolab would save a URL to notify after the submission is qualified. The goal is to avoid polling on Autolab to get the score, so Autolab notifies inmediatly to the 3rd application and the application knows when to gets the score and feedback


damianhxy:
Closing this stale PR as the feature does not seem to be needed in the foreseeable future.

PR #1030, opened 2019-04-04
Added options to new API scopes field

TheodorJ:


damianhxy (approved):
LGTM - text agrees with [doorkeeper config](https://github.com/autolab/Autolab/blob/master/config/initializers/doorkeeper.rb#L58-L59).

PR #1035, opened 2019-07-10
fixed Base64 typo

wlnirvana:
Accessing `http://localhost:8000/courses/AutoPopulated/assessments/labtemplate/extensions` results in

```txt
LoadError at /courses/AutoPopulated/assessments/labtemplate/extensions
cannot load such file -- Base64
```

This is because the `Base64` module actually uses lowercase `b`. See the [doc here](https://ruby-doc.org/stdlib-2.2.2/libdoc/base64/rdoc/Base64.html).


fanpu:
LGTM

PR #1048, opened 2019-10-09
Speedgrader handles edge cases with empty files in handin.tar

### LINKED ISSUE #1036: Annotate submission page cannot display empty files properly
In the annotation submission page, empty files (could be `__init__.py` or something like touch files) in tarball submission cannot be displayed properly. The modal spinner just goes on forever.

Alternatively, say if the 3rd file in the uploaded tarball is empty, one should be able to reproduce the error by visiting URL like `http://localhost:8000/courses/cs101-fall2019/assessments/lab1/submissions/1846/view?header_position=3`, which will redirect to the assessment page.

I believe this is caused by an empty (because the file is empty) string returned [by this read method][1], then treated as nil in [this conditional][2], leading to the [redirection][3].

[1]:https://github.com/autolab/Autolab/blob/081a07cb60a5f6de8f0d7b627d75e7d18a612bf2/lib/archive.rb#L145
[2]:https://github.com/autolab/Autolab/blob/081a07cb60a5f6de8f0d7b627d75e7d18a612bf2/app/controllers/submissions_controller.rb#L308
[3]:https://github.com/autolab/Autolab/blob/081a07cb60a5f6de8f0d7b627d75e7d18a612bf2/app/controllers/submissions_controller.rb#L310

TheodorJ:
Fixes #1036

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -305,6 +305,11 @@ class SubmissionsController < ApplicationController
 
     if params[:header_position]
       file, pathname = Archive.get_nth_file(@submission.handin_file_path, params[:header_position].to_i)
+
+      if(file.nil?)
+        file = ""
+      end
+
       unless file && pathname
         flash[:error] = "Could not read archive."
         redirect_to [@course, @assessment] and return false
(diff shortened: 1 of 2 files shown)


fanpu:
Tested with empty file and empty tar.gz archive and working, awesome!

Issue #967, opened 2018-03-02
How to add a batch of users with default passwords?


cs181-ai:
I know how to add multiple users with Roster, but it seems that I cannot set a default password for each of them. Since I don't  want to setup a email service, I am looking for a way to add a batch of users with specified passwords and email address. Is it possible?

zyx-billy:
There's unfortunately no way right now to specify the temp password when creating a user from a roster. Is it possible to have the users register themselves?

PR #1049, opened 2019-10-10
Removed Bulk Email functionality

TheodorJ:


mtoupsUNO:
Why remove this feature? We use it here and it works fine.

fanpu:
@mtoupsUNO Thanks for your comment! Within CMU internally we found out that practically none of the professors are using this feature since there are now significantly better mailing tools like AWS SES, the bulk email functionality was introduced at a time when such tools were not as popular. Therefore we decided to remove the feature to reduce bloat and prevent security vulnerabilities of running an SMTP server. We actually did not expect other schools to also be using the feature, but since you mentioned that your school is using it we will review our decision to deprecate this functionality to maintain support for existing workflows. Thanks for proactively spotting this!

mtoupsUNO:
Ah, I see, thanks for the background. I noticed some discussion on the autolab slack today that confirmed other schools are also interested in keeping this feature.

I do think it is worth reconsidering the role of SMTP servers with AutoLab, but in the end I think "send bulk email" should be an optional feature (if admins decide to enable it by uncommenting the relevant lines in `production.rb`). Then it could be disabled by default for schools where connecting Autolab to SMTP is not desired.

Rather than removing this one admin function, I think a better idea would be to remove dependency on SMTP in order to install Autolab. (Note that the Autolab OneClick install guide says an email service is required, one of the few external requirements: https://autolab.github.io/docs/one-click/)

Right now there is no way for admins to set passwords for users, this can only be done via an emailed password reset link. See #967 

Also, account registration depends on the same email functionality. Many schools (like us) will use external authentication like Shibboleth or LDAP, so account registration emails are moot. But for people installing and testing Autolab for the first time, setting up SMTP is one of the hurdles to getting everything installed and running.

So I think removing the dependency on SMTP in account creation and password reset would be good, but please leave SMTP and instructor emails available as an optional feature.

TheodorJ:
@mtoupsUNO Understood. We didn't know there were still people who wanted this, so thanks for bringing it up with us. I'll add the relevant files back in and add an option to safely disable the SMTP dependency.

PR #1060, opened 2019-10-16
Fixing checkbox

XtrKiL:
Fixed Checkboxes at mainly two places
1. MOSS Submission page
2. Delete Submission Page

The main reason why the checkboxes weren't working was because of the upgrade to the new materialize. Need to follow the format in the link below.
https://materializecss.com/checkboxes.html


XtrKiL:
Fixed the form builder, such that a materialized checkbox is generated as well.

PR #1064, opened 2019-10-27
Speed Grader PDF Annotations Fixes

XtrKiL:
Fixed Unable to Download Annotated Files
Changed text boxes on pdf to annotations on pdf
UI improvements to annotation (minimum resize, commented out annotation movement)

### DIFF
--- a/app/assets/javascripts/annotations.js
+++ b/app/assets/javascripts/annotations.js
@@ -205,13 +205,13 @@ $(document).keydown(function(e) {
 
 /* Some Helper functions */
 function copyToClipboard(str) {
-  let el = document.createElement('textarea'); // Create a <textarea> element
+  var el = document.createElement('textarea'); // Create a <textarea> element
   el.value = str; // Set its value to the string that you want copied
   el.setAttribute('readonly', ''); // Make it readonly to be tamper-proof
   el.style.position = 'absolute';
   el.style.left = '-9999px'; // Move outside the screen to make it invisible
   document.body.appendChild(el); // Append the <textarea> element to the HTML document
-  let selected =
+  var selected =
     document.getSelection().rangeCount > 0 // Check if there is any content selected previously
     ?
     document.getSelection().getRangeAt(0) // Store selection if found
--- a/app/assets/javascripts/annotations.js
+++ b/app/assets/javascripts/annotations.js
@@ -712,7 +712,7 @@ var newAnnotationFormForPDF = function(pageInd, xCord, yCord) {
       var yRatio = yCord / $("#page-canvas-" + pageInd).attr('height');
 
       var widthRatio = 200 / $("#page-canvas-" + pageInd).attr('width');
-      var heightRatio = 110 / $("#page-canvas-" + pageInd).attr('height');
+      var heightRatio = 135 / $("#page-canvas-" + pageInd).attr('height');
 
       submitNewPDFAnnotation(comment, value, problem_id, pageInd, xRatio, yRatio, widthRatio, heightRatio, newForm);
     }
@@ -1020,6 +1020,7 @@ var makeAnnotationMovable = function(annotationEl, annotationObj) {
     var curWidth = (positionArr[3] || 120);
     var curHeight = (positionArr[4] || 60);
 
+    /** 
     $(annotationEl).draggable({
       stop: function( event, ui ) {
         var xRatio = ui.position.left / $page.attr('width');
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/annotations.js
```diff
@@ -1020,6 +1020,7 @@ var makeAnnotationMovable = function(annotationEl, annotationObj) {
     var curWidth = (positionArr[3] || 120);
     var curHeight = (positionArr[4] || 60);
 
+    /**
```
jbman223: Are you sure we want to remove this function? Now it is no longer possible to drag an annotation to change its position on the document. This seems to me like a feature regression, unless I missed a discussion about removing this feature.

XtrKiL:
I changed the way the annotation boxes are made, it isn't really an "annotation" box that is drawn on the pdf, but an actual pdf comment. 

image

If you hover your mouse over you get: 
image

Can you check if that is the case for you?

app/assets/javascripts/annotations.js
XtrKiL: I intentionally wanted to remove the ability to drag around the annotation to streamline the adding annotation process.

XtrKiL:
New changes have been added. 
It has been reverted to being able to move and resize.
At the same time, there is a redesigned form, and it can minimize and maximize as well.

app/assets/javascripts/annotations.js
XtrKiL: When we create the pdf on a ubuntu based deployment, the pdf comment does show up, even when downloaded onto a Mac OS X. After discussion we decided that making this feature work on a mac os x deployment of autolab is not worth the time, as of now (because Autolab would most likely be deployed onto a Ubuntu system).

XtrKiL:
This PR also fixes the out of ordering of the annotations, as well as makes the ensuring the behavior of scroll to line

app/controllers/submissions_controller.rb
```diff
@@ -243,21 +243,28 @@ def download
...
           if !problem.nil? then problem_name = problem.name end
 
-          comment = "#{annotation.comment}\n\nProblem: #{problem_name}\nScore:#{value}"
+          comment = "#{annotation.comment}\n\nProblem: #{problem_name}\nScore: #{value}"
 
           # + 1 since pages are indexed 1-based
           pdf.go_to_page(page + 1)
-
+          
+          # Below is old code for drawing boxes and adding text to pdf
```
fanpu: I think we can remove this chunk of old code, doubt anyone will miss it. 
I have also fixed the header_position checking logic with the change in https://github.com/autolab/Autolab/pull/1068
Otherwise the backend code is LGTM - I will defer the frontend to @jbman223 !

amartinez116:
As mentioned earlier, comments are being cutoff when inserted near the bottom of the page. Looks like the next page of the pdf is blocking them. Hard to tell when the next page starts when submitting handwritten pdfs. 

screenshot2

XtrKiL:
The issue with regards to the annotation box being behind should now be fixed.

app/views/submissions/_annotation_pane.html.erb
```diff
@@ -14,7 +14,7 @@
             </div>
             <div class="collapsible-body">
               <% descriptTuples.each do |description, value, line, user, id| %>
-                <div class="descript" id="li-annotation-<%= id %>" onclick="scrollToLine(<%= line %>)">
+                <div class="descript" id="li-annotation-<%= id %>" onclick="scrollToLine(<%= (line.nil? ? nil : line+1) %>)"> <!-- Line + 1 because the code line numbers starts with 1 not. -->
```
amartinez116: Did you mean to say "... line number starts with 1 not 0."

XtrKiL:
Addressed issue of creating multiple warnings when trying to add blank comments. Fixed editing mode that is affected by this update as well

app/assets/javascripts/annotations.js
```diff
@@ -644,7 +644,10 @@ function newAnnotationBoxForPDF(annObj) {
     $(body).hide();
     $(edit).hide();
     $(score).hide();
+    $(minimize).hide();
     var form = newEditAnnotationForm(annObj.line, annObj);
+    $(box).draggable( 'disable' );
+    $(box).resizable( 'disable' );
```
amartinez116: Making the box resizing "disable" can cause users to expand the comment box and hide the buttons since the box is not expanding. I suggest either getting rid of this line or restricting the expansion of the comment box.

Screen Shot 2019-12-05 at 1 06 33 AM

Screen Shot 2019-12-05 at 1 09 20 AM

XtrKiL: Fixed the edit form expandable.

XtrKiL:
Removed the old code to prevent cluttering

PR #1069, opened 2019-11-18
Compute non-autograde scores automatically based on annotations

oliverli:
This PR improves the existing grading workflow by instructors. Previously, instructors had to copy the new score based on the sum of the modifiers from the annotation and apply them to the gradebook manually. This PR now allows the scores to be automatically updated based on the created annotations. This only applies for scores that were non-autograded, since we do not want to change autograded scores to remain idempotent.

Changes:
- Only allow users to create annotations on non-autograded problems
- Frontend UI updates score automatically
- On creation/update/destroy of an annotation, the score for the problem will be updated in the db as such: `score_problem = problem_max_score + sum of values of annotations`

Manual Tests:
- Users not able to create annotations on scores created by autograder
- Users able to create annotations on scores not created by autograder

Unit Tests:
- Score updates when annotation that changes score is applied according to formula

### DIFF
--- a/app/assets/javascripts/annotations.js
+++ b/app/assets/javascripts/annotations.js
@@ -205,13 +205,13 @@ $(document).keydown(function(e) {
 
 /* Some Helper functions */
 function copyToClipboard(str) {
-  let el = document.createElement('textarea'); // Create a <textarea> element
+  var el = document.createElement('textarea'); // Create a <textarea> element
   el.value = str; // Set its value to the string that you want copied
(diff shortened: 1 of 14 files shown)


fanpu:
@TheodorJ this is ready for review now!

TheodorJ:
Have tested:
 - Adding annotations that modify a score created by autograder
    - Update is not applied automatically
 - Adding annotations that modify a score NOT created by autograder
    - When score exists, annotation modifies existing score automatically
    - When score does not exist, a score object is created and the change is applied automatically

I'd say this is good to go!

PR #1072, opened 2019-11-26
Fix Speedgrader Div Size Class

XtrKiL:
image

As of now if you resize the screen and made it too small, the code grader will disappear. 
This PR changes the class type of the divs to ensure they show up even on small screens.

image


XtrKiL:
Closing PR for duplicate commits

PR #1076, opened 2019-12-06
Ctags functions

amartinez116:
Symbol tree account for functions in classes. 

"--extras=+q " provides class-qualified tag entry for each tag which is a member of a class. However, it does creates duplicates having both  "class.method" and  "method" in tag file.

### DIFF
--- a/app/assets/stylesheets/annotations.css
+++ b/app/assets/stylesheets/annotations.css
@@ -636,6 +636,7 @@
 
 #symbol-tree-box ul li {
   cursor: pointer;
+  display: flex;
 }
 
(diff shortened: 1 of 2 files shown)


TheodorJ:
Looks good to me! It's worth noting that, as per [The Ctags Docs](http://ctags.sourceforge.net/ctags.html), the `--extras=+q` flag should only parse class tags in C++, Eiffel, and Java, but on my setup it's emitting class tags for Python too (Haven't tested other languages).

At some point, we might want to go through the languages that Autolab supports and just verify that Ctags emits class tags for each.

PR #1078, opened 2019-12-07
Annotation requires problem fix

fanpu:
It is currently possible to create annotations without an associated problem, resulting in errors when attempting to update the score for non-autograded problems since the max_score of the problem cannot be determined.

Furthermore, in re-computing the score of the problem, we are identifying all annotations with the associated `problem_id` and `submission_id`, so it is essential for both fields to be present when saving.

Frontend changes:
Prevent users from creating annotation if there are no problems selected
Show a helpful message if there are no autograded problems available

Backend changes:
Validation that `problem_id` is present on the annotation is done before saving
In computing the score, if the max_score of the problem is nil, it defaults to 0

### DIFF
--- a/app/assets/javascripts/annotations.js
+++ b/app/assets/javascripts/annotations.js
@@ -310,13 +310,20 @@ function displayAnnotations() {
 }
 
 function attachEvents() {
-  $(".add-button").on("click", function(e) {
-    e.preventDefault();
-    var line = $(this).parent().parent().parent();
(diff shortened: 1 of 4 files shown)


victorhuangwq:
I have checked that 
- In computing the score, if the max_score of the problem is nil, it defaults to 0

In the commits I have made
- Users will not be able to create annotation if a problem has not been selected
- Users will receive additional warning on creating problems if there are no problems in the first place

PR #1081, opened 2019-12-23
Fixes Edit Annotation Issues and Resizing Annotation Issues for PDF view mode

victorhuangwq:
- Created a shared template form for add and edit. 
Now the two forms should be visually similar, changing elements on both forms should become easier to handle.
- Fixed issues with resizing. 
Previously resizing does not work as it snaps back to a pre-set sized. Now the resized box should stay.
- Further edited styles of annotation box. 
There was a weird gap in between the comment textarea and the scores. Removed that.
- Fixed issue with edit annotation not updating body text
Previously if you edit the annotation, the comment body is not being updated after you click submit.

### DIFF
--- a/app/assets/javascripts/annotations.js
+++ b/app/assets/javascripts/annotations.js
@@ -1,3 +1,9 @@
+/* Document On Ready Initializations */
+
+$(document).ready(function(){
+  $(".collapsible-body").show(); //expands all collapsible initially 
+});
+
 /* File Tree and Code Viewer Helper Functions */
 
 // Lets us open and close file folders
--- a/app/assets/javascripts/annotations.js
+++ b/app/assets/javascripts/annotations.js
@@ -665,7 +665,7 @@ function newAnnotationBoxForPDF(annObj) {
     $(box).draggable( 'disable' );
     $(box).resizable( 'disable' );
     $(box).append(form);
-    $(box).width("300px");
+    $(box).width("auto");
     $(box).height("auto");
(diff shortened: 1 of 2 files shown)


fanpu:
Great job on making things more modular and reusable!

There's one bug that I found so far, which can be replicated as follows:
1. Create annotations for the first problem
2. Create annotations for the other problems
3. Refresh the page

After refreshing, only the annotation text for the first problem shows up in the annotation container on the right:
DeepinScreenshot_select-area_20191224120829

Only by dragging any of the annotations will the annotation text of the other annotations be displayed in the container

victorhuangwq:
I have taken a look at the issue you have mentioned. 

When you add an annotation for a problem, the collapsible automatically expands and shows you everything that is in it. That is if you add an annotation for problem 1, and then add an annotation for problem 2, both collapsibles will be expanded to show you what's under both problem 1 and problem 2.

What happens when you refresh the page, the collapsible only expands for the first problem, and you have to click on the second problem to expand to see what's inside. 

So I would say the issue is more of an user interface issue. There are two options.
1. Add symbols to signpost to users that "Problem0" and "Problem1" are collapsibles and you can click on them.
2. Make them not collapsible and always show all the annotations. 

I would feel option 1 is more suitable, especially in the event where there are many annotations.

victorhuangwq:
Addressed the issue of not showing all collapsible elements initially, using option 2 as described above.

PR #1086, opened 2020-01-31
Fixes directories not showing up, by parsing directories on the fly

victorhuangwq:
The sanitization should work for both archives with directories, and archives without directories.


TheodorJ:
This looks good! Works on my setup.

PR #1087, opened 2020-02-01
Explicitly test value on whether scores should be released

fanpu:
In setting `@scores` to determine whether the scores should be released, currently this is being done:

`released: result["released"].to_i`

However, the "boolean" value in result["released"] takes on string values of "t" and "f", which always gets casted to 0 under `.to_i` (see https://apidock.com/ruby/String/to_i)

The question I'm wondering is how this piece of code could have ever worked, unless it used to  be "0"/"1" values which used to be returned. Could this possibly be a db-specified issue? I am using sqlite for my dev environment

I have fixed this by using a helper method is_truthy?() to test whether a given value matches one of the common "truthy" expressions, and using this to determine the value of released

### DIFF
--- a/app/controllers/assessments_controller.rb
+++ b/app/controllers/assessments_controller.rb
@@ -4,6 +4,7 @@ require "fileutils"
 require "rubygems/package"
 require "statistics"
 require "yaml"
+require "utilities"
 
 class AssessmentsController < ApplicationController
   include ActiveSupport::Callbacks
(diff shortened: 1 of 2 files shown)


victorhuangwq:
- Release grades make annotation adjustments appear on submission page
- Withdraw grades withdraw such grades
**LGTM**

fanpu:
Confirmed to be a DB issue:
```
DEPRECATION WARNING: Leaving `ActiveRecord::ConnectionAdapters::SQLite3Adapter.represent_boolean_as_integer`
set to false is deprecated. SQLite databases have used 't' and 'f' to serialize
boolean values and must have old data converted to 1 and 0 (its native boolean
serialization) before setting this flag to true. Conversion can be accomplished
by setting up a rake task which runs```

PR #1089, opened 2020-02-01
Fix for student can view Speedgrader scores/annotations before it is released

fanpu:
Students are currently able to view annotations, grade changes, and scores by going to the View tab of the submission, even before the grades are released. This PR ensures that only instructors and course assistants are able to view and modify these details before they are released.


victorhuangwq:
- Release all grades works
- Withdraw all grades works 
- Handles situation for when score doesn't exist yet

**LGTM**

PR #1090, opened 2020-02-02
lines with over 80 chars are now highlighted

amartinez116:
On default, the checkbox that highlights lines over 80 characters is on. Users are able to turn it that off by unchecking the box.

image


victorhuangwq:
- Checkboxes work
- Highlights lines above 80 chars

**LGTM**

PR #1095, opened 2020-02-08
Go to correct file when annotation in pane is clicked (rails-5)

oliverli:


oliverli:
> Switches to correct file. One thing to note: when clicking on an annotation in the same file, entire page refreshes.
> LGTM!

Have fixed the issue when the file position is 0

Sometimes the entire file will still refresh because it was loaded from cache - I think that's another issue altogether with the cache that we should work on the future. For now this should be good to go!

PR #1096, opened 2020-02-13
Loosen API rate limiting

TheodorJ:
Two changes here:
 - Removed rate limiting on general api routes
     - All routes except file upload
     - These are cached in the CLI, so putting a cli call in a bash for loop would only generate one request anyways.
     - Students could use curl to make raw HTTP requests to the API, but that's just as difficult as making requests to normal GUI routes that aren't rate-limited anyways.
 - Increased /oauth/device_flow_* limits to 1000
    - This should accommodate situations where every student in a class sets up the Autolab CLI at once.

### DIFF
--- a/config/initializers/rack-attack.rb
+++ b/config/initializers/rack-attack.rb
@@ -39,24 +39,6 @@ class Rack::Attack
 
   ### Throttle Requests ###
 
-  # Throttle setup requests by user id
-  #
-  # Key: "rack::attack:#{Time.now.to_i/:period}:api/general:#{req.user_id}"
-  throttle('api/setup', :limit => 1500, :period => 0.seconds) do |req|
-    if req.path.start_with?("/api/")
-      req.user_id
-    end
-  end
-
-  # Throttle all requests by user id
-  #
-  # Key: "rack::attack:#{Time.now.to_i/:period}:api/general:#{req.user_id}"
-  throttle('api/general', :limit => 10, :period => 30.seconds) do |req|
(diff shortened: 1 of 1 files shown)


config/initializers/rack-attack.rb
```diff
@@ -39,15 +39,6 @@ def user_id
 
   ### Throttle Requests ###
 
-  # Throttle all requests by user id
```
fanpu: I'm not sure if we want to completely remove this even though we have caching, as it is possible that different resources are being requested that may result in some form of cache eviction eventually. I think it would still be safer to leave this in

config/initializers/rack-attack.rb
```diff
@@ -69,7 +60,7 @@ def user_id
   # Throttle requests for device_flow_authorize
   #
   # Key: "rack::attack:#{Time.now.to_i/:period}:oauth/device_flow_authorize:#{req.user_id}"
-  throttle("oauth/device_flow_authorize", :limit => 1, :period => 5.seconds) do |req|
+  throttle("oauth/device_flow_authorize", :limit => 1000, :period => 5.seconds) do |req|
```
fanpu: Propose increasing the period to say 15 minutes, since it is unlikely that the class size would exceed that when registering, and reduces our risk to ddos

fanpu:
Other than the comments I made the rest looks good to me!

damianhxy:
Stale PR, reopen if throttling poses an issue in the future

PR #1097, opened 2020-02-14
Changes deprecated render text to render plain

victorhuangwq:
This fixes the issue of the scores not rendering correctly after updating them on the
1. view page
2. grade submissions page

It should update correctly now

### DIFF
--- a/app/controllers/assessment/grading.rb
+++ b/app/controllers/assessment/grading.rb
@@ -241,7 +241,7 @@ public
 
     updateScore(score.submission.course_user_datum_id, score)
 
-    render text: score.score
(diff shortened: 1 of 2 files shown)


fanpu:
Tested working:

- Updating grades on Speedgrader view page results in grades updated on frontend and backend
- Updating grades on Gradesheet page results in grades being updated on frontend and backend

I also noticed that the first time I try to modify the grades on the Speedgrader grades section, clicking on the score (or -) has no effect, and refreshing the page must be done in order to be able to click on it. However this is something we can solve later.

LGTM!

PR #1102, opened 2020-02-22
Makes annotation summary collapsible. Adds correct padding

victorhuangwq:
- Makes annotation box summary collapsible by initializing it
- Fixes padding issue with the internal elements in the annotation summary box

image


amartinez116:
LGTM! 

Works!

PR #1113, opened 2020-03-08
Added highlight when the score is successfully submitted

victorhuangwq:
Before submit
image
After submit
image


victorhuangwq:
A different implementation was used. Refer to https://github.com/autolab/Autolab/pull/1118

PR #1114, opened 2020-03-09
Update Readme.md to remove the old update

victorhuangwq:
Remove the old update that says that rails 5 is still WIP


fanpu:
Lgtm

PR #1116, opened 2020-03-12
Copy of GDU/PLD fix when bulk importing grades for non-existent submission for Rails 5

fanpu:
Copy of https://github.com/autolab/Autolab/pull/1115 into master


fanpu:
I'm holding off on merging this until I figure out why the tests are failing, seems to be a db issue at first glance but doesn't seem like we changed anything that would start causing it to fail

fanpu:
@victorhuangwq  could you approve this again

PR #1117, opened 2020-03-24
Hide instructor/CA's score from scoreboard

xinyis991105:
Bypass any record from instructors/CAs so the records passed on to the template will only be those of the students.


fanpu:
Tested working as follows:

1. Add scoreboard to course
2. Find a student on scoreboard, promote to CA, check that scoreboard entry disappeared
3. Demote back to student, check that scoreboard entry back
4. Repeat 2 and 3 for instructor role

Very neat and concise code that gets straight to the point, good job!!

Just as a side note for the future, you can name your branch as the name of the feature you are working on (i.e hide-non-students-from-scoreboard) to be more descriptive

PR #1118, opened 2020-03-26
Removes ability to change grades on view code page

victorhuangwq:
This PR removes the ability to change the grades directly on the view page, as it should have been replaced by the annotation feature.

- Grades should no longer be editable
- The gradebox no longer darkens on hover (so that the user don't think there is something there)
- The cursor on hover is also no longer a pointer, but just a regular cursor (to prevent the users from thinking there is something clickable)


fanpu:
So I know @amartinez116 was assigned to review this on Github but I forgot to make this a task on Asana this sprint, and since I was reviewing PRs anyway I decided to just hop on this so we could get this merged faster. Really sorry, my bad! 

Tested working, but a small nit: could we change the text color for the grades from blue to something else (maybe gray like the RHS of the score)? This is because in most of the UI on the page, blue indicates something clickable, which may be confusing/frustrating for users

victorhuangwq:
- Changed students score from blue to grey (to prevent people from thinking that it is a link)

amartinez116:
What about to a lighter gray in order to differentiate the two? Here it is when I changed the color to "color: darkgray" using inspect element. 

image vs image

victorhuangwq:
I tried that initially, but I was thinking, wouldn't making the student's grade light grey, and the maximum grades dark grey, kind of draw the attention away from the student's grade?

My current thinking is that they should be different color. For example green (for the students score) or black for the max score. So long the color isn't blue it shouldn't be confused as a link right?

amartinez116:
This is what it would look like with green. I actually don't mind these. However will people think it's clickable?

image image

victorhuangwq:
Adjusted the color such that it is grey before the grades are released
image

Dark green after it is released. I think the dark green stuck with me.
image

PR #1119, opened 2020-03-26
Red rows bulk import

amartinez116:
On Bulk Upload Grades users will only view errors in the uploaded data

image

_bulkGrade_error_entries.html.erb will only show rows that contain errors while _bulkGrade_entries.html.erb remains unchanged to show all final entries.

### DIFF
--- a/app/views/assessments/bulkGrade.html.erb
+++ b/app/views/assessments/bulkGrade.html.erb
@@ -13,16 +13,16 @@
   <% end %>
 <% end %>
 
-  <%= render "bulkGrade_entries", :entries => @entries, :problems => @assessment.problems if @entries %>
-
   <% if @valid_entries %>
+    <%= render "bulkGrade_entries", :entries => @entries, :problems => @assessment.problems if @entries %>
     <%= form_for :confirm, :url => "bulkGrade_complete", :html => { :class => "confirm" } do |f| %>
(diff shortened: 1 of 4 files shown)


app/views/assessments/bulkGrade.html.erb
```diff
@@ -13,16 +13,16 @@
...
-  <%= render "bulkGrade_entries", :entries => @entries, :problems => @assessment.problems if @entries %>
-
   <% if @valid_entries %>
+    <%= render "bulkGrade_entries", :entries => @entries, :problems => @assessment.problems if @entries %>
     <%= form_for :confirm, :url => "bulkGrade_complete", :html => { :class => "confirm" } do |f| %>
       <%= f.hidden_field :bulkGrade_csv, :value => @csv %>
       <%= f.hidden_field :bulkGrade_data_type, :value => @data_type %>
 
-      <%= f.submit 'Yes' , {:class=>"btn submit"} %><%= link_to "No", { :action => :bulkGrade }, :class => "no" %>
+      <%= f.submit 'Yes' , {:class=>"btn submit"} %><%= link_to "No", { :action => :bulkGrade }, :class => "btn submit" %>
```
amartinez116: Before with class => "no", the buttons looked like: 
image

Now they look like: 

image

PR #1121, opened 2020-04-02
Adds link to documentation about interaction between gradesheet and annotations

victorhuangwq:
Link added in the gradesheet page.

image


victorhuangwq:
Use something other than red for the font

victorhuangwq:
Changed the color to a black and blue to show that it is not a warning

image

PR #1122, opened 2020-04-02
Create Advanced tab for edit assignments

oliverli:
For less commonly used edit assignment settings


victorhuangwq:
I have tested the PR locally, and it implements the Advanced tab properly. All the original fields are working. 

The layout in the advanced tab is not as ideal still, but as this PR improves the other tabs, I think this should be merged in, and the layout in the advanced tab can be addressed at a later time (since it is less used anyways).

PR #1123, opened 2020-04-05
Update bootstrap-sass from 3.3.7 to 3.4.1

oliverli:
Missed one out because it was limited to its minor version in the Gemfile


fanpu:
Tested with some more complex widgets like date pickers, looked out for some checkboxes and noticed they were there, overall probably safe change since it's a minor version, LGTM!

PR #1124, opened 2020-04-07
GDU/PLD Surfacing Improvement

xinyis991105:
This PR improves the surfacing of late-day/grace-day information to students when they are submitting late. It adds extra information to the academic integrity confirmation statement as well as under "number of submissions left" message. And this information does not show up if a student is not submitting late.

### DIFF
--- a/app/views/assessments/_handin_form.html.erb
+++ b/app/views/assessments/_handin_form.html.erb
@@ -15,7 +15,7 @@
     <%   else %>
     <%     grace_late_info = "submitting #{pluralize(days_late, "day")} late using #{pluralize(days_late, "penalty late day")}" %>
     <%   end %>
-    <%   late_confirm = " and that I'm #{grace_late_info} under the late policy as defined in the syllabus" %>
+    <%   late_confirm = " and that I am #{grace_late_info} under the late policy as defined in the syllabus" %>
       <p><b>Warning:</b> Submitting late may result in the usage of a grace day or a grade penalty!</p>
     <% end %>
 
--- a/app/views/assessments/_handin_form.html.erb
+++ b/app/views/assessments/_handin_form.html.erb
@@ -9,13 +9,13 @@
     <%   days_late = (days_late / 1.day).ceil %>
     <%   grace_days_left = @aud.grace_days_usable %>
     <%   if grace_days_left >= days_late %>
-    <%     grace_late_info = "using #{pluralize(days_late, "grace day")}" %>
+    <%     grace_late_info = "submitting #{pluralize(days_late, "day")} late using #{pluralize(days_late, "grace day")}" %>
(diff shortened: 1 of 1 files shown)


xinyis991105:
Current phrasing:
1. "using x grace days with y additional late days"
2. If x = 0, don't show grace days at all; if grace days can cover all days late, don't show additional late days.
Proposed phrasing:
1. "submitting (x+y) days late using x grace days with y penalty late days"
Things to be confirmed:
1. Do we want to show both grace days and penalty late days regardless of the different cases? For example, if we don't have any grace days left, do we want to say "0 grace days" or not show grace days at all?
2. Are "grace days" and "penalty late days" the terms to use?
Thanks!

victorhuangwq:
"submitting (x+y) days late using x grace days with y penalty late days"

I think the above phrasing is good. 

1. I think show depending on whether it has been used or not (if no grace days used, don't show it). So for clarification I think the cases are

**Submitting on Time** : Do not show grace days or penalty late days
**Submitting Using Grace Days**: Show grace days used
**Submitting with Penalty Late Days**: Show with how many penalty late days
**Submitting with both grace days and penalty late days**: Show both

2. Yes, it is grace days and penalty late days.

xinyis991105:
I have updated the phrasing! Here are some screenshots of how they look for all four cases mentioned above:
Screen Shot 2020-04-09 at 11 09 40 PM
Screen Shot 2020-04-09 at 11 09 53 PM
Screen Shot 2020-04-09 at 11 10 04 PM
Screen Shot 2020-04-09 at 11 10 39 PM

amartinez116:
Wording looks good! Small nit: change "I'm" to "I am" 

Also, in the last picture what does "(2 unpenalized, 5 total submissions left)" mean?

xinyis991105:
It's part of the original remainingSubmissionsMsg method. I played with the settings of the assessments a little bit! It doesn't interfere with the surfacing of GDU/PLD. "I'm" has been changed to "I am"!

amartinez116:
Small edge case, not sure if it’s something we want to worry about but what if the value of the penalty late day is zero? The message still shows up with “… with xx late penalty late days " even though there is no penalty to the grade. Do we still want to show the message?

xinyis991105:
I think it's up to how the instructors actually use penalty late days and grace days. I would find it a bit weird if a penalty late day doesn't incur any penalty. But I would vote for still including "penalty" just to signify that the student is submitting late and using grace days is not sufficient. Also, I think the phrasing on Autolab's documentation justifies the use of possibly 0 penalty. What do other people think?
Screen Shot 2020-04-11 at 5 16 45 PM

amartinez116:
In that case, I would agree and say that it is a good way to distinguish between grace days and penalty days (even for some reason the penalty is zero). It all looks good to me!

PR #1143, opened 2020-05-01
Course dashboard update #2

xinyis991105:
This PR adds the course dashboard feature to the Autolab application. Compared with the previous PR, this one fixes the following issues:
1. Grace day calculation scheme is corrected. Current scheme: find the latest assessment of the course (the last one in the dependency chain of the assessments), get the corresponding assessment user datum and the number of grace days used cumulatively up until that assessment, and subtract that number from the course grace day number. When there is no assessment, the number of course grace days is returned.
2. When the course website button is clicked on the dashboard, it will create a new tab, instead of refreshing the current tab directly.

I tested locally by creating a new course with three assessments that are past deadline. After submitting for each assessment, I come back to check the new grace day number displayed on the dashboard and it seems to be consistent this time. But further testing is needed. Once this PR is approved, I will close the old one and delete the other compromised branch. Thanks!

### DIFF
--- a/app/models/assessment_user_datum.rb
+++ b/app/models/assessment_user_datum.rb
@@ -224,6 +224,10 @@ class AssessmentUserDatum < ApplicationRecord
     connection.execute insert_sql
   end
 
+  def global_cumulative_grace_days_used
+    cumulative_grace_days_used
+  end
+
 protected
 
   def cumulative_grace_days_used
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -447,7 +447,7 @@ private
   end
 
   def edit_course_params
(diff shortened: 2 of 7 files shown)


victorhuangwq:
I have tested the following on the PR:
1. Adding course website. Works as expected. 
2. Submitting assignments to see if grace days used shown on the course dashboard is reflective. Works as expected. Tested by creating a new course and adding fresh assignments, and slowly submitting to each of them.

app/models/assessment_user_datum.rb
```diff
@@ -224,6 +224,10 @@ def self.create_modulo_callbacks(params)
     connection.execute insert_sql
   end
 
+  def global_cumulative_grace_days_used
```
victorhuangwq, line 227: Just to clarify, the reason you created this public function is because you can't just use the protected function?

xinyis991105: Yes!

PR #1147, opened 2020-05-10
Guard against nil error in Gradebook

fanpu:
This guards against the issue
```
  nil can't be coerced into Integer
  app/helpers/gradebook_helper.rb:91:in `+'
```
when we get a reference to an empty cell. I'm guessing this is most likely due to deletion of data somewhere that got it into an inconsistent state. Unfortunately, I have tried to delete an assessment with submissions but was not able to replicate the error.


fanpu:
Updated with comments, could you review if this is alright @oliverli

fanpu:
Sorry about my careless mistake - I've fixed it now!

PR #1152, opened 2020-05-14
Hide contents of binary files on speedgrader

fanpu:
This fixes the issue where we obtain an error when navigating to binary files on the Speedgrader interface:

`submissions view (ActionView::Template::Error) "invalid byte sequence in UTF-8"`

This is achieved by simply not displaying the invalid byte sequences at all, as they are binary and not UTF-8 formatted. Instead, a message "Binary files not displayed" is shown. This also helps to improve overall user experience, and speed up loading times when users accidentally navigate to a large binary file.

Original UI:
DeepinScreenshot_select-area_20200514163751

After this fix:
DeepinScreenshot_select-area_20200514163806

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -322,12 +322,8 @@ class SubmissionsController < ApplicationController
       @displayFilename = @submission.filename
     end
 
-    filetype = FileMagic.open(:mime) { |fm|
-      fm.buffer(file)
-    }
-
-    # Do not display binary files, unless if they are PDF
-    if (filetype =~ /pdf/).nil? and (filetype =~ /(binary|executable)/).present?
+    mm = MimeMagic.by_magic(file)
+    if not mm.text? and not mm.subtype == "pdf"
         file = "Binary file not displayed"
     end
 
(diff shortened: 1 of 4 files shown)


app/controllers/submissions_controller.rb
```diff
@@ -321,6 +321,12 @@ def view
 
       @displayFilename = @submission.filename
     end
+
+    mm = MimeMagic.by_magic(file)
+    if not mm.text? and not mm.subtype == "pdf"
```
victorhuangwq, line 326: I have pulled and tested this locally, and the issue was that it works fine with normal code file, but then for a completely normal Readme.txt file, MimeMagic returns nil.

I did some digging into the issue, and found out that was that MimeMagic's by magic function... is not so magical afterall. It basically does pattern matching based on https://github.com/minad/mimemagic/blob/master/lib/mimemagic/tables.rb

It did not identify the a file as a plain text because it does not match the any of the patterns they used for text

My suggestion is that we should use MimeMagic as a blacklist instead of a whitelist (to disallow "application" types) instead.

victorhuangwq: Perhaps identify using regex, since MimeMagic only provides .text? but not for other types
`if mm.type =~ /application/`

victorhuangwq:
Latest commit addresses issues brought up. Now binary files are no longer displayed.

PR #1154, opened 2020-05-15
Fixes month selection not showing on the edit assessment handin tab / Fixes travis issue with regards to migration happening after setup.

victorhuangwq:
Originally the month is hidden due to inheritance from materialize-css
image

This PR addresses it by overriding the css inherited
image


victorhuangwq:
Addresses the issue of the pending migration for travis CI. Green check marks are back again :)

PR #1155, opened 2020-05-16
Caching Global Grace Days

xinyis991105:
1. This PR adds the caching feature for global grace days left on the course dashboard. This is achieved similar to how cgdub (cumulative grace days used before) is cached in assessment_user_datum.rb. Writing to cache is done in course_user_datum.rb when the number of global grace days is requested. 
2. Deleting from cache is performed along with the update of latest_submission in assessment_user_datum.rb. I have tested locally by enabling the caching feature and putting in print statements to indicate whether the quantity is freshly calculated or not (I have commented out the print statements but please feel free to uncomment to test on your end!). It seems to be working fine when the submissions to the assessment follow the assessment dependency chain order.
3. However, suppose I have asmt2 whose previous assessment is asmt1, then submitting to asmt1 after asmt2 does not correctly update the global grace days left, even though the printing record indicates that the previous global grace days has been correctly deleted. I've spent some time testing and the conclusion is that invalidate_cgdubs_for_assessments_after is not executed after latest_submission_id has changed. I don't really understand why and am not sure if this is just a local problem on my end. This PR is definitely not ready due to this issue so I've pushed it here to ask for help. Any insight would be greatly appreciated!
(4. Once the problem is solved and everything is tested, I will clean up the code a little bit! Sorry for the comments everywhere.)

### DIFF
--- a/app/models/assessment_user_datum.rb
+++ b/app/models/assessment_user_datum.rb
@@ -143,6 +143,7 @@ class AssessmentUserDatum < ApplicationRecord
 
   # TODO
   # Refer to https://github.com/autolab/autolab-src/wiki/Caching
+  # invalidate ggl for cud as well
   def invalidate_cgdubs_for_assessments_after
     CourseUserDatum.transaction do
       # acquire lock
@@ -152,6 +153,9 @@ class AssessmentUserDatum < ApplicationRecord
       auds_for_assessments_after.each do |aud|
         Rails.cache.delete aud.cgdub_cache_key
       end
+
+      cud = CourseUserDatum.find_by(id: course_user_datum_id)
+      Rails.cache.delete cud.ggl_cache_key
     end # release lock
   end
 
--- a/app/models/assessment_user_datum.rb
+++ b/app/models/assessment_user_datum.rb
@@ -22,8 +22,6 @@ class AssessmentUserDatum < ApplicationRecord
   #   submissions associated with AUDs with Zeroed and Excused grade types aren't counted as late
   #   even if they were submitted past the due date whereas Normal grade type AUD submissions are.
   after_save :invalidate_cgdubs_for_assessments_after, if: :saved_change_to_latest_submission_id? or :saved_change_to_grade_type?
-  # after_save :invalidate_cgdubs_for_assessments_after, if: :latest_submission_id_changed?
-  # after_save :invalidate_cgdubs_for_assessments_after, if: :grade_type_changed?
 
   NORMAL = 0
   ZEROED = 1
--- a/app/models/assessment_user_datum.rb
+++ b/app/models/assessment_user_datum.rb
@@ -144,6 +144,7 @@ class AssessmentUserDatum < ApplicationRecord
   # TODO
   # Refer to https://github.com/autolab/autolab-src/wiki/Caching
   def invalidate_cgdubs_for_assessments_after
+    # puts "INVALIDATING CGDUBS"
     CourseUserDatum.transaction do
       # acquire lock
       CourseUserDatum.lock(true).find(course_user_datum_id)
@@ -153,8 +154,7 @@ class AssessmentUserDatum < ApplicationRecord
         Rails.cache.delete aud.cgdub_cache_key
       end
 
-      cud = CourseUserDatum.find_by(id: course_user_datum_id)
-      Rails.cache.delete cud.ggl_cache_key
+      Rails.cache.delete course_user_datum.ggl_cache_key
     end # release lock
   end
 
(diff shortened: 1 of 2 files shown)


app/models/course_user_datum.rb
```diff
@@ -228,6 +241,14 @@ def self.find_or_create_cud_for_course(course, uid)
       end
     end
   end
+  
+  # global grace left cache key
+  def ggl_cache_key
+    # gets it into the YYYYMMDDHHMMSS form
+    dua = course.cgdub_dependencies_updated_at.utc.to_s(:number)
+
+    "ggl/dua-#{dua}/u-#{id}"
```
fanpu, line 250: I know there's many places in the rest of the codebase that doesn't follow this, but in instance methods using self.<attribute> i.e self.id is more explicit than id, which also helps to prevent shadowing problems

app/models/assessment_user_datum.rb
```diff
@@ -68,6 +69,13 @@ def update_latest_submission
       save!
     end # release lock on AUD
     # see: http://dev.mysql.com/doc/refman/5.0/en/innodb-locking-reads.html
+
+    CourseUserDatum.transaction do
+      CourseUserDatum.lock(true).find(course_user_datum_id)
```
fanpu, line 74: Small nit but for consistency mention where it's locked as well

fanpu:
I'm able to replicate your issue with latest_submission_id, in fact I'm not even getting that update_latest_submission_id is getting called currently which is strange. Investigating...

fanpu:
My bad, seems like restarting rails caused it to be called now. Let me trace and see what's going on

fanpu:
https://stackoverflow.com/a/45493283 seems relevant

victorhuangwq:
I have also taken a look at this as well, for me the `after_save :invalidate_cgubs_for_assessments_after` was not getting fired at all, even if I removed the if statement after it.

What I did was that I replaced `after_save `with `after_commit `, and I followed the [`StackOverflow: after_commit for an attribute `](https://stackoverflow.com/questions/7128073/after-commit-for-an-attribute/14732669#14732669) to make it detect changes based on the attribute, such that it is now 

`  after_commit :invalidate_cgdubs_for_assessments_after, 
  if: 
    proc { |record| 
      record.previous_changes.key?(:latest_submission_id) &&
        record.previous_changes[:latest_submission_id].first != record.previous_changes[:latest_submission_id].last
    }
`
and `invalidate_cgubs_for_assessments_after ` is now getting triggered. However, I am not particularly sure about the difference between after_commit and after_save, and also the other side effects of the implementation in the stackoverflow, so you might have to take a deeper look into it.

victorhuangwq:
> https://stackoverflow.com/a/45493283 seems relevant

I was thinking about this as well, but even if I removed the if statement it doesn't get triggered (so it is supposed to trigger on all situations), it doesn't seem to work. Not sure why.

fanpu:
A solution is to change assesesment_user_datum:24 to

```
  after_save :invalidate_cgdubs_for_assessments_after, if: :saved_change_to_latest_submission_id? or  saved_change_to_grade_type?
```

and now the callback will be triggered. It also fixes a bug in the original code since you don't want to invalidate the cache twice. Similar to @victorhuangwq  I'm not sure why it wasn't fired if both statements are there, but the above which fixes a bug in the logic and is more desirable would work

app/models/course_user_datum.rb
xinyis991105: fixed! Thanks for the suggestion!

app/models/assessment_user_datum.rb
xinyis991105: fixed! Added comment!

xinyis991105:
I've adopted Fanpu's fix and it seems to be working properly now. It definitely helped! Also, I searched a bit on the difference between after_save and after_commit. In addition to being called after create and update, the latter seems to be called after destroy as well. But like Victor said, to minimize the possible side effect of this change, I kept the original after_save. Thank you guys for the great suggestions!
Another change I've made is to move the invalidation of ggl (global_grace_days_left) in cache from update_latest_submission to invalidate_cgdubs_for_assessments_after since 1) the latter now works properly, and 2) the CUD is locked only once instead of twice.
Currently, there exists a lock-inside-lock, one at course_user_datum.rb: 173 and another at assessment_user_datum.rb: 267. I checked https://stackoverflow.com/questions/9172934/lock-inside-lock and believe this is not a huge issue. Another reason is that locking separately twice might introduce a very unlikely race condition: the invalidation of ggl comes before caching and the user might not see the correct ggl until they submit another time. I also attempted to cache ggl in assessment_user_datum.rb: 267 but because cgdub is not only called when course dashboard is displayed, it causes unnecessary calculations.
This PR is ready to be tested. Here are the results of my local testing:
1. On the first visit of course dashboard, ggl is freshly calculated.
2. Refreshing course dashboard reads ggl from cache.
3. Visiting course dashboard again after randomly navigating to other pages reads ggl from cache.
4. Submitting on any existing assessment invalidates ggl in cache and visiting course dashboard freshly calculates ggl.
5. The numerical value of ggl is correct on each update.

PR #1162, opened 2020-05-19
Remove duplicate code block (and leaking temp dirs w/ moss); fix formatting

wallaceswsu:
The existing code which controls MOSS has a duplicate block. This PR removes the duplicate, and thus also fixes a bug in which temporary directories are leaked (directory cleanup is still a problem when MOSS hangs). Formatting is fixed.

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -364,7 +364,7 @@ file, most likely a duplicate email.  The exact error was: #{e} "
       end
     end
 
-		# Create a temporary directory
+    # Create a temporary directory
     @failures = []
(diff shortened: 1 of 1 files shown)


wallaceswsu:
Hm.  I'm not sure what's happening with the CI here. The first commit in this branch (
7de0322) just changes whitespace in a comment, but fails CI. The real change is also unexercised in the test suite. Thoughts would be appreciated.

victorhuangwq:
TravisCI ran after I merged it into master, and it passed... Will take some time to figure this out if such a situation happens again. Once again thanks for the PR.

PR #1163, opened 2020-05-19
Don't chdir in moss extraction

cg2v:
moss extraction uses absolute paths, so there's no need to
change directories, and several things break when the app's working
directory is not Rails.root

I first brought this up in #583, but it was deemed too risky. My production install has included this patch since around that time with no noticeable issues. I think the big thing that broke was multiple moss runs one after the other.

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -730,29 +730,27 @@ private
   def extract_tar_for_moss(tmp_dir, external_tar, archive)
     return unless external_tar
 
-			# Directory to hold tar ball and all individual files.
-	    extTarDir = File.join(tmp_dir, "external_input")
-	 		baseFilesDir = File.join(tmp_dir, "basefiles")
-			begin
-				Dir.mkdir(extTarDir)
-			  Dir.mkdir(baseFilesDir) # To hold all basefiles
-				Dir.chdir(baseFilesDir)
-			rescue
-			end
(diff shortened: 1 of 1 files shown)


victorhuangwq:
Hi @cg2v, may I clarify what do you mean by moss extraction using absolute paths? If it means that each path is unique, like what I think it means I think it would help if it can prevent things from breaking.

May I know what breaks when the working directory is not Rails.root?

At the same time, even if it is not related to this PR, why did it break when moss runs one after the other?

cg2v:
Absolute paths, as in /opt/autolab/tmp/useridMoss20200518-41215-159f1xo/userid_handin.c instead of just userid_handin.c

All the mkdirs, opens, copies, etc, use fully qualified absolute paths (built with Rails.root.join and File.join), so they don't need chdir in order to work. chdir alters the state of the ruby interpreter process for no benefit.

I misremembered, it wasn't multiple moss runs that broke most recently (multiple moss runs broke before #585 because they tried to reference a non-absolute "courses" directory once the current dir was no longer Rails.root)

The most recent issue was https://lists.andrew.cmu.edu/mailman/private/autolab-dev/2017-June/036844.html -- course updaters running when the most recent moss directory had been deleted and getting an exception because the current directory no longer exists.

victorhuangwq:
Okay. I will be commit this to master. 

May I clarify, so is the issue of multiple moss running breaking addressed by #585?

PR #1165, opened 2020-05-21
fixed rubocop configuration error

### LINKED ISSUE #1164: Rubocop failure
It seems autolab's codebase was configured to be analyzed by rubocop, but after the gem version upgrade, the config file `.rubocop.yml` is no longer valid.

```bash
$ rubocop
.rubocop.yml: Metrics/LineLength has the wrong namespace - should be Layout
Error: unrecognized cop Rails found in .rubocop.yml
```

wlnirvana:
Closes #1164

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -13,10 +13,9 @@ AllCops:
     - 'script/**/*'
     - 'tmp/**/*'
 
-Rails:
-  Enabled: true
+require: rubocop-rails
 
-Metrics/LineLength:
+Layout/LineLength:
   Max: 100
   # To make it possible to copy or click on URIs in the code, we allow lines
   # contaning a URI to be longer than Max.


victorhuangwq:
Requiring rubocop-rails would mean that we should add the corresponding gem into our Gemfile as well. 

At the same time, we have not maintained rubocop for quite a while, and I would discuss with the team to get it properly set up and running. We will get back to you regarding this commit.

wlnirvana:
Oh yes, I forgot to add the updated Gemfile. Sorry for the inconvenience caused.

Personally I am fine with removing rubocup as well. It's just a bit annoying when developing in IDE, which automatically detects the (outdated) rubocup config and prompts error constantly. Both updating and removing the dependency would resolve this issue.

xinyis991105:
The current master of Autolab now has a functioning rubocop linter after a series of relevant pull requests. The latest one is #1469. Thanks for bringing this up in the past!

Issue #965, opened 2018-02-28
[Feature Request] Option to Email Students Results


StevenJamesMoore:
Case:
When students hand in an assignment,
there is autograding going on, but they don't get
the results immediately... there is a lag depending
on how busy the server is. But some of the
students think that if the server accepts their work,
it's ok. So they don't look at the results until after
the assignment is due and then they see the grade
when we release it. So this would mean an email for
each submission is sent out.

Implementation:
When studens make a submission,
once it's autograded, they get an email saying
"you're submission #xxx has been autograded
and here are the results: " and we can have that
feature turned on or off (maybe on early in the
semester as they get used to autolab, and later
turn it off).

Ticket:
[INC0341219](https://cmu.service-now.com/nav_to.do?uri=incident.do%3Fsys_id=c4fa0484db241f040b4a5205dc96192e%26sysparm_stack=incident_list.do%3Fsysparm_query=active=true)

victorhuangwq:
This issue might possibly be addressed through our new loader feature which is under development. https://github.com/autolab/Autolab/pull/1167

Issue #33, opened 2014-11-10
Grade should be auto-refreshed


lxfschr:
When a student submits a file the view handin history page should automatically load the grade once it has been calculated.

devanshk:
We'll probably end up implementing this with WebSockets in Rails 5. Keeping this issue open for when we get there.

victorhuangwq:
Currently being addressed and in progress with a Websocket implementation here. 
https://github.com/autolab/Autolab/pull/1167

PR #1167, opened 2020-05-21
Feature: Autograding Waiting Screen

amartinez116:
Backend work for obtaining current job queue number from tango. Utilized Action Cable to deal with WebSockets in Rails which sends the info to _jobs_queue.html.erb a partial html file used in history.html. WebSockets send this information every 5 seconds through a 'rufus-scheduler'.


amartinez116:
Currently, this information is shown at the bottom of the history handin page. (Which will change based on front end work). The code with this information is available through app/views/jobs_queues/_jobs_queue.html.erb

(Waiting Jobs are sent to tango in decreasing order of jobID even though they are misordered here. This list is just meant to showcase the jobIDs available.) 

All this information is instantaneously refreshed every 5 seconds as can be seen through the "time created" field.

Screen Shot 2020-05-19 at 9 04 12 PM

victorhuangwq:
## Description

This PR introduces the autograding waiting screen feature into the handin history page for autograded assignments.

The main purpose of having such an autograding waiting screen is such that
- users/students do not need to refresh their page to obtained their autograded score
- users/students who are new to autolab would know that they need to wait on handin history page / come back later
- provide users/students with real time feedback on the process of the grading

This features introduces two new complexities. It uses rufus-scheduler to ping the Tango server on an interval of 5 seconds. It also uses websockets to deliver the information to the student's waiting screens.

## Motivation and Context



This enhancement is a result from our team's user interviews with professors and students who uses Autolab, who felt that
- refreshing to get score was not the ideal user experience
- new students do not know that they need to wait for their scores.

It should also address the following two issues: https://github.com/autolab/Autolab/issues/965 and https://github.com/autolab/Autolab/issues/33

## How Has This Been Tested?




This feature has been tested in the following manners on a setup that uses latest Tango 2 (w/ python3):
1. submission of hello.c to this modified hello assessment [hello_20200904.zip](https://github.com/autolab/Autolab/files/5174946/hello_20200904.zip) <-- i can only upload zips, so i zipped the tar. 
2. submission of multiple hello.c to the modified hello assessment
3. submission of hello.c, closing the handin history page (or navigating out of the handin history page) and then coming back into the handin history page.

In Tango, when I was testing, I set the `POOL_SIZE ` to be 1, so as to choke up the job queue faster, such that I can see the waiting screen's different states

**In queue, nth in line**
image
 
**In queue, next to be graded**
image

**While being autograded**
image

**Autograding complete**
image

**Banner on the assignment page if something is being autograded**
This let's people who submitted using autolab-cli to be able to know about it as well
image


## Types of changes

## Checklist:

cg2v:
So, I'm really only looking at this because of concerns about rufus-scheduler on the Carnegie Mellon deployment (we use passenger, and the pointers on the rufus-scheduler page about working with passenger do not fill me with confidence), but I'm having trouble figuring out all the moving parts.

From what I can see, the rufus-scheduler is set up when assessments_controller is loaded (which means on passenger, multiple times).
Every 5 seconds it calls AssessmentChannel.speak, which adds a row to the jobs_queues table.
The JobsQueue after_create_commit callback invokes activejob to run the JobsQueueBroadcastJob
JobsQueueBroadcastJob.perform calls ActionCable.server.broadcast to send the messages to all subscribers

What cleans up the JobsQueue entries once they are no longer needed?
What is the purpose of using ActiveJob? Why can't AssessmentChannel.speak just do the broadcast itself?
Conversely, why not use ActiveJob to replace rufus-scheduler? It doesn't have an "every n seconds" model, but jobs can requeue themselves.

victorhuangwq:
Hi @cg2v I think your concerns are valid, and I am currently looking to see how we can best change to an implementation that likely do not use rufus-scheduler

app/controllers/assessments_controller.rb
```diff
@@ -514,6 +516,45 @@ def show
...
+    waiting_jobs = []
+
+    begin
+      raw_live_jobs = TangoClient.jobs
+    rescue TangoClient::TangoException => e
+      Rails.logger.debug "Error while getting job list: #{e.message}"
+    end
+
+    # Build formatted lists of the running, waiting
+    unless raw_live_jobs.nil? || raw_live_jobs.empty?
```
victorhuangwq, line 546: @michellexliu

damianhxy:
Superseded by #1645

Issue #1173, opened 2020-05-22
Late days used is not showing up on handin history page


victorhuangwq:
![image](https://user-images.githubusercontent.com/5773562/82682811-6dcc2b80-9c82-11ea-9d79-96be4bf3da22.png)

PR #1178, opened 2020-05-29
Handin history late days used

xinyis991105:
This PR addresses the issue in #1173. The issue was probably due to a lack of content in between the td tags in _submission_history_row.html.erb and _submission_summary_row.html.erb so I've added the information back in. In addition, the Internet seems to be mixed about the use of nobr tag in HTML (https://stackoverflow.com/questions/28532076/why-is-the-nobr-deprecated). I decided to keep it the way it is in HTML but we also have the option of redefining it in style.css.scss. Does anybody have any strong opinion about this? I tested by randomly navigating to handin history page and assessment index page. It seems to yield the content fine. Further check is needed.

### DIFF
--- a/app/assets/stylesheets/style.css.scss
+++ b/app/assets/stylesheets/style.css.scss
@@ -1185,3 +1185,6 @@ div.center {
   padding: 0 0 5px 15px;
 }
 
+.nobr {
+  white-space: nowrap;
+}
--- a/app/views/assessments/_submission_history_row.html.erb
+++ b/app/views/assessments/_submission_history_row.html.erb
@@ -43,7 +43,9 @@
 
   <% if @course.grace_days >= 0 then %>
     <td title="Submitted <%= submission.days_late %> days late" >
-      </nobr>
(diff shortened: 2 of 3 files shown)


victorhuangwq:
I don't think I have any issue with using nobr. But according to the stackoverflow you have just sent, it does seem like a good idea to add it into the CSS.

app/assets/stylesheets/style.css.scss
```diff
@@ -1183,5 +1183,4 @@ div.center {
 
 .collapsible-menu-link > .collapsible-header {
   padding: 0 0 5px 15px;
-}
-
+}
```
victorhuangwq, line 1186: Adding the nobr into the css would be good.

xinyis991105:
@victorhuangwq Hey Victor! I've added the CSS back. Can you do a final sanity check?

victorhuangwq:
Looks good to me

xinyis991105:
I think you need to approve one more time. Thanks!

PR #1185, opened 2020-06-08
Provide a 'semantic' presentation of assignment feedback via JSON.

alamijal:
This commit provides optional output formatting via a JSON string.
One presentation is currently supported, but the encoding allows for
future extensibility. An example assessment (hellocat) demonstrates
potential use. Implementation details are outlined in the README under examples/hellocat.
hellocatScreenshot


oliverli:
Alright, build is passing on Travis after merge - I suspect it was in a commit that updated `.travis.yml`. Thanks once again for the great PR!

Issue #1169, opened 2020-05-22
Regrade button not working


wlnirvana:
**Please note that this issue does not occur to the public demo site, which seems using an old version (prior to [this commit](https://github.com/autolab/Autolab/commit/82972aa0b69dae32eca07d2971bd314998bca295)) of Autolab shadowing this issue with another bug.**

For autograded assignment, the regrade button on `your.autolab.domain/courses/AutoPopulated/assessments/hello/submissions` page sends http get request instead post request, causing a routing error.

This might be a regression introduced in #845 and further #856

wlnirvana:
Autolab relies on jquery-ujs to support http post request based on`data-method` attribute of `<a>` elements, which [internally uses](https://github.com/rails/jquery-ujs/blob/9e805c90c8cfc57b39967052e1e9013ccb318cf8/src/rails.js#L430) jQuery's `on` method to attach handlers to `<a>` listening for clicks.

Quoting from [jQuery doc](https://api.jquery.com/on/):

> When a selector is provided, the event handler is referred to as delegated. The handler is not called when the event occurs directly on the bound element, but only for descendants (inner elements) that match the selector. jQuery bubbles the event from the event target up to the element where the handler is attached (i.e., innermost to outermost element) and runs the handler for any elements along that path matching the selector.

To summarize, the jquery-ujs handler is not directly attached to `<a>` but to the document and then delegated to `<a>`. However, changes in #845 will prevent the event from bubbling up.

victorhuangwq:
Hi @wlnirvana I am currently doing a sweep of all the issues we are having, and there was actually an issue a while back that mentioned the same issue, and it comes with a patch at https://github.com/autolab/Autolab/issues/936. 

I am looking into it right now.

victorhuangwq:
It should be resolved by the linked PR.

Issue #1168, opened 2020-05-22
Checkbox hidden


wlnirvana:
Not sure if checkboxes on other pages are intentionally hidden, but the first column of the table on page http://autolab.ml/courses/AutoPopulated/assessments/lab0/submissions is invisible checkboxes. Sort of inconvenient.

Possible root cause: https://github.com/mkhairi/materialize-sass/blob/ebb0543d72fee08053a7df2d24fd4c1f15542fb2/assets/stylesheets/materialize/components/forms/_checkboxes.scss#L8

wlnirvana:
According to their doc/demo https://materializecss.com/checkboxes.html it seems a `<span>` is needed after the `<input>`.

victorhuangwq:
Hi @wlnirvana. Thank you for pointing this out. This is an issue caused by our materialise upgrade, and we tried to fix all of the checkbox, but it seems like we have missed this one. We know how to address this so leave this to us.

victorhuangwq:
Hi just a quick update, fixing the checkbox is easy, but it doesn't seem as if the checkbox is serving any purpose. Will have to take a look closer again.

PR #1193, opened 2020-06-21
Fix Manage Subs Checkbox

victorhuangwq:
## Description

Adds checkbox back into the manage submissions page

## Motivation and Context

Addresses bug brought up in https://github.com/autolab/Autolab/issues/1168
as well as https://github.com/autolab/Autolab/issues/1169

## How Has This Been Tested?

Renders on Chrome browser. The main issue was that jquery was not imported.

## Screenshots
image
image

## Types of changes

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

### DIFF
--- a/app/views/submissions/index.html.erb
+++ b/app/views/submissions/index.html.erb
@@ -92,9 +92,12 @@
   <% for submission in @submissions %>
     <tr id="row-<%= submission.id %>" class="submission-row">
       <td>
-        <label>
-        <input class="cbox" type="checkbox" id="cbox-<%= submission.id %>">
-        </label>
+        <p>
+          <label>
+            <input class="cbox" type="checkbox" id="cbox-<%= submission.id %>">
+            <span/>
+          </label>
+        </p>
       </td>
       <td><%= "#{submission.course_user_datum.last_name}, #{submission.course_user_datum.first_name}" %>
       <%= link_to submission.course_user_datum.email,
--- a/app/views/submissions/index.html.erb
+++ b/app/views/submissions/index.html.erb
@@ -16,6 +16,7 @@
 <% end %>
 
 <% content_for :javascripts do %>
+  <%= javascript_include_tag 'application' %>
   <%= external_javascript_include_tag "lodash", "3.3.1" %>
   <%= javascript_include_tag "sorttable" %>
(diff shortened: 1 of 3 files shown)


victorhuangwq:
I have made some changes, to import jquery to actually make the checkbox functional again.

wlnirvana:
Just a reminder, rails 5 is dropping jquery-ujs in favor of rails-ujs. In fact, with the advent of ES6 and mature frameworks, the popularity of jquery itself seems declining as well.

cg2v:
jquery is imported by app/views/layouts/application.html.erb, so does not need to be imported again the way things are now. The reason for that seems to be whatever `content_for :js_widget_patch` was used for.

js_widget_patch is no longer defined, so it's probably safe to move jquery into application.js as you have done, as long as jquery is also removed from the layout.

app/views/submissions/index.html.erb
```diff
@@ -16,6 +16,7 @@
 <% end %>
 
 <% content_for :javascripts do %>
+  <%= javascript_include_tag 'application' %>
```
cg2v, line 19: This is not needed. The layout imports application.js at the beginning of the `<body>` tag

victorhuangwq: Issue addressed

app/views/submissions/index.html.erb
```diff
@@ -123,11 +123,12 @@
       </td>
       <td class="exclude-click" style="text-align:center;">
         <% if @autograded and submission.version > 0 then %>
-          <%= link_to "<i class='material-icons'>autorenew</i>".html_safe,
-              [:regrade, @course, @assessment, submission_id: submission.id],
-              { method: :post,
+          <%= button_to [:regrade, @course, @assessment, submission_id: submission.id],
```
victorhuangwq, line 126: realized that link_to still does not address the issue (where it uses get instead of post).
Changed to button_to, which is meant for post interactions in the first place.  (in fact it uses post method by default, but I just left the method specification here anyways)

victorhuangwq:
The issues brought up should have been adequately addressed. @wlnirvana I do see that rails have moved away from jquery mostly, so we will KIV changing into rails ujs as well.

PR #1210, opened 2020-08-27
Hotfix for gradebook not displaying

victorhuangwq:
## Description

This change updates the jquery-ui used, and also removed the leftover fix meant for bootstrap

## Motivation and Context

As of the current master, the gradebook is not displaying, as per a feedback from CMU instructor:
With the recent update to Autolab, the gradebook does not display. If I export, I can see the student grades, but I cannot view or search for a single student, which seems to be a defect.

## How Has This Been Tested?

Tested as both an instructor and course assistant 

Before:
image

After:
image

## Types of changes

### DIFF
--- a/app/views/gradebooks/view.html.erb
+++ b/app/views/gradebooks/view.html.erb
@@ -60,7 +60,6 @@
 
 
 <% content_for :js_widget_patch do %>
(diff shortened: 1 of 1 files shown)


fanpu (approved):
Tested that it works on my setup and allows the jquery-ui library to load properly now

PR #1214, opened 2020-09-17
Secret generation without rake

Huahan98:
## Description

This is for the one-click setup script. Since the host machine might not have bundle, we cannot use `bundle exec rake secret` to generate a random secret key. Hence a new script that uses /dev/urandom to generate a random key is added to replace the old method. 

## Types of changes

## Checklist:

### DIFF
--- a/bin/setup.sh
+++ b/bin/setup.sh
@@ -190,7 +190,7 @@ autolab_setup() {
     cp $AUTOLAB_PATH/config/school.yml.template $AUTOLAB_PATH/config/school.yml
 
     cp $AUTOLAB_PATH/config/initializers/devise.rb.template $AUTOLAB_PATH/config/initializers/devise.rb
-    sed -i "s/<YOUR-SECRET-KEY>/`bundle exec rake secret`/g" $AUTOLAB_PATH/config/initializers/devise.rb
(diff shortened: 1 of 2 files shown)


bin/setup.sh
```diff
@@ -190,7 +190,7 @@ autolab_setup() {
     cp $AUTOLAB_PATH/config/school.yml.template $AUTOLAB_PATH/config/school.yml
 
     cp $AUTOLAB_PATH/config/initializers/devise.rb.template $AUTOLAB_PATH/config/initializers/devise.rb
-    sed -i "s/<YOUR-SECRET-KEY>/`bundle exec rake secret`/g" $AUTOLAB_PATH/config/initializers/devise.rb
+    sed -i "s/<YOUR-SECRET-KEY>/`sh generate_secret.sh 128`/g" $AUTOLAB_PATH/config/initializers/devise.rb
```
Huahan98, line 193: FYI, on Mac I think the proper command is `sed -i ".bak" "s/<YOUR-SECRET-KEY>/`bash generate_secret.sh 128`/g" target/file`, i.e. we must provide a suffix (".bak") when running `sed -i`. 

I ran into this issue while following the installation guide: https://autolab.github.io/docs/

PR #1215, opened 2020-09-17
Improve accessibility of Options button

ashleyzhang:
## Description
This PR makes the Options dropdown on the student Assignment Handin page open by default. This way, the user does not have to click into the Options button to download their handout. It also fixes a minor issue on the Instructor view where the Admin Options menu would replay the dropdown animation upon clicking the header for the first time.

## Motivation and Context
Perkins Accessibility Report recommended that we make it more clear that the Options section can be expanded.

## How Has This Been Tested?
If you navigate to any assignment, there the Options menu should start in an open state, and you'll still be able to close it by clicking on the Options header.

Screen Shot 2020-09-17 at 5 26 45 PM

## Types of changes

## Checklist:


mtoupsUNO (dismissed):
this works for me, thanks!

xinyis991105 (commented):
I tested by acting as both instructor and student. The collapsible displays correctly in both cases. But when I tab through the page, it seems that it will first focus on the whole collapsible and then on the toggle button. When the focus is on the collapsible, hitting enter does nothing. When it is on the toggle button, it toggles correctly. I tried removing the tabindex attribute from the collapsible and tabbing skips the collapsible focus. Can you look into that and see if we need to fix it?

ashleyzhang:
Thanks for catching that! Made a change that should fix it.

xinyis991105 (approved):
LGTM!

PR #1224, opened 2020-10-15
Adds alert component

victorhuangwq:
## Description

A simple flash component based on the existing flash styling, that can be easily rendered using rails render function.

## Motivation and Context

This simple PR is an added as example code for components under the rails framework, at the same time it also helps make the flash component more standardized in implementation. This implementation follows the description as described here closely: https://goiabada.blog/rails-components-faedd412ce19

## How Has This Been Tested?

Component tested to render with various inputs,
- with message
- no message
- with types
- no types

image

## Types of changes


app/views/components/_alert.html.erb
```diff
@@ -0,0 +1,19 @@
...
+    Example usage: (the % are removed from the < and >)
+    < render "components/alert", type:"success" do >
+        This is a success alert with <a href="#" class="alert-link">an example link</a>.
+    < end >
+
+    It has the following types:
+    alert,success,notice
+%>
+
+<% message = yield.presence || local_assigns.fetch(:message) %>
```
damianhxy, line 13: A default message should be passed to `fetch` else we get an error when there is no message.

app/views/components/_alert.html.erb
```diff
@@ -0,0 +1,19 @@
...
+
+<% type = local_assigns.fetch(:type,:notice) %>
+
+<%= content_tag :div, id: "flash_#{type}", class: "#{type}" do %>
+    <%= message %>
```
damianhxy, line 18: Messages obtained via `yield` are marked `html_safe` and thus not escaped by Rails, leading to potential XSS.

Need to figure out how to unset/bypass the `html_safe` flag and escape all messages by default. Thereafter, add a check for `flash[:html_safe]` to conditionally not escape messages.

damianhxy (commented):
Tested that functionality works, however there is a potential XSS which needs to be addressed.

PR #1228, opened 2020-10-22
Make annotation description clicks use AJAX calls to change files and change header position in URL bar

oliverli:
## Description

Add `url` property to `localCache` when running `purgeCurrentPageCache` (which is run on initial page load, after adding/editing an annotation). This ensures that the address in the URL bar (i.e. `headerPosition`) changes when the file is changed using the local cache, so that refreshes will bring the user back to the same file.

Made the annotation descriptions on the right change files using an AJAX call (and stores result in local cache) instead of refreshing the file altogether which wipes local cache.

Updated `fillAnnotationBox` to reflect new structure of annotation descriptions.

Also updated pull request template to reflect new structure where docs are in the same repo.

## Motivation and Context

The first file that is initially loaded (e.g. header position 0) is not properly stored in the local cache. Going from file 0 to file 1 and back to file 0 using the left nav bar results in the URL bar showing file 1.

Changing files using the annotation description on the right results in a fresh page load. Local cache is wiped, which is not ideal. We have the mechanism to change files using AJAX, so we should do that.

## How Has This Been Tested?

Tested by clicking the links (you can view result using the network tab in developer tools and viewing `localCache` in console). Also tested consistency after adding and removing annotations.

## Types of changes

## Checklist:


oliverli:
Comment: someone should probably also test annotations in PDF, it _may_ have broke (or has long being broken), I'm not sure yet

xinyis991105 (approved):
Tested by uploading a tar of three files,
- randomly switched between the different files using the file tree on the left; url was updated accordingly, especially when going back to the first file from other file
- added annotations, switched between files as above, url header position ok
- with the added annotations, switched between files using the links from the annotation box, cache was called (console) and page is no longer refreshed every time
- deleted annotations and switched between files using a hybrid of the left and right links on the fly, url header position was ok
- Went through code for one last check
- LGTM!

P.S. I tried uploading a pdf file and annotated from there. It seems to be working fine functionality-wise but the cache behavior doesn't seem to follow. Is this expected?

oliverli:
Yep, PDFs don't seem to be cached at all by design, so that's expected behavior. Thanks for the PR review!

PR #1234, opened 2020-10-29
API endpoint condition retrieval

xinyis991105:
## Description
1. Added API endpoint for retrieving current risk conditions in app/controllers/metrics_controller.rb.
2. Changed the name of the column "type" for the risk condition table since it interferes with Rails subclass and inheritance management. Updated app/models/risk_condition.rb to reflect that change.

## Motivation and Context
The API endpoint offers a neat way for front end to retrieve risk condition information for a course.

## How Has This Been Tested?
The API endpoint in metrics_controller.rb simply calls the public method RiskCondition.get_current_for_course so it suffices to test the correctness of RiskCondition.get_current_for_course.
1. Open up Rails console.
2. Use RiskCondition.create_condition_for_course_with_type(course_id, type, params) to generate some risk conditions for courses. When adding these conditions, call RiskCondition.get_current_for_course(course_id) to verify consistency.
3. Randomly destroy some RiskCondition models and call RiskCondition.get_current_for_course(course_id) again to verify consistency.
4. Verify that each time RiskCondition.get_current_for_course(course_id) is called, there are no more than four models returned, each of a distinct type.

## Types of changes

## Checklist:

### DIFF
--- a/app/controllers/metrics_controller.rb
+++ b/app/controllers/metrics_controller.rb
@@ -6,15 +6,15 @@ class MetricsController < ApplicationController
 	action_auth_level :get_current_metrics, :instructor
 	def get_current_metrics
 		# This API endpoint aims to retrieve the current/latest risk conditions for a particular course
-		# User is expected to include in the parameters the course's ID with :id being the key
 		# On success, a JSON list of condition objects will be returned
 		# The type of each object is specified in a field called "condition_type"
-		# Possible types include: GRACE_DAY_USAGE, GRADE_DROP, NO_SUBMISSIONS, LOW_GRADES
-		# Other fields for the risk condition object include: parameters, version, created_at, updated_at, and course_id
-		# On error, a flash error message will be shown and nothing gets returned (i.e. nil gets returned)
+		# Possible types include: grace_day_usage, grade_drop, no_submission, low_grades
+		# Other fields for a risk condition object include parameters, version, created_at, updated_at, and course_id
+		# In particular, the parameters field includes specific information of the condition corresponding to its type
+		# On error, a flash error message will be shown and nil gets returned
 		begin
-			course_id = params[:id]
-			conditions = RiskCondition.get_current_for_course(course_id)
+			course_name = params[:course_name]
+			conditions = RiskCondition.get_current_for_course(course_name)
 			render json: conditions
 		rescue => error
 			flash[:error] = error.message
--- a/app/controllers/metrics_controller.rb
+++ b/app/controllers/metrics_controller.rb
@@ -14,28 +14,10 @@ class MetricsController < ApplicationController
 		begin
 			course_id = params[:id]
 			conditions = RiskCondition.get_current_for_course(course_id)
-			conditions_json_string = conditions.to_json
-			conditions_json = JSON.parse(conditions_json_string)
-			conditions_json.map! { |c| c["condition_type"] = type_string_of_enum(c["condition_type"]); c }
-			render json: conditions_json
+			render json: conditions
 		rescue => error
 			flash[:error] = error.message
 			return
-		end
-	end
-
-private
-	def type_string_of_enum(t)
-		if t == 1
-			return "GRACE_DAY_USAGE"
-		elsif t == 2
-			return "GRADE_DROP"
-		elsif t == 3
-			return "NO_SUBMISSIONS"
-		elsif t == 4
-			return "LOW_GRADES"
-		else
-			raise "Condition retrieval failed! Got invalid condition type for metric conditions!"
 		end
 	end
 end
--- a/app/controllers/metrics_controller.rb
+++ b/app/controllers/metrics_controller.rb
@@ -3,6 +3,7 @@ class MetricsController < ApplicationController
 	def index
 	end
 
+	action_auth_level :get_current_metrics, :instructor
 	def get_current_metrics
 		# This API endpoint aims to retrieve the current/latest risk conditions for a particular course
 		# User is expected to include in the parameters the course's ID with :id being the key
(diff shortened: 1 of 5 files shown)


victorhuangwq:
I have tested the this feature using the Rails console, and I did the following to check if the function works

- RiskCondition.create_condition_for_course_with_type to add a new condition
- RiskCondition.get_current_for_course, to see if the condition has been added
- RiskCondition.create_condition_for_course_with_type to add another condition
- RiskCondition.get_current_for_course, to see if the conditions have been updated.

So the functions by themselves work.

app/controllers/metrics_controller.rb
```diff
@@ -2,4 +2,14 @@ class MetricsController < ApplicationController
 	action_auth_level :index, :instructor
 	def index
 	end
+
+private
```
victorhuangwq, line 6: Because this API supposed exposed to the public (It is a public API, but with authentication, which should be handled by the action auth level)

app/controllers/metrics_controller.rb
```diff
@@ -2,4 +2,14 @@ class MetricsController < ApplicationController
...
+	def get_current_metrics(course_id)
+		# returned as a hash, with each type's integer index as a key
+		# GRACE_DAY_USAGE = 1 # :grace_day_threshold, :date
+		# GRADE_DROP = 2 # :percentage_drop, :consecutive_counts
+		# NO_SUBMISSIONS = 3 # :no_submissions_threshold
+		# LOW_GRADES = 4 # :grade_threshold, :count_threshold
+		return RiskCondition.get_current_for_course(course_id)
```
victorhuangwq, line 13: the API shouldn't be returning enums of 1, 2, 3, 4, because the user of the API might not necessarily know the exact implementation. I suggest that it be changed to a string value. 

At the same time, I don't think you can return hash You might need to convert it to json to return. 
```
class UsersController < ApplicationController
    def show
        @user = User.find(params[:id])
        render json: @user
    end
end
```
At the same time, it should return appropriate error messages as well.

victorhuangwq (changes requested):
The current get_current_metrics function is structured as a private function on the controller page, but we actually require it to be a public API that we can call.

app/controllers/metrics_controller.rb
xinyis991105: Changes pushed. Can you take another look at it? Thanks!

xinyis991105:
Screen Shot 2020-10-30 at 8 45 45 PM
Here's how the endpoint looks as a JSON response. Testing procedure is the same in terms of adding and deleting risk conditions to check for consistency. Only this time, when creating a new risk condition, instead of specifying an integer for a type, one should use the enum symbol (see the updated app/models/risk_condition.rb). The endpoint has also been updated to use :course_name in params as an identifier so there is no need to pass in anything any longer.

victorhuangwq:
I have proceeded to test the updated version

- Added a new RiskCondition for a course
- Go to metrics/get_current_metrics endpoint. 
- Add a new RiskCondition for a course
- Go to metrics/get_current_metrics endpoint, and I see the metrics being updated.

And I can see the changes being reflected on the endpoint itself.

app/controllers/metrics_controller.rb
```diff
@@ -2,4 +2,23 @@ class MetricsController < ApplicationController
 	action_auth_level :index, :instructor
 	def index
 	end
+
+	action_auth_level :get_current_metrics, :instructor
+	def get_current_metrics
```
victorhuangwq, line 7: this finally works, and I tried accessing the endpoint without logging in and it redirects me to the homepage as well

app/models/risk_condition.rb
```diff
@@ -1,35 +1,50 @@
...
+    options = { course_id: course_id, condition_type: type, parameters: params, version: version }
     newRiskCondition = RiskCondition.new(options)
     if not newRiskCondition.save
       raise "Fail to create new risk condition with type #{type} for course #{course_id}"
     end
   end
 
+  def self.get_current_for_course(course_name)
+    conditions = []
+    course_id = Course.find_by(name: course_name).id
```
victorhuangwq, line 33: that is surprisingly simple a fix!

app/models/risk_condition.rb
```diff
@@ -1,35 +1,50 @@
...
+  # :no_submissions => :no_submissions_threshold
+  # :low_grades => :grade_threshold, :count_threshold
 
   def self.create_condition_for_course_with_type(course_id, type, params)
-  	# parameter check shouldn't surface to user and is for debug only
-    if (type == GRACE_DAY_USAGE && (params[:grace_day_threshold].nil? || params[:date].nil? || params.length != 2)) ||
-       (type == GRADE_DROP && (params[:percentage_drop].nil? || params[:consecutive_counts].nil? || params.length != 2)) ||
-       (type == NO_SUBMISSIONS && (params[:no_submissions_threshold].nil? || params.length != 1)) ||
-       (type == LOW_GRADES && (params[:grade_threshold].nil? || params[:count_threshold].nil? || params.length != 2))
+  	# parameter check shouldn't surface to user and is for sanity check during development only
```
victorhuangwq, line 15: I was adding new condition using this function, but I realize there might be some issue here?
Because when the user updates the condition, they update all four conditions in one go, not one by one. And they might deselect (remove) some conditions as well.

For example:
**First time saving**: The user sets grace_days_usage, and low grades.
**Second time saving**: The user sets no_submissions, and low grades, and unselects grace_days_usage.
So in this case I forsee grace_days_usage being stuck at v1? 

Since I think this is to be addressed in your separate ticket, and I will proceed to approve the PR. That way you can address this in your next PR instead. (Maybe it is how you intend this to be? I'm not sure)

victorhuangwq (approved):
There might be an issue, but I think it would be more relevant to address it in the update risk conditions endpoint instead.

PR #1235, opened 2020-10-29
Basic frontend for Watchlist tabs

amartinez116:
Creates the basic frontend functionality for the watchlist, which goes into the "Watchlist" tab of the metrics feature.

## Description

Whether student metrics are defined is currently hard coded by the variable @define_metrics. This will be change when we integrate frontend and backend changes.

image

With @define_metrics = true, we get the following: 
image

## How Has This Been Tested?

I tested this locally by navigating to the metrics page.

## Types of changes

## Checklist:


app/views/metrics/index.html.erb
```diff
@@ -6,3 +6,61 @@
...
+        <a class="item active" data-tab="new_tab">New</a>
+        <a class="item" data-tab="contacted_tab">Contacted</a>
+        <a class="item" data-tab="resolved_tab">Resolved</a>
+        <a class="item" data-tab="archived_tab">Archived</a>
+      </div>
+    </div>
+    <div class="thirteen wide column">
+      <% if @define_metrics then %>
+        <div class="ui tab segments active" data-tab="new_tab">
+          <%= render partial: 'student_row', locals: {student_name: "Jane Doe", grace_days: 3, low_scores: 2} %>
```
victorhuangwq, line 40: I think the student row looks great, it works like a component and is separate from the main watchlist, but I am concerned about using render partial in rendering the rows themselves. Because render partial is carried out on the server side, not the client side, as discussed previously in our hacking session.  This is an issue as it would mean that a refresh of the page is needed whenever an entry is refreshed / resolved / contacted

victorhuangwq: would be good to collaborate with @ashleyzhang to resolve this issue

amartinez116: This will be handled by @ashleyzhang  PR, this PR is for simplified frontend design.

victorhuangwq (approved):
Navigated to the metrics page, and the tabs and the left tabs are rendered accordingly. At the same time it is window responsive as well.

As the watchlist entries will be done by @ashleyzhang, I did not inspect it as much.
A small nit is that the empty inbox icon is too large, would prefer it to be around 2/3 of it's current size.

PR #1236, opened 2020-11-03
Update rake to rails to follow Rails 5 syntax

fanpu:
## Description
For Rails 5, `rails` should now be used instead of `rake`

See https://weblog.rubyonrails.org/releases/:
> One Rails command to rule them all
> Why are some commands living in bin/rails and some commands living in bin/rake? That’s a common question, especially for beginners, and we never had a good answer (just lots of technical excuses). So now we’ve committed to making bin/rails the one master command to rule them all. All your rake commands are available through here as a gateway, but we’ll eventually port many of them over. So your fingers will now have to get used to bin/rails db:migrate instead of bin/rake db:migrate. That should only take a few months!

## Types of changes

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -53,7 +53,7 @@ RUN chown -R app:app .
 USER app
 
 # precompile the Rails assets
-RUN RAILS_ENV=production bundle exec rake assets:precompile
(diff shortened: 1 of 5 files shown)


oliverli (approved):
LGTM

PR #1241, opened 2020-11-10
API endpoint for updating metrics condition

xinyis991105:
## Description
This PR adds the API endpoint for updating metrics conditions for the metrics feature.

## Motivation and Context
When instructor submits the form to update their current risk metrics conditions, they need a way for the backend to update the database information accordingly. This PR does exactly that.

## How Has This Been Tested?
1. Due to Autolab's level of security, to test this PR with the fetch API in console, you need to disable protect_from_forgery temporarily. This is located at app/controllers/application_controller.rb. Comment out the relevant lines as below. Remember to uncomment them back once you are done with testing. (Let me know if this is not the legit way to get around the problem of authentication.)
Screen Shot 2020-11-10 at 4 46 03 PM
2. The update_current_for_course method in app/models/risk_condition.rb has some commented-out debug print statements. Feel free to uncomment them during testing. Request me to remove these once you think this PR is ready to merge.
Screen Shot 2020-11-10 at 4 51 12 PM

3. Start your rails server. Go to 'courses/AutoPopulated/metrics'. Open up your console. The parameter requirement has been stated in the documentation of the API endpoint in app/controllers/metrics_controller.rb. Use fetch API to simulate POST requests with url 'courses/AutoPopulated/metrics/update_current_metrics' and offer a body according to the parameter criteria. For example,
Screen Shot 2020-11-10 at 4 56 54 PM
Check back on your terminal. If you uncommented the debug statements, you should see relevant information printed out.
Screen Shot 2020-11-10 at 4 59 14 PM
Otherwise, you can also go to 'courses/AutoPopulated/metrics/get_current_metrics' using the previously defined API to verify the latest risk conditions.
4. Start from no risk conditions defined at all. Perform the following basic test:
- Send a request with no parameter; no change is made to database and get_current_metrics returns nothing.
- Send a request with some parameter of your choice; database is updated to reflect that change and get_current_metrics returns the conditions consistent with your parameters.
- Send a request with the same parameter as above; no change is made to database; get_current_metrics returns the same conditions as before.
- Send a request with some different non-empty parameters; change is made to database; get_current_metrics returns conditions consistent with the new parameters.
- Send a request with an empty parameter; change is made to database; get_current_metrics returns nothing.
- Send a request with an empty parameter again; no change; get_current_metrics returns nothing.
- Send a request with non-empty parameter; change; get_current_metrics returns conditions consistent with your parameters.
5. Use fetch API to either break the parameter format or do another cascade of test similar to the one above. When format breaks, no change is made to database and refreshing 'courses/AutoPopulated/metrics' should display the error in the flash message.

## Types of changes

## Checklist:

1. What should the POST API return at the end? Should it still be render json:?
2. Because the error cases are more complicated and varied for this API call, how should I surface these errors? In addition, most of the errors will be for the sake of debugging when developing, do we need to consider how to surface those to potential users?

### DIFF
--- a/app/controllers/metrics_controller.rb
+++ b/app/controllers/metrics_controller.rb
@@ -48,23 +48,35 @@ class MetricsController < ApplicationController
 		# The request parameters must take on the following format:
 		# params[type_a] is nil if instructor does not want to include a risk condition of type_a during update
 		# params[type_a] is an object satisfying the parameter requirement for type_a condition
-		# e.g. params[1] should have form { :grace_day_threshold => 3, :date => "2020-03-05" }
+		# e.g. params["grace_day_usage"] should have form { "grace_day_threshold" => 3, "date" => "2020-03-05" }
 		# Currently, four types of risk conditions are supported:
-		# 1 for :grace_day_usage with parameters :grace_day_threshold, :date
-		# 2 for :grade_drop with parameters :percentage_drop, :consecutive_counts
-		# 3 for :no_submission with parameter :no_submissions_threshold
-		# 4 for :low_grades with parameters :grade_threshold, :count_threshold
-		# 0 is reserved for when a course assistant deselects all conditions; user of this endpoint doesn't need to account for this
-		# On error, a flash error message will be shown and nil gets returned
+		# "grace_day_usage" with parameters "grace_day_threshold", "date"
+		# "grade_drop" with parameters "percentage_drop", "consecutive_counts"
+		# "no_submissions" with parameter "no_submissions_threshold"
+		# "low_grades" with parameters "grade_threshold", "count_threshold"
+		# On error, a flash error message will be rendered and nil gets returned
 		begin
 			course_name = params[:course_name]
-			# TODO: need to further investigate what form data passed in takes on
-			conditions = RiskCondition.update_current_for_course(course_name, params.except(:course_name))
-			# TODO: what exactly does render do?
+			params_filtered = new_metrics_params
+			if params_filtered.nil?
+				params_filtered = {}
+			else
+				params_filtered = params_filtered.to_h
+			end
+			conditions = RiskCondition.update_current_for_course(course_name, params_filtered)
 			render json: conditions
 		rescue => error
-			flash[:error] => error.message
-			return
+			flash[:error] = error.message
+			render json: {}
 		end
+	end
+
+private
+	
+	def new_metrics_params
+		return params.require(:metric).permit(:grace_day_usage => [:grace_day_threshold, :date],
+											  :grade_drop => [:percentage_drop, :consecutive_counts],
+											  :no_submissions => [:no_submissions_threshold],
+											  :low_grades => [:grade_threshold, :count_threshold]) if params[:metric].present?
 	end
 end
--- a/app/controllers/metrics_controller.rb
+++ b/app/controllers/metrics_controller.rb
@@ -17,7 +17,7 @@ class MetricsController < ApplicationController
 			conditions = RiskCondition.get_current_for_course(course_name)
 			render json: conditions
 		rescue => error
-			flash[:error] = error.message
+			render :text => 'Not Found', :status => '404'
 			return
 		end
 	end
@@ -63,11 +63,14 @@ class MetricsController < ApplicationController
 			else
 				params_filtered = params_filtered.to_h
 			end
+			if params_filtered != params[:metric]
+				raise "Invalid update parameters for risk conditions! Make sure your request body fits the criteria!"
+			end
 			conditions = RiskCondition.update_current_for_course(course_name, params_filtered)
 			render json: conditions
 		rescue => error
-			flash[:error] = error.message
-			render json: {}
+			render json: { error: error.message }, status: :bad_request
+			return
 		end
 	end
 
--- a/app/controllers/metrics_controller.rb
+++ b/app/controllers/metrics_controller.rb
@@ -24,7 +24,18 @@ class MetricsController < ApplicationController
 
 	action_auth_level :update_current_metrics, :instructor
 	def update_current_metrics
-		# The code below assumes the post data is consistent with the type/parameter relation as is defined in risk_condition.rb
+		# This API endpoint aims to update current/latest risk conditions for a particular course
+		# On success, a JSON list of condition objects that are freshly created out of the request parameters will be returned
+		# The request parameters must take on the following format:
+		# params[type_a] is nil if instructor does not want to include a risk condition of type_a during update
+		# params[type_a] is an object satisfying the parameter requirement for type_a condition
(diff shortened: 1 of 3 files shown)


victorhuangwq:
I have tested the API in the following manner:

Started with some risk metrics inside.
1. Send a request with no parameter; Calling get_current_metrics returns nothing. 
1. Send a request with some parameter of ; get_current_metrics returns the conditions that are consistent
1. Send a request with some parameter of ; get_current_metrics returns the new conditions that are consistent
1. Send a request with the same parameter as above; no change is made to database; get_current_metrics returns the same conditions as before.

However, sending requests with malformed body does not throw an error:

Example: 
`{"no_submissions":{"no_submissions_t3hreshold":32}}` cause nothing to change, but there is no error message
and 
`{"no_submi312ssions":{"no_submissions_threshold":4}}` is equivalent to sending an empty parameter and removes all the conditions.

Suggestion:
- Only remove conditions on empty parameter
- Check fields at each level of the json.

app/controllers/metrics_controller.rb
```diff
@@ -41,4 +41,42 @@ def get_watchlist_instances
...
+			course_name = params[:course_name]
+			params_filtered = new_metrics_params
+			if params_filtered.nil?
+				params_filtered = {}
+			else
+				params_filtered = params_filtered.to_h
+			end
+			conditions = RiskCondition.update_current_for_course(course_name, params_filtered)
+			render json: conditions
+		rescue => error
```
victorhuangwq, line 68: Do not flash the error message, because it is not rendering the page, so the flashing of the error cannot be shown, instead return an appropriate status instead.

```
rescue => error
	render :text => 'Not Found', :status => '404'
	return
end
```

refer to here for the status codes
https://gist.github.com/mlanett/a31c340b132ddefa9cca

app/models/risk_condition.rb
```diff
@@ -11,44 +11,107 @@ class RiskCondition < ApplicationRecord
...
   # :low_grades => :grade_threshold, :count_threshold
 
-  def self.create_condition_for_course_with_type(course_id, type, params)
+  def self.create_condition_for_course_with_type(course_id, type, params, version)
   	# parameter check shouldn't surface to user and is for sanity check during development only
-    if (type == :grace_day_usage && (params[:grace_day_threshold].nil? || params[:date].nil? || params.length != 2)) ||
-       (type == :grade_drop && (params[:percentage_drop].nil? || params[:consecutive_counts].nil? || params.length != 2)) ||
-       (type == :no_submissions && (params[:no_submissions_threshold].nil? || params.length != 1)) ||
-       (type == :low_grades && (params[:grade_threshold].nil? || params[:count_threshold].nil? || params.length != 2))
+    if (type == 1 && (params[:grace_day_threshold].nil? || params[:date].nil? || params.length != 2)) ||
```
victorhuangwq, line 16: this seems like a regression? Was there an issue when it comes to using an enum?

app/models/risk_condition.rb
```diff
@@ -11,44 +11,107 @@ class RiskCondition < ApplicationRecord
...
+       (type == 2 && (params[:percentage_drop].nil? || params[:consecutive_counts].nil? || params.length != 2)) ||
+       (type == 3 && (params[:no_submissions_threshold].nil? || params.length != 1)) ||
+       (type == 4 && (params[:grade_threshold].nil? || params[:count_threshold].nil? || params.length != 2))
       raise "Parameters for risk condition does not match type!"
     end
 
-    version = RiskCondition.get_version(course_id, type)
     options = { course_id: course_id, condition_type: type, parameters: params, version: version }
     newRiskCondition = RiskCondition.new(options)
```
victorhuangwq, line 24: Nit picking here. but you were using underscore casing, but this one is in camel case. Suggest changing to new_risk_condition to be consistent

app/models/risk_condition.rb
```diff
@@ -11,44 +11,107 @@ class RiskCondition < ApplicationRecord
...
+              break
+            end
+          end
+        else
+          no_change = false
+        end
+
+        unless no_change
+          # puts "case 4: instructor changed conditions this time; previous conditions were either unset, or different from current parameters"
+          params.map do |k, v|
```
victorhuangwq, line 84: this is somehow not throwing an error for malformed bodys.

`{"no_submi312ssions":{"no_submissions_threshold":4}}`

app/models/risk_condition.rb
```diff
@@ -11,44 +11,107 @@ class RiskCondition < ApplicationRecord
...
+        #   puts "case 5: previous conditions and current conditions match and no update is needed"
+        end
+        return conditions
+      end
+    else
+      # puts "case 6: previous conditions were not set (i.e. max_version == 0) and instructor wants to set it to something"
+      conditions = []
+      params.map do |k, v|
+        new_condition = create_condition_for_course_with_type(course_id, self.condition_types[k], v, max_version + 1)
+        conditions << new_condition
```
victorhuangwq, line 98: case 6 seems to do the exact same thing as case 4. Could combine them

victorhuangwq (changes requested):
I think there is some issue with the malformed params. Could you look into it?
At the same time, when the malformed params are given a http error should be thrown with the appropriate status codes.

app/controllers/metrics_controller.rb
xinyis991105: See latest commit.

app/models/risk_condition.rb
xinyis991105: I believe comparing an int against an enum hash doesn't work as intended. But enum does help with drawing correspondence between the hash and the int.

app/models/risk_condition.rb
xinyis991105: See latest commit.

app/models/risk_condition.rb
xinyis991105: See latest commit.

app/models/risk_condition.rb
victorhuangwq: I think that is likely due to the fact that you have added :no_condition_selected to the front of the enum. The number of the enum is ordered based on the index in the array.

A solution to your issue would be to specific the enum index specifically in this manner
`  enum status: { active: 0, archived: 1 }`

xinyis991105: I tried the hash notation instead of array and it still doesn't work. Actually, if you take a look at the previous code, 1 represents grace_day_usage and so on so forth so 0 reserved for no condition selected has already been accounted for.

victorhuangwq: hmm okay. Anyway this doesn't really affect the functionality, so we can just leave this at that. Can come back to address this in a future PR.

victorhuangwq (approved):
LGTM

Issues with malformed datatype has been addressed accordingly.

PR #1242, opened 2020-11-12
Update Watchlist Instances Endpoint

victorhuangwq:
## Description

This endpoints allows the frontend to update the watchlist endpoint at `metrics/update_watchlist_instances`

## Motivation and Context

This is for the purpose of the metrics feature.

## How Has This Been Tested?

This has been tested using the following scenario:

Started with instance 1 and 2
1. `{method:"contact",ids:[2]}` causes 2 to update to `contacted`
1. `{method:"contact",ids:[1]}`causes 1 to be updated to `contacted`
1. `{method:"contact",ids:[1]}` raises an error, as you cannot contact it again
1. `{method:"resolve",ids:[1,2]}` causes 1 and 2 to both be updated, in a single transaction
1. `{method:"contact",ids:[3]}` throws an error cause 3 cannot be found
1. `{methoed:"contact",ids:[1]}` throws an error as method cannot be found
1. `{method:"contct",ids:[1]}` throws an error as method "contct" is not allowed 
1. `{method:"contact",idss:[1]}` throws an error cause ids cannot be found

## Types of changes

## Checklist:

### DIFF
--- a/app/controllers/metrics_controller.rb
+++ b/app/controllers/metrics_controller.rb
@@ -36,7 +36,7 @@ class MetricsController < ApplicationController
 			instances = WatchlistInstance.get_instances_for_course(course_name)
 			render json: instances
 		rescue => error
-			render :text => 'Not Found', :status => '404'
+			render json: {error:error.message}, status: :not_found
 			return
 		end
 	end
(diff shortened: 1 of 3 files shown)


app/controllers/metrics_controller.rb
```diff
@@ -33,12 +33,52 @@ def get_watchlist_instances
...
+		# example json body {"method":"resolve","ids":[1,2,3]}
+		# method: update, resolve
+		# ids: [1,2,3...] list of ids to be updated
+		
+		begin
+			course_name = params[:course_name]
+		rescue => error
+			render json:  {error:"Course Not Found"}, :status => :not_found
+			return
+		end
```
xinyis991105, line 58: I believe when params does not contain a pair with the key being course_name, it will simply return nil. The double check needed here is probably to try to obtain a course row with the provided course name (raise an error if course_name is nil) and see if such a row exists and raise an error if it doesn't.

xinyis991105 (changes requested):
Tested according to procedures specified in PR. Endpoint achieves its functionality. Left a small comment regarding error checking.

app/controllers/metrics_controller.rb
victorhuangwq: okay! I will amend that accordingly

xinyis991105 (approved):
I was actually thinking whether we need to attempt to find a course row with the corresponding course name and report an error if such a row cannot be found. But I also realize when updating the watchlist instance ids, the methods check for whether these instances exist themselves so checking whether a course name is blank is probably sufficient for now. LGTM!

PR #1243, opened 2020-11-12
Get Number of New Watchlist Instance Endpoint

victorhuangwq:
## Description

This endpoint allows you to get the number of new watchlist instance

## Motivation and Context

This is for the purpose of the notification badge on the course page, as we would prefer not needing to get all watchlist instances just to get the number of new ones.

## How Has This Been Tested?

Started with no new watchlist instances in the records
1. called endpoint, returns `{num_new:0}`
1. Add a new instance, called endpoint, returns `{num_new:1}`
1. Add another new instance, called endpoint, returns `{num_new:2}`
1. Call with invalid course, gets redirected to 404.

## Types of changes

## Checklist:

### DIFF
--- a/app/controllers/metrics_controller.rb
+++ b/app/controllers/metrics_controller.rb
@@ -33,10 +33,25 @@ class MetricsController < ApplicationController
 		# On error, a 404 error is returned
 		begin
 			course_name = params[:course_name]
-			conditions = WatchlistInstance.get_instances_for_course(course_name)
-			render json: conditions
+			instances = WatchlistInstance.get_instances_for_course(course_name)
+			render json: instances
 		rescue => error
-			render :text => 'Not Found', :status => '404'
+			render json: {error:error.message}, status: :not_found
+			return
+		end
+	end
+
+	action_auth_level :get_number_new_instances, :instructor
+	def get_number_new_instances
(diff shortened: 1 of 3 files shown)


app/controllers/metrics_controller.rb
```diff
@@ -33,10 +33,25 @@ def get_watchlist_instances
 		# On error, a 404 error is returned
 		begin
 			course_name = params[:course_name]
-			conditions = WatchlistInstance.get_instances_for_course(course_name)
-			render json: conditions
+			instances = WatchlistInstance.get_instances_for_course(course_name)
+			render json: instances
```
xinyis991105, line 37: To be consistent across all endpoints, should we also add status: :ok here as well? If necessary, I'll make sure I add these for subsequent endpoints too.

xinyis991105 (changes requested):
To be consistent, can we also add :status => :ok

app/controllers/metrics_controller.rb
victorhuangwq: Yes, I agree that we should return status =>:ok. I will amend that as well

xinyis991105 (approved):
LGTM!

PR #1248, opened 2020-11-15
Inconsistent font sizes in Edit Assessment > Edit Problem

amartinez116:
## Description

Adjusted the buttons in 'Edit Assessment > Edit Problem' in order to make the "Delete this problem" and "Save problem" buttons quite the same size.

Old:
image

New:
image

## Motivation and Context

The format of the buttons is quite different since the "Save Problem" button is an < input > tag while "Delete this problem" is an < a > tag. 

## How Has This Been Tested?

Tested this locally on chrome and safari browsers. 

## Types of changes

## Checklist:

### DIFF
--- a/app/views/problems/_fields.html.erb
+++ b/app/views/problems/_fields.html.erb
@@ -18,6 +18,8 @@ unchecked." %>
 <div class="action_buttons">
 	<%= f.submit "Save Problem" , {:class=>"btn primary"} %>
 	<% if @problem then %>
-	  <%= link_to "Delete this Problem", [@course, @assessment, @problem], method: :delete, :style=> 'font-size:1em;', class: 'btn btn-danger', data: { confirm: "Deleting will delete all associated problem data and cannot be undone. Are you sure you want to delete this problem?" } %>
+	<button class="delete_btn"> 
+	  <%= link_to "Delete this Problem", [@course, @assessment, @problem], method: :delete, class: 'btn btn-danger', data: { confirm: "Deleting will delete all associated problem data and cannot be undone. Are you sure you want to delete this problem?" } %>
+	</button>
 	<% end %>
 </div>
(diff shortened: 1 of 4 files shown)


app/views/problems/_fields.html.erb
```diff
@@ -18,6 +18,6 @@ unchecked." %>
 <div class="action_buttons">
 	<%= f.submit "Save Problem" , {:class=>"btn primary"} %>
 	<% if @problem then %>
-	  <%= link_to "Delete this Problem", [@course, @assessment, @problem], method: :delete, class: 'btn btn-danger', data: { confirm: "Deleting will delete all associated problem data and cannot be undone. Are you sure you want to delete this problem?" } %>
+	  <%= link_to "Delete this Problem", [@course, @assessment, @problem], method: :delete, :style=> 'font-size:1em;', class: 'btn btn-danger', data: { confirm: "Deleting will delete all associated problem data and cannot be undone. Are you sure you want to delete this problem?" } %>
```
oliverli, line 21: The problem is that the Save Problem button is an `<input type="submit">` button while the Delete This Problem button is an `<a>` link formatted to look like a button using CSS, so they end up rendering differently even though they're both `btn` classes in CSS.

If it is possible to refactor this code a little so that both are `<a>` anchors that will be ideal because inline CSS is messy, but if that's not possible/too high effort let me know and I'll approve it

oliverli (changes requested):

app/views/problems/_fields.html.erb
amartinez116: Yup, I noted the tag difference in the context section. 

Changing the rendering is a bit difficult since the save button is utilizing forms for rendering while the delete button is simply utilizing the delete method. Thus, a better solution (instead of changing a bunch of code around) is to encapsulate the delete button `<a>` into a `<button>` tag in order to keep the same style as the `<input type="submit">`. That way, we can repeat this style for future buttons that are being rendered differently.

oliverli (approved):
LGTM

PR #1249, opened 2020-11-18
API endpoint for refreshing watchlist instances

xinyis991105:
## Description
This PR adds the API endpoint for bulk refreshing watchlist instances for a course. In addition, it also patched app/models/risk_conditions.rb such that the method update_current_for_course now utilizes transaction.

## Motivation and Context
This is part of the metrics feature functionality.

## How Has This Been Tested?
1. Add a new course with a reasonable number of grace days.
2. Add at least 3 users to that course.
3. Add a reasonable number of assignments, each consuming preferably different number of grace days (by tweaking due date and end date). For ease of later testing, set only one problem to each assignment with weight 100. Act as each user added in step 2. Submit as these users such that the number of no-submissions for each is varied. Act as instructor again and grade their submissions. Make sure the grades are spread over a reasonable range and pay attention to the ups and downs of one particular user. Take note of each user's trend so that comparison is easier for later purposes. After all of this is done, make sure to change the end date of each assignment to a past date. Unreleased grades are not considered for the grade-related conditions.
**Testing is first done case by case.**
4. Add a grace day usage condition to the course. Visit AutoPopulated/metrics/refresh_watch_list_instances. See if the returned instances match expectation. Tweak the condition to see if refreshed instances reflect the modification.
5. Add a grade drop condition. Visit AutoPopulated/metrics/refresh_watch_list_instances. See if the returned instances match expectation. Tweak the condition to see if refreshed instances reflect the modification.
6. Add a no submissions condition. Visit AutoPopulated/metrics/refresh_watch_list_instances. See if the returned instances match expectation. Tweak the condition to see if refreshed instances reflect the modification.
7. Add a low grades condition. Visit AutoPopulated/metrics/refresh_watch_list_instances. See if the returned instances match expectation. Tweak the condition to see if refreshed instances reflect the modification.
**Testing is then done in a hybrid manner.**
8. Update current conditions such that they are a hybrid of some of the four types. See if the refreshed instances match expectation.
**Lastly**
9. Refresh twice in a row. There shouldn't be any change in the records returned by the endpoint.
10. It should also be the case that none of the returned instances should include instructors' information.

## Types of changes

## Checklist:

### DIFF
--- a/app/models/watchlist_instance.rb
+++ b/app/models/watchlist_instance.rb
@@ -15,9 +15,9 @@ class WatchlistInstance < ApplicationRecord
 
   def self.refresh_instances_for_course(course_name)
     begin
-      course_id = Coures.find_by(name: coures_name).id
-      current_conditions = RiskCondition.get_current_for_course(coures_name)
-      current_instances = WatchlistInstance.where(course_id: course_id)
+      course = Course.find_by(name: course_name)
+      current_conditions = RiskCondition.get_current_for_course(course_name)
+      current_instances = WatchlistInstance.where(course_id: course.id)
     rescue NoMethodError
       raise "Course #{course_name} cannot be found"
     end
@@ -25,7 +25,7 @@ class WatchlistInstance < ApplicationRecord
     # check whether the watchlist instances are up-to-date
     current_instances_condition_ids = (current_instances.map { |inst| inst.risk_condition_id }).uniq.to_set
     current_condition_ids = (current_conditions.map { |c| c.id }).uniq.to_set # should be unique by design but add the suffix to ensure
-    if current_instances_conditions_ids == current_instances_conditions_ids
+    if current_instances_condition_ids == current_condition_ids
       return current_instances
     else
       # update!
@@ -39,9 +39,11 @@ class WatchlistInstance < ApplicationRecord
         return []
       else
         # case 2: current risk conditions exist
-        new_instances = self.add_new_instances_for_conditions(current_conditions, course_id)
+        new_instances = self.add_new_instances_for_conditions(current_conditions, course)
+        return new_instances
       end
     end
+  end
 
   def archive_watchlist_instance
     if self.new_watchlist?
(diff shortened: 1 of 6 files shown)


app/models/watchlist_instance.rb
```diff
@@ -22,6 +22,38 @@ def self.get_num_new_instance_for_course(course_name)
...
+      course = Course.find_by(name: course_name)
+      current_conditions = RiskCondition.get_current_for_course(course_name)
+      current_instances = WatchlistInstance.where(course_id: course.id)
+    rescue NoMethodError
+      raise "Course #{course_name} cannot be found"
+    end
+
+    # check whether the watchlist instances are up-to-date
+    current_instances_condition_ids = (current_instances.map { |inst| inst.risk_condition_id }).uniq.to_set
+    current_condition_ids = (current_conditions.map { |c| c.id }).uniq.to_set # should be unique by design but add the suffix to ensure
```
victorhuangwq, line 36: This condition check for to reduce calculation actually causes a problem. Because if the conditions are kept the same, but the assignment deadlines etc are changed, this will actually prevent the instructor from updating the watchlist.

victorhuangwq: Removing this check solved my issue

app/models/watchlist_instance.rb
```diff
@@ -83,4 +115,176 @@ def resolve_watchlist_instance
...
+
+        when "grade_drop"
+          percentage_drop = (condition.parameters[:percentage_drop]).to_f
+          consecutive_counts = condition.parameters[:consecutive_counts]
+          
+          categories = course.assessment_categories
+          asmt_arrs = categories.map { |category| course.assessments_with_category(category).ordered }
+          asmt_arrs.select! { |asmts| asmts.count >= consecutive_counts}
+          for cud in course_user_data
+            new_instance = self.add_new_instance_for_cud_grade_drop(course, condition.id, cud, asmt_arrs, consecutive_counts, percentage_drop)
```
victorhuangwq, line 148: should typecheck and convert consecutive counts to integers, else the comparison will fail later

app/models/watchlist_instance.rb
```diff
@@ -83,4 +115,176 @@ def resolve_watchlist_instance
...
+            new_instances << new_instance unless new_instance.nil?
+          end
+
+        when "no_submissions"
+          no_submissions_threshold = condition.parameters[:no_submissions_threshold]
+
+          for cud in course_user_data
+            new_instance = self.add_new_instance_for_cud_no_submissions(course, condition.id, cud, no_submissions_threshold)
```
victorhuangwq, line 156: should typecheck and convert no_submissions_threshold to integer, else the comparison will fail later

app/models/watchlist_instance.rb
```diff
@@ -83,4 +115,176 @@ def resolve_watchlist_instance
...
+            new_instances << new_instance unless new_instance.nil?
+          end
+
+        when "low_grades"
+          grade_threshold = condition.parameters[:grade_threshold].to_f
+          count_threshold = condition.parameters[:count_threshold]
+
+          for cud in course_user_data
+            new_instance = self.add_new_instance_for_cud_low_grades(course, condition.id, cud, grade_threshold, count_threshold)
```
victorhuangwq, line 165: should typecheck and convert grade_threshold, count_threshold to integer, else the comparison will fail later

app/models/watchlist_instance.rb
```diff
@@ -83,4 +115,176 @@ def resolve_watchlist_instance
...
+        
+        when "grace_day_usage"
+          grace_day_threshold = condition.parameters[:grace_day_threshold]
+          date = condition.parameters[:date]
+          asmts = course.assessments.ordered
+          for cud in course_user_data
+            asmts_before_date = asmts.where("updated_at < ?", date)
+            latest_asmt = asmts_before_date.last
+            next if latest_asmt.nil?
+            new_instance = self.add_new_instance_for_cud_grace_day_usage(course, condition.id, cud, asmts_before_date, grace_day_threshold)
```
victorhuangwq, line 136: should typecheck and convert grace_day_threshold to integers, else the comparison will fail later

app/models/watchlist_instance.rb
victorhuangwq: percentage_drop should be float

app/models/watchlist_instance.rb
victorhuangwq: grade_threshold should be float

app/models/watchlist_instance.rb
```diff
@@ -83,4 +115,176 @@ def resolve_watchlist_instance
...
+            new_instances << new_instance unless new_instance.nil?
+          end
+
+        when "grade_drop"
+          percentage_drop = (condition.parameters[:percentage_drop]).to_f
+          consecutive_counts = condition.parameters[:consecutive_counts]
+          
+          categories = course.assessment_categories
+          asmt_arrs = categories.map { |category| course.assessments_with_category(category).ordered }
+          asmt_arrs.select! { |asmts| asmts.count >= consecutive_counts}
```
victorhuangwq, line 146: here as well, need to do a typecasting for consecutive_counts to ensure that it is an int

victorhuangwq:
Testing Steps:
1. Create a new course with 3 users
1. Create 3 assignments
1. Tested No Submissions, check to see if  appropriate results based on threshold. Changed Assignment deadlines and tested as well
1. Tested Low Grades. Check to see if appropriate results are obtained based on threshold.
1. Tested Grade Drop. 
1. Tested Grace Days Used.

Below are the issues that I have found

**Functional Issues Found**
- [ ] No Submissions Considered Assignments that are still ongoing. Given three assignments, 1, 2 & 3, and the watchlist is looking out for 2 no submissions. The student did not submit assignment 1. At the same time, the student did not submit assignment 3 yet, because it is still ongoing. 
**Expected**: Student should not be counted 
**Actual** Student considered as two no submissions. 
**Suggestion** Do no consider assignments that are still ongoing 
- [ ] Low Grades Assignment Considered Assignments that Grades have not been released. Given an assignments, and the watchlist is looking out for low grades of less than 50. The  student obtained 100 points, but the grades has not been released.
**Expected**: Student should not have considered as having low grades, because the assignment's grades has not been released
**Actual** Student considered as having low grades
**Suggestion** Do no consider grades of assignments that has not been released

victorhuangwq (changes requested):
The main overall issues to consider:
- The types of the parameters coming from `condition.parameters` are likely to be uncertain (can be string or integer). So they need to be typechecked and typecasted accordingly.
- Ensuring that the grading deadline and grade has been released for low grades, as well as when assignment has been submitted for no_submission

xinyis991105:
> Testing Steps:
> 
>     1. Create a new course with 3 users
> 
>     2. Create 3 assignments
> 
>     3. Tested No Submissions, check to see if  appropriate results based on threshold. Changed Assignment deadlines and tested as well
> 
>     4. Tested Low Grades. Check to see if appropriate results are obtained based on threshold.
> 
>     5. Tested Grade Drop.
> 
>     6. Tested Grace Days Used.
> 
> 
> Below are the issues that I have found
> 
> **Functional Issues Found**
> 
>     * [x]  No Submissions Considered Assignments that are still ongoing. Given three assignments, 1, 2 & 3, and the watchlist is looking out for 2 no submissions. The student did not submit assignment 1. At the same time, the student did not submit assignment 3 yet, because it is still ongoing.
>       **Expected**: Student should not be counted
>       **Actual** Student considered as two no submissions.
>       **Suggestion** Do no consider assignments that are still ongoing
> 
>     * [ ]  Low Grades Assignment Considered Assignments that Grades have not been released. Given an assignments, and the watchlist is looking out for low grades of less than 50. The  student obtained 100 points, but the grades has not been released.
>       **Expected**: Student should not have considered as having low grades, because the assignment's grades has not been released
>       **Actual** Student considered as having low grades
>       **Suggestion** Do no consider grades of assignments that has not been released

There is actually a problem with the low grades situation. I managed to get the check for whether the final score has been released working for students who submitted something. But for those who didn't submit, I believe Autolab defaults to 0.0 as a final score for both when the student or the instructor tries to view the final score. In this case, suppose the grading deadline has already passed but a student didn't make any submission at all, do we just take it a 0 or would we rather check whether the same assessment's score has been released for other students? The latter might take a while a figure out though.

victorhuangwq:
So the situation is that :
1. the student did not submit
1. grades are released, they get a 0 for their score
1. they are counted as having a low grade 

I was wondering if you could just do a quick check if there is a latest_submission_id in assessment_user_datum?

xinyis991105:
You mean when a student didn't submit, we don't count that as low grades? That could also work. In fact, that's so much simpler, I'll do a latest submission check as a solution.

xinyis991105:
New commit patches:
1. Resolved all the type casting issues. (See resolved conversation and relevant code.)
2. Added grading deadline check, latest submission check, and score released check for :grade_drop and :low_grades conditions. Refreshing watchlist instances this time won't consider any assessment user datum that falls into one of the three buckets aforementioned.
3. Removed check for change of current risk conditions at the beginning of refreshing.
4. For grace day usage check, instead of including assessment user data whose updated_at is before the specified date in the parameter condition, end_at date is used. It makes more sense since normally, students won't be allowed to submit after the end date.

victorhuangwq:
In addition to the previous tests, the following test was done
1. Checking if no submission condition considers ongoing assessment by creating a new assessment and calling the endpoint
1. No submission is not counted for grade drop or low grades

Everything seems okay!

victorhuangwq (approved):
LGTM

PR #1251, opened 2020-11-25
Callbacks for grace day usage

xinyis991105:
## Description
This PR adds the necessary callbacks that incrementally update watchlist instances for a course's grace day usage condition.

## Motivation and Context
This change is necessary for the metrics feature.

## How Has This Been Tested?
1. Add a grace day usage condition to the current metrics of the course of your preference. For initial testing, set grace day threshold to 1 and date to after testing date.
2. Create a new assignment for the course such that it is not yet past end date but already past due date.
3. Act as a student and make a submission to that assignment. Visit the endpoint to obtain current watchlist instances. Confirm that a new instance is created for that student.
4. Change the assignment's due date such that the student's use of grace days also changes. Revisit get_watchlist_instances and confirm that a new instance has replaced the old one with updated information.
5. Change the assignment's max grace days such that the student's use of grace days changes. Revisit the endpoint. Confirm a new instance has replaced the old one with updated information.
6. Change the course's late slack such that the student's use of grace days changes. Revisit get_watchlist_instances. Confirm a new instance has replaced the old one.
7. Change the course's grace days such that the student's use of grace days changes. Revisit get_watchlist_instances. Confirm a new instance has replaced the old one.
Note that in the above cases, if the change doesn't suffice to change a student's use of grace days, you should still observe that the previous watchlist instance is gone since it was archived, and if there is still a current one, its id should have been incremented.
8. Try with different combinations of the changes above. Compare the results of get_watchlist_instances with your expectation.

## Types of changes

## Checklist:

See engineering channel for the callback condition issue. It has been fixed in this PR but whether it's a local or a more generic issue is yet to be confirmed.


app/models/assessment_user_datum.rb
```diff
@@ -155,6 +155,8 @@ def invalidate_cgdubs_for_assessments_after
       end
 
       Rails.cache.delete course_user_datum.ggl_cache_key
+
+      course_user_datum.update_cud_gdu_watchlist_instances
```
victorhuangwq, line 159: hi this is just because I'm not very clear on the caching is working here, but could you explain how this part works?

app/models/course.rb
```diff
@@ -195,6 +195,48 @@ def sanitized_name
...
     save!
+
+    update_course_gdu_watchlist_instances
+  end
+
+  # Update the grace day usage condition watchlist instances for each course user datum
+  # This is called when:
+  # - Grace days or late slack have been changed and the record is saved
+  # - invalidate_cgdubs are somehow incurred
+  def update_course_gdu_watchlist_instances
```
victorhuangwq, line 206: I suggest that this function be moved into the watchlist instance model instead. This function modifies the state of the watchlist instance, not really that of the course

app/models/course_user_datum.rb
```diff
@@ -249,6 +249,88 @@ def ggl_cache_key
...
+    end
+  end
+
+  # This returns an unsaved watchlist instance if current course user datum qualifies for
+  # being put onto the watchlist. The caller of this method needs to save it on their own if necessary
+  # This can be called from:
+  # - Watchlist's bulk refreshing
+  # - Callback style update within course user datum (see above)
+  # - Callback style update on a course level (see course.rb)
+  def new_gdu_watchlist_instance(grace_day_threshold, date, condition_id)
```
victorhuangwq, line 302: I feel that this should be in the watchlist instance as well, because this adds a new watchlist instance, and not modifies course_user_datum

app/models/course_user_datum.rb
```diff
@@ -249,6 +249,88 @@ def ggl_cache_key
     "ggl/dua-#{dua}/u-#{self.id}"
   end
 
+  # This method call is used specifically for the purpose of callback style update to watchlist
+  # Need to archive old instances and also add new instances
+  def update_cud_gdu_watchlist_instances
```
victorhuangwq, line 254: having a callback function make sense in the course_user_datum, but then it should call functions in the watchlist.rb model to do the actual update. 

Do let me know if this is not clear.

victorhuangwq (commented):
Tested changing both the 
1. Course Grace Days
1. Assignment Grace Days

And the watchlist instances are updated accordingly

app/models/course_user_datum.rb
```diff
@@ -249,6 +249,88 @@ def ggl_cache_key
...
+    # At this point, all relevant previously cached grace day usage information should be invalidated
+    # Calls to calculate grace day usage should be from scratch
+    
+    # Ignore if this CUD is an instructor or CA
+    return unless self.student?
+    
+    # Get current grace day condition
+    current_conditions = RiskCondition.get_current_for_course(self.course.name)
```
victorhuangwq, line 262: i think getting a particular type of condition should be helper functions inside risk condition model

app/models/course_user_datum.rb
```diff
@@ -249,6 +249,88 @@ def ggl_cache_key
...
+    
+    # Check for whether there exists one of type :grace_day_usage
+    grace_day_usage_condition = current_conditions.select { |c| c.condition_type.to_sym == :grace_day_usage }
+    return if grace_day_usage_condition.count != 1
+    grace_day_usage_condition = grace_day_usage_condition[0]
+    
+    # Get :grace_day_threshold and :date
+    condition_id = grace_day_usage_condition.id
+    grace_day_threshold = grace_day_usage_condition.parameters[:grace_day_threshold].to_i
+    date = grace_day_usage_condition.parameters[:date]
```
victorhuangwq, line 273: the logic here should be handled in risk condition model as well

victorhuangwq (changes requested):
I think in general the functionality is achieved, but I have some suggestions with regards to the structure and location of where the bulk of the logic should be located for the purpose of clarity and modularity.

app/models/assessment_user_datum.rb
xinyis991105: This added line doesn't have anything to do with the caching part. I added this line inside of the transaction with the lock on just to maintain consistency. I believe whenever cgdubs are invalidated, we need to update the watchlist instances (since anything could happen to grace days used) so I thought it would be a good idea to immediately ask CUD to update after invalidating.

xinyis991105:
Added modularity for the following functionality:
- Moved asmts_before_date into app/models/course.rb(:252)
- Moved get_gdu_condition_for_course into app/models/risk_condition.rb(:105)
- Moved the actual update of watchlist instances into app/models/watchlist_instance.rb(:54) and such methods are called from other active records

Fixed some subtle details:
- Instead of only checking for student status, update for cud also checks for whether cud has dropped or not (watchlist_instance.rb:94)
- course.rb already has a method for acquiring all qualified students so I've switched to that instead of using CourseUserDatum.where(...) to get students for a course. (watchlist_instance.rb:63)
- When acquiring assessments before a date, due_at is evaluated instead of end_at since that's what is used to order the assessments when you call course.assessments.ordered. (course.rb:254)
- I've removed the save operation from self.add_new_instance_for_cud_*condition* in watchlist_instance.rb(:257) so that whoever calls these private methods are responsible for saving them in transactions.

Lastly, a call to refresh_instances_for_course is added inside of update_current_for_course(course_name) in risk_condition.rb(:47) so that whenever an instructor updates current metrics for a course, watchlist instances are automatically updated too.

app/models/watchlist_instance.rb
```diff
@@ -113,62 +195,67 @@ def resolve_watchlist_instance
...
-        case condition.condition_type
+    
+    conditions.each do |condition|
+      case condition.condition_type
+      
+      when "grace_day_usage"
+        grace_day_threshold = condition.parameters[:grace_day_threshold].to_i
+        date = condition.parameters[:date]
+        asmts_before_date = course.asmts_before_date(date)
+        next if asmts_before_date.count == 0 # go to the next condition loop if there is no latest assessment
```
victorhuangwq, line 206: I saw that there are some changes here, could you let me know in general what this change is about?

app/models/risk_condition.rb
```diff
@@ -93,10 +95,26 @@ def self.update_current_for_course(course_name, params)
       # else
       #   puts "case 5: previous conditions and current conditions match and no update is needed"
       end
+      WatchlistInstance.refresh_instances_for_course(course_name)
```
victorhuangwq, line 98: I wonder if there is a better way to this as compared to having it at three different areas?

victorhuangwq: But I think as of now this is okay.

victorhuangwq (approved):
- Tested both changing the grace days of the course, and the grace days available for the assignments, and see that the watchlist instances updated accordingly. 
- The watchlist instance also refreshes upon updating of the settings now. 

LGTM

app/models/watchlist_instance.rb
xinyis991105: Previously, the 'next' is inside of the loop for each course_user_datum when I fail to acquire their corresponding latest aud before date. I thought it would be better if I pull it out and check this condition on an assessment level and stop a little early if there are no assessments at all before the specified date.

app/models/risk_condition.rb
xinyis991105: Yeah I was thinking about that too. But this was the fastest fix. We could add optimize this later PRs. I'll take a note.

PR #1253, opened 2020-12-01
Callbacks for grade related conditions

xinyis991105:
## Description
This PR adds the necessary callbacks for updating watchlist instances of grade related conditions (i.e. grade_drop and low_grades).

## Motivation and Context
This is required for the new metrics feature.

## How Has This Been Tested?
Changes or checkpoints that incur grade related condition watchlist updates are listed as below:
1. Assessment (only perform callback if the assessment is **past its end date**):
- [x] after_save: Due date changed
- [x] after_save: Grace day changed
- [x] after_save: Version threshold changed
- [x] after_destroy
2. Course:
- [x] after_save: Late slack changed
- [x] after_save: Grace days changed
- [x] after_save: Late penalty changed
- [x] after_save: Version threshold changed
- [x] after_save: Version penalty changed
3. Problem (only perform callback if the problem's assessment is already **past its end date**):
- [x] after_save: Max score changed
- [x] after_destroy
- [x] after_create
4. Score (only perform callback if the score's **submission is a latest one** and if the score is **released**):
- [x] after_destroy
- [x] after_save: Released or score changed
5. Submission (only perform callback if the submission is a **latest one**):
- [x] after_save: Tweak ID changed

Testing has been done on each of these individual conditions. For further testing, feel free to combine any of these. In addition, the logic for whether we should add an instance or not has been tested in the PR for refreshing watchlist instances so there is no urgent need to confirm the correctness of that.

## Types of changes

## Checklist:

### DIFF
--- a/app/models/assessment.rb
+++ b/app/models/assessment.rb
@@ -337,6 +337,11 @@ class Assessment < ApplicationRecord
     end
   end
 
+  # to be able to calculate total score for an assessment from another model
+  def default_total_score
+    problems.sum :max_score
+  end
+
 private
 
   def saved_change_to_grade_related_fields?
--- a/app/models/assessment.rb
+++ b/app/models/assessment.rb
@@ -42,6 +42,7 @@ class Assessment < ApplicationRecord
   after_save :invalidate_course_cgdubs, if: :saved_change_to_due_at_or_max_grace_days?
   after_save :update_course_grade_watchlist_instances_if_past_end_at, if: :saved_change_to_grade_related_fields?
   after_create :create_AUDs_modulo_callbacks
+  after_destroy :update_course_grade_watchlist_instances_if_past_end_at
 
   # Constants
   ORDERING = "due_at ASC, name ASC"
@@ -350,9 +351,7 @@ class Assessment < ApplicationRecord
 private
 
   def saved_change_to_grade_related_fields?
-    return (saved_change_to_due_at? or saved_change_to_late_penalty_id? or
-            saved_change_to_max_grace_days? or saved_change_to_version_threshold? or
-            saved_change_to_version_penalty_id?)
+    return (saved_change_to_due_at? or saved_change_to_max_grace_days? or saved_change_to_version_threshold?)
   end
 
   def saved_change_to_due_at_or_max_grace_days?
--- a/app/models/assessment.rb
+++ b/app/models/assessment.rb
@@ -228,6 +228,8 @@ class Assessment < ApplicationRecord
     # config file might have an updated custom raw score function: clear raw score cache
     invalidate_raw_scores
 
+    update_course_grade_watchlist_instances
+
     logger.info "Loaded #{config_file_path}"
   end
 
--- a/app/models/assessment.rb
+++ b/app/models/assessment.rb
@@ -227,8 +227,6 @@ class Assessment < ApplicationRecord
     # config file might have an updated custom raw score function: clear raw score cache
     invalidate_raw_scores
 
-    update_course_grade_watchlist_instances_if_past_end_at
-
     logger.info "Loaded #{config_file_path}"
   end
 
--- a/app/models/assessment.rb
+++ b/app/models/assessment.rb
@@ -39,8 +39,7 @@ class Assessment < ApplicationRecord
   trim_field :name, :display_name, :handin_filename, :handin_directory, :handout, :writeup
   after_save :dump_yaml
   after_save :dump_embedded_quiz, if: :saved_change_to_embedded_quiz_form_data?
-  after_save :invalidate_course_cgdubs, if: :saved_change_to_due_at?
-  after_save :invalidate_course_cgdubs, if: :saved_change_to_max_grace_days?
+  after_save :invalidate_course_cgdubs, if: :saved_change_to_due_at_or_max_grace_days?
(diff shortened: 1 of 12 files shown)


xinyis991105:
Bug fixes:
- See customized callback conditions for assessment.rb and course.rb. Previously, I wrongly prefixed the saved_change_to_somefield with a colon, which renders the callback function useless since it returns true all the time. It should now have the ability to distinguish between the various conditions.
- The function all_scores_released? in submission.rb:282 now also accounts for when the number of a submission's score does not equal that of the problems in an assessment. This might occur when one creates a new problem for an assessment and a new score is yet to be assigned.

Post Discussion Decisions:
- Callbacks that specifically deal with latest submission change for an assessment user datum are removed.
- Despite ↑, we retain the callback on submission.rb:33 when there is a tweak change for a submission.
- A new function final_score_ignore_grading_deadline has been added to assessment_user_datum.rb:113 to account for the fact that instructors might release scores before grading deadline. This function is now used in watchlist_instance.rb:378/440 to calculate the final score of a student's submission.
- The callbacks for an assessment now takes into consideration whether an assessment has passed its end date (assessment.rb:344), after which, students are no longer able to turn anything in. If before end date, we do not initiate a course-level watchlist instances refresh, since it is likely no instances will be added anyways.
- Callbacks in problem.rb are delegated to assessment.rb:344, which, again, checks whether an assessment is past end date. So any change made to problems now also takes an assessment's end date into account.

Edge Cases:
- Callbacks are removed for latest submission change, so there might be the case when a latest submission is removed, and a previous version fills in, but no update of watchlist instance is initiated. We rely on instructors using the refresh button to deal with this situation.
- For submissions and courses, the score adjustments, whenever updated, be it penalty or tweak, a new score adjustment row is always produced. Hence, we can rely on saved_change_to_xxx_id? to detect an update in score adjustment. However, for assessments, the updates are done with respect to fixed-id score adjustment rows. So I added a mini callback for penalty.rb to account for this. But this introduces the probability of double watchlist instance refreshing. Should we leave this edge case up to the instructors to deal with as well? Or is the double updating fine? Or are there other ways to get around this?

app/models/assessment_user_datum.rb
```diff
@@ -106,7 +106,14 @@ def self.grade_type_to_sym(grade_type)
 
   def final_score(as_seen_by)
     @final_score ||= {}
-    @final_score[as_seen_by] ||= final_score! as_seen_by
+    @final_score[as_seen_by] ||= {}
+    @final_score[as_seen_by][:consider_grading_deadline] ||= final_score! as_seen_by
```
victorhuangwq, line 110: may I understand why is modification to the original `final_score` necessary?

app/models/assessment.rb
```diff
@@ -338,10 +341,22 @@ def default_total_score
...
+    end
+  end
+
 private
 
+  def saved_change_to_grade_related_fields?
+    return (saved_change_to_due_at? or saved_change_to_late_penalty_id? or
+            saved_change_to_max_grace_days? or saved_change_to_version_threshold? or
+            saved_change_to_version_penalty_id?)
+  end
```
victorhuangwq, line 356: I think it would be possible to do the time inside the saved_change_to_grade_related_fields instead of having an additional wrapper function

app/models/submission.rb
```diff
@@ -371,6 +376,12 @@ def aud
     assessment.aud_for course_user_datum_id
   end
 
+  def update_cud_grade_watchlist_instances_if_latest
+    if self.aud.latest_submission_id == self.id
+      self.course_user_datum.update_cud_grade_watchlist_instances
```
victorhuangwq, line 381: Would this mean that when one submission is tweaked, all the watchlist instances is updated?

victorhuangwq (commented):
Here are my comments looking at the latest code changes

xinyis991105:
The latest commit refines the code such that it fits all the updated callback criteria (as are reflected in the changes of the original PR comment) and fixes some subtle issues as well, including:
1. Changing the name of the course user datum related updates methods to update_individual_grade_...
2. Reverted final_score method in assessment_user_datum.rb so that the new final_score_ignore_grading_deadline and the old one don't interfere with one another.

victorhuangwq:
**How I tested**

Followed the above checklist in testing setting with:
No metrics set up at all: No issues 👍 
Low grades and Grace Days set up metric conditions set up:  No issues 👍 

I will note that when metrics conditions are set up, the above actions are significantly slower (at least on a development setup) due to the need to recalculate metrics.

We should proceed with adding this PR in first for testing. However, if from our testing, we realize that all these callbacks significantly affects experience on a production system, we will do a feature regression to keep only the callbacks for score release.

app/models/assessment.rb
victorhuangwq: Let me know if this is possible?

victorhuangwq (approved):
As there are no functional issues with this PR, I will be approving it.  There is however, one small note regarding combining of functions I have left on one line of the code. 

Thanks for the work!

app/models/assessment.rb
xinyis991105: Do you want me to add back the saved_change_to_penalty_id?'s? Sure I can do that! I'll take a note and add it in a separate PR.

PR #1254, opened 2020-12-03
Metrics notification badge

ashleyzhang:
## Description
Adds a notification badge on the icon in the metrics button on the main course page. The badge shows the number of new watchlist instances by calling the get_num_new_instances endpoint.
image

## Motivation and Context
This alerts instructors that there are a number of students that their risk metrics have flagged in the watchlist without relying on the instructors to constantly check the watchlist.

## How Has This Been Tested?
Tested locally without setting any risk metrics so that number of new instances is 0, did not show notification badge. Set risk metrics of "students with 1 submitted assignment below 100%" so that number of new instances is 100, notification badge displayed as in image above.

## Types of changes

## Checklist:

### DIFF
--- a/app/views/assessments/index.html.erb
+++ b/app/views/assessments/index.html.erb
@@ -11,11 +11,17 @@
       $.getJSON('metrics/get_num_new_instances',function(data, status){
         if(status=='success' && data['num_new'] != 0){
           // when there is a non-zero number of new instances
-          var new_instances_str = `${data["num_new"]} new`;
-          $('.new.badge.metrics-badge').attr('data-badge-caption', new_instances_str);
-        } else {
-          // otherwise remove the notification badge
-          $('.row.metrics').empty().append("<i class='material-icons left'>assessment</i> Metrics");
+          var new_instances_str = 
+            `<div class='col s1 metrics'>
+              <i class='material-icons left metrics'>assessment</i>
+            </div>
+            <div class='col s4 metrics'>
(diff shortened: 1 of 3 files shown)


app/views/assessments/index.html.erb
```diff
@@ -8,6 +8,16 @@
...
+      $.getJSON('metrics/get_num_new_instances',function(data, status){
+        if(status=='success' && data['num_new'] != 0){
+          // when there is a non-zero number of new instances
+          var new_instances_str = `${data["num_new"]} new`;
+          $('.new.badge.metrics-badge').attr('data-badge-caption', new_instances_str);
+        } else {
+          // otherwise remove the notification badge
+          $('.row.metrics').empty().append("<i class='material-icons left'>assessment</i> Metrics");
+        }
+      });
```
victorhuangwq, line 20: as the this function has to wait for the ajax call to complete before removing the notification badge, if there are no new watchlist instances, the "new" button appears momentarily before disappearing. 

Would it be possible to add the badge only if there is non-zero? This is important as most people will start out not using the metrics feature yet.

victorhuangwq (changes requested):
Badge is of a good size and location. Just a small fix required about the badge appearing momentarily even for situations when there are no watchlist instances.

app/views/assessments/index.html.erb
ashleyzhang: Good catch - just pushed a fix for it!

victorhuangwq (approved):
Tested and working.
LGTM!

PR #1257, opened 2020-12-03
SQL queries fix

fanpu:
## Description
Autolab currently breaks when using MySQL Server 8.0 as its backend. This is due to several ActiveRecord queries that uses raw strings to access the SQL interface directly. Without the benefit of Rail's ORM this greatly reduces the database compatibility so the breakage is not surprising.

This PR fixes two of the offending queries. I have not done a complete sweep of other places where this may break and I plan to check sometime soon.

## Motivation and Context
Previously `nightly.autolabproject.com` was giving this error:

```
ActionView::Template::Error (Mysql2::Error: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'system = TRUE)' at line 1: SELECT `announcements`.* FROM `announcements` WHERE `announcements`.`persistent` = TRUE AND (course_id = 1 OR system = TRUE)):
    1: <!--
    2: <% if local_assigns[:announcements] && params[:controller] != 'announcements' then %>
    3:   <ul class="annoucement gray-box">
    4:     <%= render local_assigns[:announcements] %>
    5:   </ul>
    6: <% end %>
    7: -->
```

## How Has This Been Tested?
After fix, page renders successfully

DeepinScreenshot_select-area_20201203161927

## Types of changes

## Checklist:


app/controllers/assessments_controller.rb
```diff
@@ -66,9 +66,11 @@ class AssessmentsController < ApplicationController
 
   def index
     @is_instructor = @cud.has_auth_level? :instructor
-    @announcements = Announcement.where("start_date < :now AND end_date > :now", now: Time.now)
-                                 .where("course_id = ? OR system = ?", @course.id, true)
-                                 .where(persistent: false).order(:start_date)
+    announcements_tmp = Announcement.where("start_date < :now AND end_date > :now",
+       now: Time.now)
+      .where(persistent: false)
+    @announcements = announcements_tmp.where(course_id: @course.id)
```
victorhuangwq, line 72: Just a nit, but why is it necessary to create a temp variable here?

victorhuangwq (approved):
Just a small nit picking. But other than that there are no issues, so I'm approving it first.

app/controllers/assessments_controller.rb
cg2v: look at the line below. Announcements_tmp is referenced again. The activerecord .or is not very flexible

fanpu: Chaskiel's comment is correct

app/controllers/application_controller.rb
```diff
@@ -292,8 +292,9 @@ def run_scheduler
   end
 
   def update_persistent_announcements
-    @persistent_announcements = Announcement.where(persistent: true)
-                                            .where("course_id = ? OR system = ?", @course.id, true)
+    @persistent_announcements = Announcement
+      .where(persistent: true, course_id: @course.id)
+      .or(Announcement.where(persistent: true, system: true))
```
wasabigeek, line 297: random comment: can you move `Announcement.where(course_id: ...).or(Announcement.where(system: true))` into a scope and use that instead? Though I'd imagine it would need to be the start of the chain

fanpu: Hey Nic, what a surprise!! :D It's definitely been a while hahah. 

Yeah, I agree that using a scope will help with the code repetition here, haven't used that in a long time so did not think about that one. Great suggestion!!

wasabigeek: Happened to chance across this, let's catch up separately haha

PR #1262, opened 2020-12-07
Watchlist functionality integration with endpoints

amartinez116:
## Description

Watchlist now obtains necessary information from the API and displays it. (Styling will be handled in front-end PR). 
Correct student information goes to the designated side tab (new, contacted, resolved, archived). 

image

## Motivation and Context

In order to integrate API endpoints with watchlist front end to have full working functionality. 

## How Has This Been Tested?

Tested locally by setting up different risk metric parameters and having the watchlist information update in real time. 
There are different scenarios which are accounted for:
1) Risk metrics have not been set up. Therefore we view the unidentified metrics message:
image
 
2) Risk metrics have been set up but no students are identified in the current sidetab:
image

3) Risk metrics have been set up and students have been identified: 
image

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation, which is located at [Autolab Docs](https://github.com/autolab/docs)
- [ ] I have updated the documentation accordingly, included in this PR

## Other issues / help required

The following still need to be handled:
- Selected students to contact/resolve and pass their instance_ids to $('#contact_button').click() and  $('#resolve_button').click() functions
- Refresh watchlist when an instructor updates grades
- Same student appears multiple times with different violation tags. Need to condense same student into one row.

If unsure, feel free to submit first and we'll help you along.

PR #1267, opened 2020-12-12
Watchlist entries

ashleyzhang:
## Description
This PR uses the endpoint integrations from https://github.com/autolab/Autolab/pull/1262 to complete the watchlist entries front-end for the new metrics feature. It combines instances for each student into a single watchlist row (i.e. each tab has at most one row for each student). The code for generating the watchlist is moved into a new watchlist.js file.
Screen Shot 2020-12-11 at 10 53 48 PM
Screen Shot 2020-12-11 at 10 53 30 PM
Screen Shot 2020-12-12 at 1 19 08 AM

## Motivation and Context
Completes the watchlist front-end for the new metrics feature (except for the refresh button), with API endpoints integrated.

## How Has This Been Tested?

This is one particular flow that I tried:

1. Set up risk metrics - I set metrics that would lead to a large number of watchlist instances (1 grace day, 1% grade drop for 2 consecutive assignments, 1 no submission, 1 submitted assignment below 100%
2. Go to the watchlist tab -> make sure students are appearing in the `new` tab
3. Click on any student's outgoing gradebook link -> make sure that it links to the correct student's grades
4. Hover over the condition tags -> make sure the correct information about assignments appears
5. Click the contact button for any student -> make sure that it links to a mail client with the correct student email filled in (the browser's default email client has to be set for the mailto to work - if you're using Chrome, you can do that [here](chrome://settings/handlers))
6. Go back to the watchlist page -> make sure the contacted student is no longer in the `new` tab and has moved to `contacted`
7. In the `new` tab, click the resolve button for any student -> make sure the resolved student is no longer in the `new` tab and has moved to `resolved`
8. Check the boxes for a few students and click the top-level contact button -> check that the correct emails are listed in the mailto and that the students are moved into the `contacted` tab
9. Check the boxes for a few students and click the top-level resolve button -> check that the students are moved into the `resolved` tab
10. Resolve a student in the `contacted` tab -> make sure the student moves to the `resolved` tab
11. In the `new` tab, click the top-level select-all checkbox and click contact -> check that the correct emails are listed in the mailto and that all students are moved into the `contacted` tab (there should be no students in `new` now)
12. In the `contacted` tab, click the top-level select-all checkbox and click resolve -> check that all students are moved into the `resolved` tab (there should be no students in `contacted` now)
13. Return to the main course page -> the metrics button should no longer have a notification badge
14. Change a risk metric and return to the watchlist -> there should now be entries in the `archived` tab, and students in other tabs who also appear in `archived` should have an exclamation icon in the row
15. Additionally, check that the instructor cannot attempt to contact students from any tab besides the `new` tab and cannot attempt to resolve students from any tab besides the `new` and `contacted` tabs.

## Types of changes

## Checklist:

Since the watchlist combines instances for a particular student into a single row, the metrics notification badge - which currently counts the number of instances - can be misleading. We may want to change the endpoint for getting the number of  new instances to instead get the number of students with new instances.
Additionally, the refresh button still needs to update the "last updated" date and use throttle/debounce to prevent too many refreshes. 
I'd also appreciate any feedback on the appearance, especially the gradebook link icon and the archived student icon, as these were not specified in the original design. The gradebook link icon was included so that the student name could be a label on the checkbox (allowing the user to click on the name in order to check the box).


app/assets/javascripts/watchlist.js
```diff
@@ -0,0 +1,431 @@
...
+      return new_instances;
+    case "contacted_tab":
+      return contacted_instances;
+    default:
+      console.log(`${tab} is not a valid tab for this action`)
+      return;
+  }
+}
+
+// TODO: update logic here when contacting / resolving students
```
ashleyzhang, line 374: I wasn't too sure what to update with this function. It's not being used right now (as far as I know), so I was thinking about just removing it. If someone sees a purpose for it though, please let me know!

victorhuangwq:
During testing, my get_current_metrics endpoint returned a null for some of the violations parameters. While this is unexpected behavior, the front end was not able to fail safely (by ignoring the null), and did not render the other watchlist instances.

It would be important to carry out null checks for the violation parameters. Specifically, for the case where it failed for me was at 

image

A quick & simple (but not as readable) way to do null checks in  javascript is with [nullish coalescing operator ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) together with [optional chaining ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)

which in this specific case would be like

``condition_string = '${Object.keys(violations ?? {})?.length} no submission`;``

1. so the `violations ?? {}` checks if violation is null, if it is null, it uses {} instead
1. the ?.length means that if the first part is null, it does not call the length and returns null instead.

This is a common thing in javascript programming however, else you could do normal JSON checks as well.

victorhuangwq:
Downward drop is within a category, so it would be visually important to separate them into different sections. Else it would be hard for the instructors to identify the issue properly.
image

A suggestion would be having a simple horizontal line divider between homework, quiz and lab

victorhuangwq:
There is a feature regression due to the addition of the watchlist, in the metrics selection form.  

The risk metrics code, in metrics.js at line 28:
`$('.checkbox').change(function(){`
detects change based on all checkboxes. Which includes the one on the watchlist as well.

This causes the risk metrics form to perform sluggishly. 

Suggestion: Make the .checkbox function more specific to the one in risk metrics, by adding new classes possibly.

victorhuangwq:
**What I have tested:**
1. No risk metrics: Shows "you have not defined your risk metrics" appropriately ✅
1. Added many risk metrics to a course full of data.
1. Clicked on the user link out. Leads to the correct user profile  ✅
1. Checked the "downward trend", "no submission", "low score" and "grace days" label.

- Downward trend label is okay. ✅
- No submission is constantly **1 no submission**, despite there being more than 1 no submission when hover over. I think it's because there are always only one object, but the one object is an array of no submissions.
- Low score trend label is okay (correct number) ✅
- Grace days used is not being show. So I went to check the code and I realized that it is currently being rendered as low score:

```
      case "grace_day_usage":
        condition_string = `${Object.keys(violations).length} low score`;
        var violation_list = [];
        $.each(violations ?? {}, function( lab, val ) {
          violation_list.push(`${lab}: ${val}`);
        });
        violation_string = violation_list.join(" | ");
        break;
```
5. Tested Individual Contact and Resolve ✅
6. Tested Batch Contact and Resolve ✅
7. Archived content has a info button at the end ✅

app/assets/javascripts/watchlist.js
```diff
@@ -0,0 +1,431 @@
...
+            $.each(dict, function( lab, val ) {
+              inner_list.push(`${lab}: ${val}`);
+            });
+          });
+          violation_list.push(`${inner_list.join(" | ")}`);
+        });
+        violation_string = violation_list.join(" | ");
+        break;
+      case "low_grades":
+        condition_string = `${Object.keys(violations).length} low score`;
```
victorhuangwq, line 59: null check here

app/assets/javascripts/watchlist.js
```diff
@@ -0,0 +1,431 @@
...
+        var violation_list = [];
+        $.each(violations, function( lab, val ) {
+          violation_list.push(`${lab}: ${val}`);
+        });
+        violation_string = violation_list.join(" | ");
+        break;
+      case "no_submissions":
+        condition_string = `${Object.keys(violations).length} no submission`;
```
victorhuangwq, line 67: null check here

app/assets/javascripts/watchlist.js
```diff
@@ -0,0 +1,431 @@
...
+        var violation_list = [];
+        violations.no_submissions_asmt_names.forEach(val => {
```
victorhuangwq, line 69: null check here, can use optional null chaining

app/assets/javascripts/watchlist.js
```diff
@@ -0,0 +1,431 @@
...
+          violation_list.push(val);
+        });
+        violation_string = violation_list.join(" | ");
+        break;
+      case "grace_day_usage":
+        condition_string = `${Object.keys(violations).length} low score`;
```
victorhuangwq, line 75: null check here

app/assets/javascripts/watchlist.js
```diff
@@ -0,0 +1,431 @@
...
+function get_watchlist_function(){
+  var new_instances = {}
+  var contacted_instances = {}
+  var resolved_instances = {}
+  var archived_instances = {}
+
+  var selected_user_ids = []
+
+	$.getJSON(watchlist_endpoints['get'],function(data, status){
+	    if(status=='success') {
```
victorhuangwq, line 170: there should preferably be an error situation as well

app/assets/javascripts/watchlist.js
victorhuangwq: I think this was a boiler plate by @amartinez116.  Seeing that you have written your own update_watchlist it should be fine?

app/views/metrics/index.html.erb
```diff
@@ -41,65 +34,20 @@
...
+        <div class="ui tab segments" data-tab="resolved_tab", id="resolved_tab"></div>
+        <div class="ui tab segments" data-tab="archived_tab", id="archived_tab"></div>
+      </div>
+      <div class="ui tab segments active" id="undefined_metrics">
+        <center>
+          <i class="huge inbox icon"></i> 
             <h3> You have not defined your risk metrics </h3>
             Set up identification conditions at in the <b>Risk Metrics</b> tab. <br> 
             Identified possible at-risk students will appear here. 
-          </center>
```
victorhuangwq, line 100: I feel that it should not default to no metrics state, because it will be shown while the api is being called, confusing the user.

victorhuangwq (changes requested):
Great work handling the complexity of the task! 

I have tested and left appropriate comments in this thread. Let me know when you have addressed them and I will be happy to re-review them!

app/assets/javascripts/watchlist.js
amartinez116: ah yes, I wrote this here to help your PR out in updating students in the watchlist when the instructor contacts/resolves them.

amartinez116: oh wait after taking a closer look: this was the old version of the update_watchlist function. The correct version is the one finalized at the bottom. Line 393. Please use that one and make sure to delete lines 374-391.

ashleyzhang: Got it, thanks for letting me know! I removed it

app/assets/javascripts/watchlist.js
ashleyzhang: Made the error case show the same banner as when the POST requests don't work.

app/views/metrics/index.html.erb
ashleyzhang: Made this hidden by default with `display:none`

app/assets/javascripts/watchlist.js
```diff
@@ -0,0 +1,420 @@
...
+	});
+}
+
+// instructor clicks on 'refresh' button
+$('#refresh_btn').click(function(){
+	refresh_watchlist();
+});
+
+function refresh_watchlist(){
+	$.getJSON(watchlist_endpoints['get'],function(data, status){
```
xinyis991105, line 415: Should this be watchlist_endpoints['refresh']? I was trying to click the refresh button and it only seems to re-obtain the current watchlist instances.

xinyis991105:
Tested according to the PR description:
- Go to the watchlist tab and students appear in the new tab
- Students' outgoing gradebook link links to the correct student's grades
- Hover over the condition tags and the correct information about assignments appears (For the downward trend, each category now has a separate line)
- Contact button for individual students work correctly. Default mailing app pops up and student is then moved into contacted tab.
- Resolve button for individual students work correctly. Student is moved into resolved tab.
- Contacting multiple students works as expected. The students' email addresses are put into the recipient of a single email.
- Resolving multiple students (both from new and contacted) works as expected.
- Select-all works as expected
- When all instances are resolved, on the main course page, the metrics button no longer has a notification badge
- Changing a risk metric correctly refreshes the instances and archived instances are now in archived tab. Students who've been archived in the new tab now has an exclamation mark.
- Contact is disabled for contact tab. Contact and resolve are disabled for resolve tab.

I've left a comment regarding the refresh button in the code.
Probably nit-picking but the watchlist doesn't seem to handle resizing of browser well. When a student violates too many conditions and has been archived before, this could be a problem.
Screen Shot 2020-12-15 at 10 18 44 AM
This is actually an edge case. Let me know what you think of it!

app/views/metrics/index.html.erb
victorhuangwq: no longer displays! :)

victorhuangwq (approved):
@xinyis991105 thanks for the great catch and review!

I revisited the code and tested the features again, and I think we are good to go. 
I have made the small fix of changing the endpoint from get to refresh to address the problem that Xinyi found!

victorhuangwq (changes requested):
@ashleyzhang I apologize, but after further testing, I realized that after your new changes the resolve and contact button are no longer working, at least on my setup.
- Resolve and Contact Individually
- Resolve and Contact in batch.
Both of them do not result in the intended change (nothing happens). 

Could you take a look and test them out?

app/assets/javascripts/watchlist.js
ashleyzhang: Thanks for catching that!

ashleyzhang:
> @ashleyzhang I apologize, but after further testing, I realized that after your new changes the resolve and contact button are no longer working, at least on my setup.
> 
> * Resolve and Contact Individually
> * Resolve and Contact in batch.
>   Both of them do not result in the intended change (nothing happens).
> 
> Could you take a look and test them out?

The problem was that it was calling `refresh_watchlist` to update the watchlist on every resolve and contact, which takes too long before actually calling `get_watchlist_function`. I just changed it so that it calls `get_watchlist_function` directly without trying to refresh first.

With regard to handling resizing, I couldn't find a quick solution, so I figured that since it's an edge case we can handle it later if it becomes a more prevalent issue? I could also make the spacing a little bit smaller so that it can fit the four tags. If anyone has thoughts on this please let me know.

victorhuangwq:
Thanks for the quick work on the issue! 
I have checked the resolve and contact button, and they are both working as expected 👍 

I agree that there is an issue with regards to the resizing, but I do also believe that it's not as likely an issue, thus we would be able to put that issue aside for now. I have added an issue in our sprint backlog for that.

app/assets/javascripts/watchlist.js
```diff
@@ -0,0 +1,421 @@
...
+		},
+		error:function(result, type){
+			render_banner({
+				type:"negative",
+				header:"Currently unable to " + method + " students",
+				message: "Do try again later",
+				timeout: -1
+			});
+		},
+		complete:function(){
```
victorhuangwq, line 405: I think it would be better to remove this empty function

victorhuangwq (approved):
LGTM! 👍

PR #1270, opened 2020-12-17
Enhances Watchlist Refresh Button 

victorhuangwq:
## Description

- Updates Last Updated Time based on watchlist entry with the greatest last updated time
- Adds state handling to the refresh button (loading state, success and error messages). You can't press on the refresh button again if it is currently loading.

## Motivation and Context

This is as required as part of the metrics feature. The main important aspect of this is to update the last updated time accordingly, and to prevent refresh from being pressed multiple times in one go.

## How Has This Been Tested?

To check last updated time
- Pressed the refresh button. The last updated time should change
- Change the risk metrics. The last updated time should change

To check refresh button state
- Pressed the refresh button, try clicking on it while loading. It should only trigger it once (thus only showing only one success banner).
- Changed the refresh endpoint to an invalid one, the refresh call should fail and show an appropriate error message. 

**Success:**
image

**Failure**
image

## Types of changes

### DIFF
--- a/app/assets/javascripts/metrics.js
+++ b/app/assets/javascripts/metrics.js
@@ -176,6 +176,7 @@ $('#save').click(function(){
 				header:"You have successfully saved your conditions",
 				message:"Your watchlist should reflect your new conditions",
 			});
+			get_watchlist_function();
 		},
 		error:function(result, type){
 			render_banner({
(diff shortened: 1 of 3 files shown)


ashleyzhang (approved):
LGTM!

- Clicked refresh button -> refresh endpoint was called, success banner showed, 'last updated' changed accordingly
- Tried clicking refresh button multiple times while it was loading -> refresh endpoint was only called once
- Changed refresh endpoint to an invalid one -> error banner showed

The fomantic 'loading' class is a nice touch! Definitely useful for showing progress.

PR #1276, opened 2021-01-11
Change null coalescing and null chaining operators to lodash alternatives

victorhuangwq:
## Description

Changes Null Coalescing and Null Chaining Operators into lodash alternative (https://lodash.com/docs/4.17.15#get)

## Motivation and Context

This is a required as Uglifier, the javascript precompiling library we are using, do not currently have support for the two operators as mentioned.

This change utilizes the lodash equivalent, which is correct but slower.

## How Has This Been Tested?

- Ensured that it is functionally similar when using the metrics page (everything renders as per usual)
- Ran `RAILS_ENV=production bundle exec rails assets:precompile` to ensure that it precompiles

## Types of changes

### DIFF
--- a/app/assets/javascripts/metrics.js
+++ b/app/assets/javascripts/metrics.js
@@ -76,35 +76,35 @@ $.getJSON(metrics_endpoints['get'],function(data, status){
 
 		// switch case to check each checkbox field
 		data.forEach(condition => {
-			switch(condition?.condition_type){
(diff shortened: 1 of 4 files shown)


fanpu (approved):
Looked through the code, learnt a few new behaviors about js and tested that the asset precompilation step works. LGTM!

PR #1285, opened 2021-02-28
Request for institution name in the bug template

victorhuangwq:
## Description

This adds a line in the bug report template, as knowing which institution's Autolab we are talking about provides lots of valuable context in helping us identify and address the issues.


xinyis991105 (approved):
LGTM

PR #1291, opened 2021-03-12
Fix infinite speedgrader redirects when header_position = 0

oliverli:
## Description

Change "is defined" logic from `if params[:header_position]` to `if params.include?(:header_position)`. The former can return `false` if `header_position` is defined with a value of zero.

The change from `Archive.archive? @filename` to `Archive.archive?(@submission.handin_file_path)` is immaterial, and is only done for consistency with subsequent lines, allowing a dev to trace the if else logic a bit more clearly.

## Motivation and Context

It was reported that viewing a tar submission in this speedgrader bug was causing "not redirecting properly" errors on multiple browsers (causing by infinite redirection loops).

## How Has This Been Tested?

Tested by opening tars with PDFs and C files in the speedgrader, and verifying that the infinite loop bug with `header_position = 0` disappears, auto redirection when `header_position` is undefined works, and no major issues arise when an invalid value is specified for `header_position`.

## Types of changes

## Checklist:

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -297,7 +297,7 @@ class SubmissionsController < ApplicationController
       }]
     end
 
-    if Archive.archive? @filename && params[:header_position]
+    if Archive.archive?(@submission.handin_file_path) && params.include?(:header_position)
       file, pathname = Archive.get_nth_file(@submission.handin_file_path, params[:header_position].to_i)
 
       if(file.nil?)


ashleyzhang (approved):
LGTM except maybe one thing to look out for: it seems to be failing for [this submission](https://autolab-dev.andrew.cmu.edu/courses/15213-f27/assessments/testassignmentforinc0580430/history?cud_id=1239) on autolab-dev that links to header_position=2, which may suggest that there's a different issue unrelated to header_position being 0? This is definitely a worthwhile fix anyway though. Thanks for the quick investigation!

PR #1295, opened 2021-03-15
Consider Duplications When Refreshing

xinyis991105:
## Description
This PR considers duplications of watchlist instances when refreshing a watchlist.

## Motivation and Context
Currently, if an instructor resolves/contacts some students on the watchlist and hits refresh button just for fun, those who have been resolved/contacted reappear on the watchlist. If the resolved/contacted student does not add any fresh information to the watchlist, we don't want to add them back.

## How Has This Been Tested?

1. Set up metrics for a course and check that watchlist instances are initialized (through refreshing the watchlist).
2. Randomly resolve or contact some students. Hit refresh again. The watchlist should remain the same, without the resolved/contacted students coming back.
3. Change your metrics and come back and check the watchlist has been correctly refreshed.

## Types of changes

## Checklist:

As of now, the callbacks set up for refreshing individual students' watchlist status only consider conditions relevant to that callback, e.g. when there is a score change for the student, the callback only considers low grades and downward trend conditions. Suppose previously, an instructor has resolved/contacted this student on the grace day condition, then the score callback is not responsible for bringing the grace day condition instance back. Since we bring back all of a single student's instances when refreshing, I was wondering whether this would be considered an inconsistency.

### DIFF
--- a/app/models/watchlist_instance.rb
+++ b/app/models/watchlist_instance.rb
@@ -32,19 +32,22 @@ class WatchlistInstance < ApplicationRecord
     end
 
     # update!
-    # archive previous watchlist instances first
-    for instance in current_instances
-      instance.archive_watchlist_instance
-    end
+    new_instances = []
     # now check current conditions
     if current_conditions.length == 0
       # case 1: no current risk conditions exist
-      return []
+      # no-op
     else
       # case 2: current risk conditions exist
-      new_instances = self.add_new_instances_for_conditions(current_conditions, course)
-      return new_instances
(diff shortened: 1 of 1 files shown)


amartinez116:
I think the issue of only considering conditions relevant to that callback is something we'd want to address. In the example you gave, since the student's score was updated and now they meet the low score requirement, they should still have the grace day tag to remain consistent. Do you think this is something that is outside the scope of this PR? We can separate it into a different task if so.

xinyis991105:
I think we can separate that into another task. The callbacks are all over the place and probably require some more concentrated efforts.

amartinez116 (approved):
LGTM. Fixes duplication. Will address callback logic in different PR.

PR #1297, opened 2021-03-15
Fixes Bulk Resolve and Resolving Quickly Throwing Errors

victorhuangwq:
## Motivation and Context

Currently, when you 
1.  Bulk resolve many instances (~> 5) after doing a refresh it is sure to have an error
2.  Resolve using the individual buttons too quickly, it will also throw an error

## Description

image

The two issues mentioned above are actually two unrelated issues

The first issue is because the function `get_watchlist_function` is called whenever changes are made (update_watchlist, refresh_watchlist). Each time it is called, it uses jquery's `click` helper to add a new function to the resolve and contact buttons. However, the click function does not override the existing function, but appends the new behavior. As such the previously binded function is called alongside the newly binded ones. This causes multiple calls to `update_watchlist_function`.

The second issue is because the single resolve buttons are not disabled after pressing, so pressing quickly will just register twice. This is fixed simply by disabling it, and re-enabling it if there's an error.

## How Has This Been Tested?

For bulk resolve
1. Have instances in the watchlist
2. Refresh watchlist instance (this is to get `get_watchlist_function` to be called again
3. Select all instances and click resolve
4. Repeat the same for contact as well

For quick resolve
1. Double click / triple click on the individual resolve buttons
2. Repeat the same for individual  contact button as well

No errors should be thrown for both

## Types of changes

## Checklist:

### DIFF
--- a/app/assets/javascripts/watchlist.js
+++ b/app/assets/javascripts/watchlist.js
@@ -298,20 +298,40 @@ function get_watchlist_function(){
       $('.ui.circular.label.condition').popup();
 
       $('.ui.button.contact_single').click(function() {
-        $(this).prop('disabled', true);
+
+        // disable all action buttons
+        var button_group = $(this).parent().find('button');
+        button_group.prop('disabled', true);
+
         method = "contact";
         var user_id = $(this).parent().parent().attr('id');
         window.open(`mailto: ${new_instances[user_id]["email"]}`, "_blank");
         console.log(new_instances[user_id]["instance_ids"]);
-        update_watchlist(method, new_instances[user_id]["instance_ids"],this);
+        
+        // re-enabling buttons on failure
+        function enable_buttons () {
+          button_group.removeAttr("disabled");
+        } 
+
+        update_watchlist(method, new_instances[user_id]["instance_ids"], enable_buttons);
       })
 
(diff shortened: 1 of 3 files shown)


xinyis991105 (commented):
Regarding the bugs we found, LGTM. But as I randomly clicked around, some other unexpected behavior ensued. Specifically, if I click resolve for a single student and quickly click contact for that same student, my email application does pop up but an error message saying "unable to contact student" also shows up. I believe this is because that person is resolved before getting contacted and contacting a resolved person is not legal in our API. Should we disable both buttons when we contact/resolve a person?

victorhuangwq:
I agree that the both buttons should be disabled. I have pushed a new fix that should address the issue

xinyis991105 (approved):
LGTM!
One small small nit-picking thing: When the reload process is extremely slow (like on my local end), clicks for configuration 1 might be carried to the refreshed configuration. For example, if I click resolve for student 1 and while both buttons are disabled for student 1 I click on student 1's disabled contact button, then that click is registered for contacting student 2 in the refreshed configuration (assuming student 2 comes below student 1 originally). I do believe this is a corner case as on the product end refreshing should be fast so I wouldn't worry too much about it for now.

PR #1298, opened 2021-03-16
Hide metrics button in course home page from students

oliverli:
## Description

Change metrics button auth level logic in course index page to instructors only.

## Motivation and Context

Students were able to see the metrics button in the course home page, although they will be hit with an authentication error upon actually entering the metrics page. The button should be hidden from students altogether.

Note: I have removed the button for course assistants as well since the auth level defined in `metrics_controller.rb` is instructors only, excluding course instructors. If we decide to give CAs access to the metrics feature as well then it should come as part of another larger sprint task.

## How Has This Been Tested?

Tested using the "Act as user" feature in Manage Course, for CAs and students.

## Types of changes

## Checklist:

### Addendum
The Gradebook button logic in the navbar has also been updated to be consistent with the Gradebook button in an assignment

### DIFF
--- a/app/views/assessments/index.html.erb
+++ b/app/views/assessments/index.html.erb
@@ -54,13 +54,16 @@
        <h3 style="border-bottom: 4px solid #dddddd"></h3>
           <p style="font-size: 18px">Section: <b><%= "#{@cud.section}" %></b></p>
           <p style="font-size: 18px">Number of grace days remaining: <b><%="#{@cud.global_grace_days_left}"%></b></p></br>
+
           <% if @course.website.nil? || @course.website.eql?("") %>
-          <span class="btn btn-large disabled"><%= "<i class='material-icons left'>web</i> Course Website".html_safe %></span>
+            <span class="btn btn-large disabled"><%= "<i class='material-icons left'>web</i> Course Website".html_safe %></span>
           <% else %>
-          <%= link_to "<i class='material-icons left'>web</i> Course Website".html_safe, @course.website, target: "_blank", title: "Go to course website", :class => "btn btn-large red darken-3" %>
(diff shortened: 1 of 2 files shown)


app/views/assessments/index.html.erb
```diff
@@ -54,13 +54,16 @@
...
           <p style="font-size: 18px">Number of grace days remaining: <b><%="#{@cud.global_grace_days_left}"%></b></p></br>
+
           <% if @course.website.nil? || @course.website.eql?("") %>
-          <span class="btn btn-large disabled"><%= "<i class='material-icons left'>web</i> Course Website".html_safe %></span>
+            <span class="btn btn-large disabled"><%= "<i class='material-icons left'>web</i> Course Website".html_safe %></span>
           <% else %>
-          <%= link_to "<i class='material-icons left'>web</i> Course Website".html_safe, @course.website, target: "_blank", title: "Go to course website", :class => "btn btn-large red darken-3" %>
+            <%= link_to "<i class='material-icons left'>web</i> Course Website".html_safe, @course.website, target: "_blank", title: "Go to course website", :class => "btn btn-large red darken-3" %>
           <% end %>
-          <% if @cud.instructor? || !(@cud.course_assistant? && !@cud.section.blank?) %>
```
amartinez116, line 62: In the previous logic, we have set it up so CA's that have an assigned section cannot view the gradebook. I believe this is because some instructors prefer to give certain CA's limited visibility (i.e only have access to grading their section and not viewing all of the students grades). So, I think this logic is still needed for the gradebook.

amartinez116 (changes requested):

amartinez116 (dismissed):
LGTM. Works as expected

amartinez116 (approved):
LGTM! Tested and works. Also nice catch with the navbar.

PR #1300, opened 2021-03-16
Send BCC on emailing to more than 1 student from metrics

oliverli:
## Description

When contacting multiple students using the metrics feature, email students in a mass BCC list to ensure student privacy (so that students do not know the identities of other contacted students).

## Motivation and Context

Alleviates student privacy concerns

## How Has This Been Tested?

Tested by selecting 0, 1, 2, and 10+ students in the metrics feature.

Found some bugs and possible issues in metrics:

* When 0 students are selected, every single previous selection opens `mailto` again (and `update_watchlist`) is called for each of them which results in an error banner message. The called instances should be cleared after each successful contact action
* If a user is using an external application (e.g. Outlook) as their default mail client to handle `mailto` links, clicking contact results in a new `about:blank` page as we are opening the link as a popup, which can be annoying. I tried to make the local tab the target of the `mailto` link itself, but a user using a web application as the mail client will be redirected there instead of a popup. Seems to be a catch-22
* When a popup blocker is enabled and blocking the links, an error message is displayed. Perhaps a more targeted "please disable your popup blocker" message is in order?

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/app/assets/javascripts/watchlist.js
+++ b/app/assets/javascripts/watchlist.js
@@ -326,7 +326,10 @@ function get_watchlist_function(){
       instance_ids = instance_ids.concat(new_instances[user_id]["instance_ids"]);
     });
   
-    if (instance_ids.length > 0) {
+    if (instance_ids.length > 1) {
+      window.open(`mailto: ?bcc=${emails}`, "_blank");
+      update_watchlist(method, instance_ids);
+    } else if (instance_ids.length > 0) {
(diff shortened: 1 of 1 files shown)


ashleyzhang (approved):
LGTM

Note: discussed in meeting about bugs found in test plan - the first issue about the error banner is being addressed in another PR, the external application problem seems relatively minor, and the popup blocker error message should be addressed in the future.

PR #1301, opened 2021-03-16
Contact and Resolved Button Not Greyed Out when there are no students

amartinez116:
Updated the logic of graying out buttons when there are not any students in that specific tab.

## Description

I moved the tab switching logic into its own function. Now, we check if that tab has any students. If there are not any students then we have to make the buttons not active. If there are students, then we show the relevant buttons. 

## Motivation and Context

(Before)
image

(After)
image

## How Has This Been Tested?

Tested this locally by having empty tabs with no students along with other tabs with students. Logic should still work when moving students from one tab to the other. Example: Moving students from new --> contact should update the buttons visibility if the tab goes from empty to nonempty (or vice versa, non empty to empty). 

## Types of changes

## Checklist:

### DIFF
--- a/app/assets/javascripts/watchlist.js
+++ b/app/assets/javascripts/watchlist.js
@@ -169,7 +169,7 @@ function get_watchlist_function(){
   var selected_user_ids = []
 
 	$.getJSON(watchlist_endpoints['get'],function(data, status){
-	    if(status=='success') {
+	    if (status=='success') {
 	    	var new_empty = 1;
 	    	var contacted_empty = 1;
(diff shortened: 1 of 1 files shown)


xinyis991105 (approved):
LGTM!
Tested by
1) Resolving all students in the new tab. Both buttons are greyed out as expected.
2) After refreshing new instances (either through metrics change or the buggy refresh button), both buttons are enabled again.
3) Changing one tab from another. The buttons are greyed out or enabled according to its tab information as expected.
4) Resolving all students in contact tab. The resolve button is disabled as expected.

PR #1303, opened 2021-03-23
Remove duplicates of archived instances

ashleyzhang:
## Description
This PR fixes a bug in the metrics watchlist where archived instances are duplicated in other tabs. The expected behavior is that when an instructor changes the metrics, the watchlist refreshes and moves all previous contacted and resolved instances to archived. The contacted and resolved tabs should then be empty until the instructor takes further action.

The existing code does not consider that archived instances still have a status of contacted or resolved. The fix is to add an `else` block so the new, contacted, and resolved tabs are only populated if the current instance is not archived.

## Motivation and Context
Part of a series of bug fixes for the new metrics feature.

## How Has This Been Tested?
When the metrics are changed and saved, the contacted and resolved tabs should be empty. The archived tab should contain all instances previously in contacted and resolved. The contacted and resolved tabs should only populate after clicking contact or resolve on any new instances.
image
image
image
Instances in the new tab should reflect the changed metrics.

## Types of changes

## Checklist:

### DIFF
--- a/app/assets/javascripts/watchlist.js
+++ b/app/assets/javascripts/watchlist.js
@@ -201,23 +201,24 @@ function get_watchlist_function(){
           if (_.get(watchlist_instance,'archived')) {
             archived_empty = 0;
             addInstanceToDict(archived_instances, id, user_id, course_id, user_name, user_email, condition_type, violation_info);
-          }
-          switch(_.get(watchlist_instance,'status')){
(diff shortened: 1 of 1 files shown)


victorhuangwq (approved):
LGTM. Straightforward fix.
Tested and the watchlist items are no longer appearing twice.

PR #1139, opened 2020-04-29
(Dark) Theme implementation

tamnguyen98:
This is a feature I've been working on for my school (Washington State University - Vancouver) version of Autolab. I have communicated with you guys on Slack about this feature and it seems like you guys may be interested in implementing this feature on your repository. You may wonder why there are a lot of changes being erb's file and the reason was that I had to make it work with the main scss stylesheet. 

Summary of this pull request:

- Ability to change the theme from navbar, and account settings
- All visible CSS (like style and navbar) files reference the _theme.css color variables
- All color changes should only need to be made in _theme.css (follow the dark class layout)
- To minimize the number of files to be edited for a future theme, style.css.scss overrides any CSS that is from a third-party tool (such as slickgrid)
    - All tables should have the same (roughly) theme

Minor edits to some layouts consist of:
- Added a divider under the 'Courses' text
- Change the viewFeedback page to output code in a code block
- Fixed some erb's incorrect HTML tags and surrounding text in text tags (e.g., i, p, ref, etc.)
- Remove in-page static style attributes, such as "red" and "white" class (this overrides the theme)

Some things to consider:
- Changing the _theme.css visible variable's name to something meaningful instead of hex values

PR #1306, opened 2021-03-28
Migrations for git submission

fanpu:
## Description
This adds relevant migrations for the Git submission feature

## Types of changes

## Checklist:

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_03_24_223858) do
+ActiveRecord::Schema.define(version: 2021_03_28_201958) do
 
   create_table "annotations", force: :cascade do |t|
     t.integer "submission_id"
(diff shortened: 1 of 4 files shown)


db/schema.rb
```diff
@@ -334,7 +338,6 @@
     t.string "school"
     t.string "major"
     t.string "year"
-    t.string "theme", default: "default"
```
fanpu, line 337: The theme field was erroneously introduced into the users schema based on https://github.com/autolab/Autolab/pull/1234/files#diff-cbf8b25d6d853acf58fa9057841f407d29ffe90649c75cf9e3f51b2d6e3aa1d3R337. I believe the reason was because Xinyi ran the migrations for https://github.com/autolab/Autolab/pull/1139/files#diff-e625c8fa2a7f730fcafe9acb6d8bf1c1fba45511b896c86185cf4d06c0bc354eR3 when reviewing the PR and committed the schema as well by accident. But in any case, this will remove this field

db/schema.rb
```diff
@@ -154,6 +156,8 @@
     t.datetime "cgdub_dependencies_updated_at"
     t.text "gb_message"
     t.string "website"
+    t.string "git_access_key"
+    t.string "classroom_name"
```
ashleyzhang, line 160: nit: would it be clearer to name this "git_classroom_name"?

ashleyzhang (approved):
Just a small nit, but LGTM otherwise if you decide it's better as is

db/schema.rb
fanpu: Good suggestion, fixed!

PR #1308, opened 2021-03-29
Generalize Course/Assessment "Average" to allow arbitrary titles/functions

Joshlete:
* Move defaultCategoryAggregation into CourseBase
   (allows per-course overrides)

* Don't show Aggregate rows in manage course when agg name is nil
   (stream lines display when aggregate is unused/unneeded)

* courseAggregate takes two parameters to prevent key collisions

* Fix Statistics.mean doesn't work for averages

## Description

This commit changes how Autolab calculates and handles course averages and scoring. Our changes allow
new functions to be defined in the course.rb file for a given course, which can calculate assignment
category grades based on assignments, allowing for custom weights and other calculations. The overall
course average function can call these assignment category functions to generate a fully customized
weighted grade to return to the user. We also allow for specific column names for each function, and
the column will be hidden/not shown if the column name is nil. Some variables have been changed from
scalar to hash format, and we now accept two parameters to the final courseAggregate function to avoid
key collisions. 

There is also a small bugfix included, where the backup for Statistics.mean doesn't work in the 
CourseBase.rb file, which needed to be fixed before our code worked.

## Motivation and Context

This feature was requested by one of our professors so that they could better represent student grades
in the Autolab gradebook; specifically, he wanted to be able to calculate more specific weighted
averages and incorporate them into an accurate final grade.

## How Has This Been Tested?

The course aggregate feature is in-use on the WSUV production version of Autolab, on top of our 
WSUV-specific commits and a recently rebased cmu-master pull. We have also created some virtual 
development instances with CMU's master branch pulled, and only the aggregate commits on top of that
branch, to ensure that the commits work on an unmodified version of Autolab's current commits.

## Types of changes

## Checklist:

- [x] My change requires a change to the documentation, which is located at [Autolab Docs]
       - (possibly, if documentation would be helpful in features.md)

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -50,15 +50,16 @@ class CoursesController < ApplicationController
       cols[asmt.name] = final_scores
     end
 
-    # category averages
+    # category aggregate scores
     @course.assessment_categories.each do |cat|
       next unless matrix.has_category? cat
+      next if matrix.category_aggregate_name(cat).nil?
 
-      cols["#{cat} Average"] = matrix.averages_for_category cat
+      cols["#{cat} #{matrix.category_aggregate_name(cat)}"] = matrix.aggregates_for_category cat
     end
 
-    # course averages
-    cols["Course Average"] = matrix.course_averages
+    # course aggregate
+    cols["Course #{matrix.course_aggregate_name}"] = matrix.course_aggregates
 
     # calculate statistics
     @course_stats = {}
--- a/app/controllers/gradebooks_controller.rb
+++ b/app/controllers/gradebooks_controller.rb
@@ -87,15 +87,15 @@ class GradebooksController < ApplicationController
       cols[asmt.name] = final_scores
     end
 
-    # category averages
+    # category aggregates
     @course.assessment_categories.each do |cat|
       next unless matrix.has_category? cat
 
-      cols["#{cat} Average"] = matrix.averages_for_category cat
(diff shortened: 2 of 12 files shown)


oliverli:
Hi @Joshlete! I've looked through this PR a couple of times and I'm having a hard time understanding what functionality it is adding on top of the custom weighted averages feature we current have. Could you include a sample use case and `course.rb` for course aggregates?

Joshlete:
Hi @Oliverli,

Sorry for the confusion, I don’t think we were completely clear on the scope of the changes in our commits and the last WSUV team’s commits. The main feature of our Course Aggregates is customizing the headers for the graded columns; in Gradebook view, the :name value assigned by each function (including the overall course aggregate function) will appear as the display name for the corresponding category columns in the Gradebook, in both student and teacher views. 

To enable this, several backend definitions were changed so that rather than returning a scalar of the category average, the functions in course.rb instead return a hash. The hash contains both that score value and a :name value that will appear at the top of the corresponding columns in Gradebook view.

We have also largely changed the notation to refer to these scores as ‘aggregates’ rather than ‘averages’. 

There was a single BugFix in our code as well, where the statistics.mean return value and backup grade calculations in CourseBase.rb did not work in our production version, but that issue may have been resolved already.

As an example, if an instructor wants to display a category to students as “Homework Total Score” instead of “Homework Average”, changing the :name returned by that category in its hash will change the displayed name in Gradebook views, reflecting the change to the specified :name.

Thank you!

WSUV Autolab Team

PR #1310, opened 2021-03-30
Fix the bug where saving no risk conditions throws an error

xinyis991105:
## Description
Currently, on master, if you save on no risk conditions in metrics, an error is thrown.
Screen Shot 2021-03-31 at 1 53 41 AM
This PR fixes this. This PR also only considers unarchived instances as current instances when refreshing. This is because when we check for redundancy, we only consider instances with current risk condition id's. Instances with current risk condition id's are not archived by our design and only those with deprecated risk condition id's are archived. This is to filter out some unnecessary instances to load.

## Motivation and Context
This is for the integrity of the metrics feature.

## How Has This Been Tested?
A holistic testing on the refresh mechanism is recommended. Step 1 is required for this PR.
1. Select a course. Save on no risk conditions. A success message should pop up and there is no instance in the new, contacted, or resolved tabs.
2. Select some risk conditions and save. Confirm that there are expected instances in the new tab and nothing in contacted or resolved. Resolve/contact some of them. Make sure resolved/contacted instances show up in the corresponding tabs. Hit refresh without changing your risk conditions. Confirm that nothing changes in the three tabs.
3. Bulk contact all instances. Hit refresh without changing your risk conditions. Nothing should've changed.
4. Bulk resolve all instances. Hit refresh. Nothing should've changed.
5. Remove your conditions one by one. Every time after saving, make sure the tags for each student in the new tab are consistent with your conditions. Do so until you save on no risk condition again and make sure that works too. During this process, feel free to resolve/contact and refresh without changing risk conditions.
6. Add your conditions one by one. Make sure everything works as expected.
7. Change your conditions in a partitioned manner. New risk conditions do not overlap with old conditions. Save and make sure everything works as expected.
8. Change your conditions in an overlapping manner. Make sure everything works as expected.
9. Don't change your conditions but tweak the parameters of your conditions. Save and see that everything works as expected.
10. Don't change your conditions. Just uncheck and check back your old conditions. Save and see that everything works as expected.

## Types of changes

## Checklist:

### DIFF
--- a/app/models/watchlist_instance.rb
+++ b/app/models/watchlist_instance.rb
@@ -26,13 +26,14 @@ class WatchlistInstance < ApplicationRecord
     begin
       course = Course.find_by(name: course_name)
       current_conditions = RiskCondition.get_current_for_course(course_name)
-      current_instances = WatchlistInstance.where(course_id: course.id)
+      current_instances = WatchlistInstance.where(course_id: course.id, archived: false)
     rescue NoMethodError
       raise "Course #{course_name} cannot be found"
     end
 
     # update!
     new_instances = []
+    deprecated_instances = current_instances
     # now check current conditions
     if current_conditions.length == 0
       # case 1: no current risk conditions exist


victorhuangwq (approved):
Tested Locally. Saving no risk conditions no longer throws an error.

PR #1328, opened 2021-04-12
Add emails to watchlist

ashleyzhang:
## Description
This PR changes the metrics feature watchlist to add student email addresses directly below student names in small grey font.
Screen Shot 2021-04-11 at 11 00 55 PM

## Motivation and Context
Since instructors often search for students based on the student IDs associated with their emails, we want to make student emails visible in the watchlist.

## How Has This Been Tested?
The email address should always appear directly below the student's name.

## Types of changes

## Checklist:

### DIFF
--- a/app/assets/javascripts/watchlist.js
+++ b/app/assets/javascripts/watchlist.js
@@ -18,12 +18,13 @@ function get_html_empty_message(message){
       </div>`;
 }
 
-function get_name_html(name) {
+function get_name_email_html(name, email) {
   return `
       <div class="ui checkbox select_single">
         <input type="checkbox"/>
         <label>
           <p class="name_label"> ${name} </p>
+          <p class="email_label"> ${email} </p>
         </label>
       </div>`;
 }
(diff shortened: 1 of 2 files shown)


victorhuangwq:
Thanks for the quick work! Looking at the screenshot I'm wondering if it would look better if the checkbox on each instance is top down centered?

ashleyzhang:
Good point, I agree - here's how it looks now!
Screen Shot 2021-04-12 at 11 00 22 AM

victorhuangwq:
I think this looks okay. @amartinez116 can you weigh in as well?

amartinez116:
I think it looks good! 
Nit: in the gradebook we made the emails clickable and that leads to the same link as our icon (student's gradebook). Not sure if we want to pick consistency over clarity?

ashleyzhang:
> Nit: in the gradebook we made the emails clickable and that leads to the same link as our icon (student's gradebook). Not sure if we want to pick consistency over clarity?

Ooh that's definitely a good point - but I actually think it's better as is, solely because the name and email are labels on the checkbox, which means that clicking the name or email also checks the checkbox. It might be a little frustrating if going to the gradebook caused the checkbox to be clicked as well. We could change it so that clicking the name and email doesn't affect the checkbox, but that'd significantly reduce the size of the clickable area.

Although I do think it'd look better than having the current outgoing link icon for the gradebook... open to more feedback!

victorhuangwq:
My personal opinion is that this is good for now. With our watchlist instance I don't think we need them to click into the gradebook often anyways

amartinez116:
Sounds good to me, we can leave it as is.

victorhuangwq (approved):
LGTM!
Tested Locally

Issue #1166, opened 2020-05-21
Hidden files in archive with filenames beginning with . (period) do not appear on Speedgrader


oliverli:
As per title. My friend reports that these hidden files aren't being extracted for autograding as well, but I'm still trying to replicate that.

PR #1335, opened 2021-04-19
Show hidden file in speedgrader

xinyis991105:
## Description
This PR addresses #1166.

## Motivation and Context
The current master does not support showing hidden files whose file name starts with a period, which could be problematic for some instructors. This is now enabled.
(To whoever reviews this PR: But can I know why hidden files are also considered mac_bs_file in the archive library? What does the term mean in the first place? 😂 )

## How Has This Been Tested?
Submit an archive file with a hidden file in it. It should now appear alright in the speedgrader.
hidden file

## Types of changes

## Checklist:

### DIFF
--- a/lib/archive.rb
+++ b/lib/archive.rb
@@ -25,8 +25,7 @@ module Archive
         header_position: i,
         mac_bs_file: pathname.include?("__MACOSX") ||
           pathname.include?(".DS_Store") ||
-          pathname.include?(".metadata") ||
-          File.basename(pathname).start_with?('.'),
+          pathname.include?(".metadata"),
         directory: looks_like_directory?(pathname)
       }
     end


lib/archive.rb
```diff
@@ -25,8 +25,7 @@ def self.get_files(archive_path)
         header_position: i,
         mac_bs_file: pathname.include?("__MACOSX") ||
```
amartinez116, line 26: Not related to your changes but  "__MACOSX" is a folder not a file. Therefore, it still appears in the speedgrader.

image

amartinez116:
LGTM. I think they called it "mac_bs_file" because of all the extra files macs make :/ also can we make a separate task for the "__MACOSX" folder issue?

amartinez116 (approved):

lib/archive.rb
xinyis991105: Just to make sure, does it mean we don't want to display any folders in the archive that contains "__MACOSX"?

amartinez116: Yup. That's a folder that mac creates when someone archives something.

xinyis991105: Cool I'll submit another PR to fix this soon!

Issue #1357, opened 2021-05-30
Announcements not displaying correctly


fanpu:
**Your environment**
Linux, Autolab master

**Steps To Reproduce**
1. Course -> Manage Announcements

**Current behavior**
Start/end times of announcements are not displaying correctly:

![Screenshot_select-area_20210530003315](https://user-images.githubusercontent.com/9707110/120092132-551f0d00-c143-11eb-8f6a-b4e67e0c08fe.png)

**Expected behavior**
Time should display properly

**Screenshots**

Thanks @cg2v for reporting this bug

PR #1359, opened 2021-05-30
Datetime picker result display fix

fanpu:
## Description

Addresses #1357

## Motivation and Context

The bug comes from the fact that altInput is specified, but the formatting is not specified. Usually one would do it with altFormat, but in this case we already have a template from @oliverli that also shows the time zone as well, so I decided to go with that to be more descriptive and consistent. 

## How Has This Been Tested?

Announcement start/end times now render properly, now with the time zone being specified as well

Screenshot_select-area_20210530012825

I have also checked that handin start/end times datepicker values, and extension times still render properly. 

## Types of changes

## Checklist:

### DIFF
--- a/app/assets/javascripts/initialize_datetimepickers.js
+++ b/app/assets/javascripts/initialize_datetimepickers.js
@@ -13,8 +13,13 @@
       $(datetimeElts[i]).flatpickr({
         enableTime: true,
         altInput: true,
-        altFormat: "Y-m-d H:i",
         defaultDate: defaultDate,
+        formatDate: (date, format) => {
+          return moment(date).format(format);
+        },
+        parseDate: (datestr, format) => {
+          return moment(datestr, format, true).toDate();
+        },
       })
     }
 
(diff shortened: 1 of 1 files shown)


victorhuangwq (approved):
Tested Locally. 
LGTM!

fanpu:
I will look at this re:Chaskiel's email on the scheduler's datepicker having issues

victorhuangwq:
Hi @fanpu do you think this could be merged in? Seems like the recent issue with the scheduler not working could be fixed by this PR.

fanpu:
In the ticket it seems like this is not just a display issue, but also impacts whether the values are actually persisted. We'll need to test on that front as well

fanpu:
This PR appears to fix the issue:

```
2.6.6 :001 > Scheduler.last
  Scheduler Load (0.1ms)  SELECT  "scheduler".* FROM "scheduler" ORDER BY "scheduler"."id" DESC LIMIT ?  [["LIMIT", 1]]
 => #<Scheduler id: 1, action: "Test", next: "2021-09-08 15:53:00", interval: 10, course_id: 1, created_at: "2021-09-07 15:53:39", updated_at: "2021-09-07 15:53:39"> 
```

PR #1363, opened 2021-06-10
Fix initialize_user resulting in email confirmation when this was unwanted

fanpu:
## Description

Previously we were using .create in the rake task, which actually creates and stores the user in the db. .new only creates the object but does not actually store it yet, causing the devise mailing callbacks to not fire, which is what we want

## Motivation and Context

https://github.com/autolab/docker/issues/34

## How Has This Been Tested?

Newly created user is persisted, and no more error is shown

## Types of changes

## Checklist:


oliverli (approved):
Great catch, thanks Fanpu!

PR #1388, opened 2021-09-24
Github OAuth2 integration

fanpu:
### Setting Up
- Navigate to https://github.com/settings/developers in order to create a new OAuth app
- Homepage URL should be the URL that Autolab will be served on (i.e http://localhost:3000). Authorization callback URL is the URL that Github will make a callback to after authentication, which must be the homepage URL appended with `/users/github_oauth_callback` (i.e http://localhost:3000/users/github_oauth_callback)

See example:
DeepinScreenshot_select-area_20211113132623
- After registering the application, you should now have a Client ID for your application. Create a Client Secret as well, you should see the following:
DeepinScreenshot_select-area_20211113133047
- Set the GITHUB_CLIENT_ID and GITHUB_CLIENT_SECRET environment variable to your Github Client ID and Client Secret respectively when running Autolab. Alternatively, you can modify `config.x.github.client_id` or `config.x.github.client_secret` in `config/environments/development.rb` or/and `config/environments/production.rb` 
- After you restart your deployment, go to `http://localhost:3000/admin/github_integration` to verify whether the credentials have been setup correctly

## Types of changes

## Checklist:


fanpu:
TODOs from Chaskiel:

1. GithubIntegration.clone_repo does not clean up its temporary files.
It might be easier to use git archive instead of doing separate clone and tar tasks.

2. AssessmentHandin:handin appears to accept github handins even if github submission is not enabled for the assessment. I can't tell for sure that this is not handled at a higher layer, but it's not good enough for the form to not offer github. The handin handler has to reject it too if not allowed..

3. AssessmentHandin:handin needs to sanitize `params[:repo]`, as it is passed to `system`. In particular, it should not contain control characters, spaces or ;, but ideally, you should check it for alphanumeric, plus . - _ and /

fanpu:
TODOs resolved by #1432

xinyis991105 (approved):
1. The upper right corner dropdown still doesn't work on the assignment submission page.
2. Git submission, if enabled, overrides upload submission, even if the tab is switched to upload and the submit button is pressed from there.
3. Not a huge concern but if I spin up Autolab with properly configured github credentials and enable git submission for assessments, shut Autolab down, misconfigure github credentials, and then spin up Autolab again, the assessments that previously enabled git submission still have it enabled, even though on the edit assessment page, the checkbox is greyed out. I don't think this will be a common situation in production but it'd be nice if we could fix the bug before production deployment.

Otherwise, the bugs from last local testing are now gone: cancelling authorization properly redirects; checking repository size is fast.

Overall, great job guys! Let's merge this in and test on autolab-dev first!

PR #1390, opened 2021-09-25
Fix metrics index "undefined method `[]' for nil:NilClass"

oliverli:
## Description

Add check if `course.assessments.group("category_name").count().max()` is `nil` before trying to access array value, because that can be `nil` if an instructor just created a course (hence no categories) and tries to view the student metrics page

## Motivation and Context

Fixed that bug which we were made aware of by Chakiel

## How Has This Been Tested?

To replicate bug:
* create new course
* try to view student metrics page
Confirmed bug has been fixed after this change, with desired effects on the dropdown box using `@max_consecutive_assessments`, i.e. it is basically disabled

## Types of changes

## Checklist:

### DIFF
--- a/app/controllers/metrics_controller.rb
+++ b/app/controllers/metrics_controller.rb
@@ -9,7 +9,7 @@ class MetricsController < ApplicationController
 		if (course_max == nil)
 			@max_consecutive_assessments = 0
 		else
-			@max_consecutive_assessments = course.assessments.group("category_name").count().max()[1] - 1
+			@max_consecutive_assessments = course_max[1] - 1
 		end
 	end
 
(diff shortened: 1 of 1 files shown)


app/controllers/metrics_controller.rb
```diff
@@ -4,7 +4,13 @@ def index
 		course = Course.find_by(name: params[:course_name])
 		@course_grace_days = course.grace_days
 		@num_assessments = course.assessments.count()
-		@max_consecutive_assessments = course.assessments.group("category_name").count().max()[1] - 1
+		
+		course_max = course.assessments.group("category_name").count().max()
+		if (course_max == nil)
+			@max_consecutive_assessments = 0
+		else
+			@max_consecutive_assessments = course.assessments.group("category_name").count().max()[1] - 1
```
ashleyzhang, line 12: nit: could we change `course.assessments.group("category_name").count().max()` to just `course_max`?

ashleyzhang (commented):
Small nit but looks good otherwise!

app/controllers/metrics_controller.rb
oliverli: good catch! done

ashleyzhang (approved):
LGTM! Thanks for the fix

PR #1392, opened 2021-09-27
Better output for non-ascii autograder feedback

fanpu:
## Description

Improves on the feedback returned when non-ASCII output appears from the autograder (see https://github.com/autolab/Autolab/pull/1384). Many thanks to @cg2v for the enhancements.

## Motivation and Context
See https://github.com/autolab/Autolab/pull/1384

## How Has This Been Tested?
Upload something like a binary file on hellolab, you should be able to see the feedback even though it contains non-ASCII characters.

To test the encoding workflow, upload a file containing just \xc2 (see https://apidock.com/ruby/v2_5_5/String/valid_encoding%3F), and verify that we get \xc2 instead of the actual representation in the feedback 

## Types of changes

## Checklist:


oliverli (approved):
Cool approach!

PR #1402, opened 2021-10-20
More detailed roster error messages & sort by account status

michellexliu:
## Description
Updated the error messages to include more information about roster uploads and CSV parsing; added a "sort by account status" toggle that displays newly-added and newly-dropped accounts at the top of the confirmation screen.

## Motivation and Context
This updated roster upload aims to address two main issues:
1. Lack of detailed error messages - Multiple professors have had issues uploading their rosters due to a lack of detailed error messaging when a roster doesn’t satisfy Autolab’s parsing requirements.
2. New & dropped students are difficult to locate - In large roster uploads, the user would need to scroll throughout the entire roster to locate which students have newly dropped or added the course (indicated by green or red text).

## How Has This Been Tested?
(incomplete)

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ ] My change requires a change to the documentation, which is located at [Autolab Docs](https://github.com/autolab/docs)
- [ ] I have updated the documentation accordingly, included in this PR

## Other issues / help required
2 remaining rubocop errors:
```
app/controllers/courses_controller.rb:851:5: W: Lint/SuppressedException: Do not suppress exceptions.
    rescue StandardError
    ^^^^^^^^^^^^^^^^^^^^
app/controllers/courses_controller.rb:864:5: W: Lint/SuppressedException: Do not suppress exceptions.
    rescue StandardError
```
Not sure if removing the `rescue StandardError` from here will break anything, so I'd like to have someone else look at this as well.

PR #1406, opened 2021-10-28
Further roster upload validation

oliverli:
## Description

Adds further roster upload validation: all fields are required

To think about: is that a good idea?

## How Has This Been Tested?

Tested by uploaded dummy bad CSVs

## Types of changes

## Checklist:

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -548,7 +548,9 @@ private
 
     begin
       csv = detectAndConvertRoster(params["upload"]["file"].read)
+      rowNum = 0
       csv.each do |row|
+        rowNum += 1
         next if row[1].nil? || row[1].chomp.size == 0
 
         newCUD = { email: row[1].to_s,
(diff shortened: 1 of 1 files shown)


app/controllers/courses_controller.rb
```diff
@@ -560,14 +564,44 @@ def parse_roster_csv
                    section: row[10].to_s.chomp(" ") }
         cud = @currentCUDs.find do |cud|
           cud.user && cud.user.email == newCUD[:email]
+
+        if newCUD[:email].nil? || newCUD[:email].size == 0
```
michellexliu, line 568: Should we be running all of these if/elsif checks manually? Originally, the error-checking was linked to the active record validations of User (see the line `validates :first_name, :last_name, :email, presence: true' in `app/models/user.rb'), and based on that, only first_name, last_name, email, are currently required, which I believe is a reasonable requirement. I feel like checking with if/elsif could also cause problems in the future, because if we ever need to change the required fields in the model, we now have to remember to change them here too.

Also, I haven't been able to run this because I'm currently having issues with Ruby, but it looks like it currently only shows the first error that shows up in the roster. Ideally, it should return all errors so that users don't have to resubmit to check if there are any other errors after they fix the first one.

michellexliu (changes requested):
Thanks for making these updates! I think there might've been a miscommunication as to what the task should be. I've already updated the error messages to be more detailed (see [this earlier PR](https://github.com/autolab/Autolab/pull/1402/commits), so the goal of this task was to make the error display right after initial upload instead of after confirmation.

I'm not sure if you read my design doc, but for more detailed info, see [my response](https://docs.google.com/document/d/1QhyftGYiRvDD33nAoCSzjJGKUF1IySJ1QJaRt6c-nZo/edit) to Victor's comments on the first implementation milestone.

Sorry if this wasn't clear!

PR #1415, opened 2021-11-06
Prerenders previous and next submission page in submission view page

victorhuangwq:
As from our interview with Angela, we realized the next and previous page loading takes up a lot of time. This PR attempts to address that by making use of browser prerendering

## Description

In order to preload the content of the next and previous submission page, this PR makes use of Prerender Hints [Link types: prerender](https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/prerender)

This hint is only supported as of now by Chromium based browsers, but it is a useful and fast fix nonetheless [Supported Browser](https://caniuse.com/link-rel-prerender)

## Motivation and Context

As from our interview with Angela, we realized going to the next and previous page loading takes up a lot of time when grading. 

## How Has This Been Tested?

Testing locally has very minimal network latency, as such I used [Chrome's network throlling](https://www.browserstack.com/guide/how-to-perform-network-throttling-in-chrome) to simulate network effects, using **slow 3G.**

1. Loading a page. Record the time taken.
2. Wait a bit ~ 5 seconds (to simulate an instructor grading on the page, and to let the prerender happen)
2. Load the previous or next page. Observe the time taken

We should see that when loading the previous and next page, the time taken is lesser.

**Load time of previous page before adding hint**
image

**Load time of previous page after adding hint** 
image

There still is room for improvement, especially when it comes to the the loading of JavaScript libraries (somehow they are still being reloaded when going to the previous and next page)

## Types of changes

## Checklist:

### DIFF
--- a/app/views/submissions/view.html.erb
+++ b/app/views/submissions/view.html.erb
@@ -86,6 +86,7 @@
         <%= link_to(url_for([:view, @course, @assessment, @prevSubmission]), class: "valign-wrapper", id: "prev_submission_link") do %>
           <i class="material-icons" style="margin-right: 3px">arrow_back</i> <%= @prevSubmission.course_user_datum.user.email %>
         <% end %>
+         <link rel="prerender" href= <%= url_for([:view, @course, @assessment, @prevSubmission])%> />
       <% end %>
     </div>
     <div class="col s8 center-align valign-wrapper">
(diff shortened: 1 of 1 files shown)


xinyis991105 (commented):
I'm using Chrome 95 so prerender is supported. Here's the response time I get after enabling throttling (slow 3G). May I know which request I should inspect? Are there additional settings I need to make it work?
Screen Shot 2021-11-14 at 12 19 08 PM

victorhuangwq:
You should just be looking at the Load time. You might have to zoom out to see the metric
image

xinyis991105 (approved):
LGTM!
Load time stabilizes around 8 seconds under slow 3G without prerendering and around 6 seconds with prerendering. Thanks a lot!

PR #1418, opened 2021-11-11
Correct handin between git submission and file upload

ashleyzhang:
## Description
Submits file upload when there's a file, and removes file upload when switching to git submission tab.

## Types of changes

## Checklist:


fanpu (approved):
Works beautifully, tested with the following workflows:

- If student selects something for Git Submission but then decides to submit via File Upload, what was chosen for File Upload should be uploaded
- If student selects something for file upload but then decides to submit via Git Submission, what was chosen for Git Submission should be uploaded

PR #1421, opened 2021-11-13
Bug fixes and enhancements for github oauth2

fanpu:
## Description
Fixes the following bugs:

1. Proper error handling if user cancels OAuth flow

Adds the following enhancements:
1. Checks repository size before cloning it 

*NOTE*: please rollback db/migrate/20210929220050_create_github_integrations.rb (perform `rails db rollback` until the specific migration is rolled back, or do `rails db:rollback STEP=2`  if  there are 2 migrations to rollback), since I updated it to enforce uniqueness of user ID on github integrations as an additional layer of data consistency checks

## How Has This Been Tested?
1. If user submits a large repo, the repo will not be cloned and instead user will get a message that repo size is too large
2. If user cancels OAuth flow, user gets redirected to homepage instead of causing an internal server error

## Types of changes

## Checklist:

### DIFF
--- a/app/controllers/assessment/handin.rb
+++ b/app/controllers/assessment/handin.rb
@@ -35,7 +35,7 @@ module AssessmentHandin
       github_integration = current_user.github_integration
 
       begin
-        @tarfile_path = github_integration.clone_repo(params[:repo])
+        @tarfile_path = github_integration.clone_repo(params[:repo], @assessment.max_size * (2 ** 20))
       rescue StandardError => msg
         flash[:error] = msg
         redirect_to(action: :show)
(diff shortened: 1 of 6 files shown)


cg2v:
GithubIntegration.clone_repo does not clean up its temporary files.

It might be easier to use [git archive](https://git-scm.com/docs/git-archive) instead of doing separate clone and tar tasks.

cg2v:
AssessmentHandin:handin appears to accept github handins even if github submission is not enabled for the assessment. I can't tell for sure that this is not handled at a higher layer, but it's not good enough for the form to not offer github. The handin handler has to reject it too if not allowed..

cg2v:
AssessmentHandin:handin needs to sanitize `params[:repo]`, as it is passed to `system`. In particular, it should not contain control characters, spaces or ;, but ideally, you should check it for alphanumeric, plus . - _ and /

ashleyzhang (approved):

fanpu:
@cg2v Thanks for your comments, we will resolve them on the main branch (`oauth2-gh`)!

PR #1430, opened 2021-11-20
Encrypt Github token, add some docs for gh integration setup, using .env for sourcing application secrets

fanpu:
## Description

This PR adds several changes:

1. Encryption of Github tokens, so that in the event that database is compromised the tokens are safe since only the ciphertext is stored. 
2. Using `dotenv` Gem for storing/tracking secrets. This is because we now have both Github, Lockbox, and Devise secrets, and it seems silly to have to force users to modify configs in 3 different places manually. This results in the following changes:
   -  `./bin/initialize_secrets.sh` script for generating application secrets
   - Deprecation of `devise.rb.template`. We will just use `devise.rb` now, because there are no longer secrets stored in the file, since we are now sourcing the devise secret using environment variables
3. Update docs to reflect the changes above, and to also add instructions on how to update their Github application secrets

## Motivation and Context
Provide additional security for users, better secrets management

## How Has This Been Tested?

This change requires database rollback, as I modified the github_integration migration:

`rails db:rollback STEP=2`

`rails db:migrate`

1. Connect account to Github
2. In Rails console, do the following:
  - Verify that `GithubIntegration.last.access_token` gives usable token, check this by cloning any private repo you own using the access token
  - Verify that `GithubIntegration.last.access_token_ciphertext` gives much longer ciphertext, and check that it is not a valid token by doing the same as above
  - Check that only the ciphertext is stored in database. For the case of SQLite, you can do the following:
    - Run `sqlite3`
```
sqlite> .open db/autolab_development # replace with your database name
sqlite> select * from github_integrations;
1||5dBiqSWoFJiXGRtewGRpiOmMPcMUokqri0ey8xv452DuYvjiCpZLPB60/F2SsoVmGW3k6tIbSKJZP+L3nlimvwMHOQg=|54|2021-11-20 18:23:11.654143|2021-11-20 19:52:00.524538
```
From above we see that only the ciphertext is stored, since it's clearly a base64 encoded string and not a token.

## Types of changes

## Checklist:


xinyis991105 (approved):
LGTM!
Followed the testing steps:
1. Access token retrieved through rails is plaintext and can be used to access repo
2. Only the cipher text is stored in the database

PR #1436, opened 2021-12-02
Fixed bugs discovered in internal testing

michellexliu:
## Description
1. Previously dropped students can now be undropped if they are added back to the course via another roster upload
2. Fixed unintelligible error message “wrong number of arguments (given 2, expected 0..1)” that appears when new student gets created (incl. blank emails)

## How Has This Been Tested?
See internal testing spreadsheet

## Types of changes
- [x ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -572,6 +572,7 @@ private
                                                   grade_policy: new_cud[:grade_policy],
                                                   lecture: new_cud[:lecture])
         existing.assign_attributes(params.permit(:lecture, :section, :grade_policy))
+        existing.dropped = false
         existing.save(validate: false) # Save without validations.
       end
       rowNum += 1
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -141,8 +141,8 @@ class User < ApplicationRecord
     user.password_confirmation = temp_pass
     user.skip_confirmation!
 
-    # Rails.logger.debug("user email: ", user.email)
-    # Rails.logger.debug("user pswd: ", user.password)
+    Rails.logger.debug("user email: #{user.email}")
+    Rails.logger.debug("user pswd: #{user.password}")
 
     user.save!
     user
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -141,8 +141,8 @@ class User < ApplicationRecord
     user.password_confirmation = temp_pass
     user.skip_confirmation!
 
-    Rails.logger.debug("user email: ", user.email)
-    Rails.logger.debug("user pswd: ", user.password)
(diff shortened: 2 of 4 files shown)


app/models/user.rb
```diff
@@ -141,8 +141,8 @@ def self.roster_create(email, first_name, last_name, school, major, year)
     user.password_confirmation = temp_pass
     user.skip_confirmation!
 
-    Rails.logger.debug("user email: ", user.email)
-    Rails.logger.debug("user pswd: ", user.password)
+    # Rails.logger.debug("user email: ", user.email)
```
victorhuangwq, line 144: This was there before I edited it. Could you see if there's a way to log this without throwing the error?

app/controllers/courses_controller.rb
```diff
@@ -572,6 +572,7 @@ def write_cuds(cuds)
                                                   grade_policy: new_cud[:grade_policy],
                                                   lecture: new_cud[:lecture])
         existing.assign_attributes(params.permit(:lecture, :section, :grade_policy))
+        existing.dropped = false
```
victorhuangwq, line 575: Could you quickly explain what is happening by setting dropped = false? I assume this is for when dropped students are added back, they should become "undropped"

db/migrate/20210929220050_create_github_integrations.rb
```diff
@@ -3,7 +3,7 @@ def change
     create_table :github_integrations do |t|
       t.string :oauth_state
       t.text :access_token_ciphertext
-      t.references :user, foreign_key: true, index: { unique: true }
+      t.references :user, foreign_key: true, type: :integer, index: { unique: true }
```
victorhuangwq, line 6: @fanpu this is the fix for the github integration issue btw

db/schema.rb
```diff
@@ -12,7 +12,7 @@
 
 ActiveRecord::Schema.define(version: 2021_10_23_185328) do
```
victorhuangwq, line 14: @michellexliu I think you should un-checkin this commit, because you are currently using MySQL and it added more options

victorhuangwq (changes requested):
Hi Michelle thanks for fixing the error so quickly. I have left my comments on each file, and once they are addressed, we should be good to go.

app/controllers/courses_controller.rb
michellexliu: Yep, that's correct! When we parse a "red" student, existing.dropped gets set to true, but previously we never set it back to false when the student gets added back in.

victorhuangwq (approved):
LGTM

PR #1441, opened 2021-12-17
[BUGFIX] fixed race condition bug for embedded quiz handin

ugogon:
If two students submit at the same time it may happen that the out.txt gets written by two requests at the same time resulting in invalid submissions. This fixes that by creating a tempfile so very submission gets its own file.

## How Has This Been Tested?
I submitted an embedded form.

## Types of changes

## Checklist:


fanpu (approved):
Thank you Ugo, this is a great catch. We hardly use embedded forms in CMU so we've never encountered the issue, so thanks for raising the issue and providing a fix!

PR #1442, opened 2021-12-17
[BUGFIX] fixed visibility of autograder output

ugogon:
Student can see their autograding output also in the case that the score is not directly release. This Commit fixes that

## How Has This Been Tested?
I viewed the submission and now you can not see the Autograder Output until the scores are released.

## Types of changes

## Checklist:

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -303,14 +303,14 @@ class SubmissionsController < ApplicationController
 
     # Adds autograded file as first option if it exist
     # We are mapping Autograder to header_position -1
-    unless @submission.autograde_file.nil?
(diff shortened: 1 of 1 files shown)


xinyis991105:
Hi Ugo! It looks like you're fixing it by hiding autograded feedback when an assessment's grading deadline isn't past yet. Some instructors do want their students to see the feedback before the deadline so they can perfect their submissions. I believe based on your description of the PR, the problem should instead be the following? When the autograder is configured not to release scores upon completion, students are still able to see their scores, even before the deadlines. May I confirm whether this is the issue you are facing?

ugogon:
yes, that is exactly the problem. I get the problem with my fix. I will propose another solution.

PR #1013, opened 2018-11-15
Extension Improvements

jbman223:
Fix page UI. fix characters in names. Also, fix yml not existing issues.

Fixes #931 
Fixes #1010 
Fixes #880

PR #1443, opened 2021-12-17
[BUGFIX+IMPROVEMENT] Fixed and added autocomplete for inputing students

ugogon:
There are mutiple bugs regarding the autocompletion for students in the new submission site and the assessment extension site. This PR fixes them and adds autocompletion to the sudo site as well.

## Motivation and Context
There are multiple bugs due to code redundancies and base64 that breaks the encoded string after 60 characters which results in invalid javascript. I removed the base64 conversion entirely because its not necessary.

## How Has This Been Tested?
Autocompletion and submitting the form on the new submission, assessment extension and sudo site.

## Types of changes

## Checklist:

This is not an optimal solution as we still have code redundancies. We should only write one autocompletion javascript file and import it wherever we need it.

### DIFF
--- a/app/controllers/extensions_controller.rb
+++ b/app/controllers/extensions_controller.rb
@@ -15,10 +15,8 @@ class ExtensionsController < ApplicationController
   def index
     @extensions = @assessment.extensions.includes(:course_user_datum)
     @users = {}
-    @usersEncoded = {}
     @course.course_user_data.each do |cud|
       @users[cud.full_name_with_email] = cud.id
-      @usersEncoded[Base64.encode64(cud.full_name_with_email.strip).strip] = cud.id
     end
     @new_extension = @assessment.extensions.new
   end
--- a/app/views/course_user_data/sudo.html.erb
+++ b/app/views/course_user_data/sudo.html.erb
@@ -1,10 +1,43 @@
+<% content_for :javascripts do %>
+  <script type="application/javascript">
+    jQuery(function() {
+      userData = {
+        <% @users.each do |k,v| %>
+          "<%= k %>": "<%= v %>",
+        <% end %>
+      };
+
+      /* user autocomplete */
+      $studentAutocompleteField = $('#student_autocomplete');
+      $hiddenCUDField = $('#sudo_email');
+      $studentAutocompleteField.autocomplete({
+        data: {
+          <% @users.each do |k,v| %>
+            "<%= k %>": null,
+          <% end %>
+        }
+      });
+
+      /* track changes in student autocomplete field */
+      $studentAutocompleteField.on('change', function() {
+        $hiddenCUDField.val(userData[$studentAutocompleteField.val()]);
+      })
+    });
+  </script>
+<% end %>
+
 <h2>Act as User</h2>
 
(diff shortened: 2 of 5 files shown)


cg2v:
I'm not necessarily suggesting that my versions of these fixes are better, but you might want to investigate https://github.com/cg2v/Autolab/commit/057493b3e518376d54fa62dc74b4ac7ba46353a5 and https://github.com/cg2v/Autolab/commit/752a44f942fe65168de0de003060d073c09817ea as alternatives to 
8d76c19

app/controllers/extensions_controller.rb
```diff
@@ -15,10 +15,8 @@ class ExtensionsController < ApplicationController
   def index
     @extensions = @assessment.extensions.includes(:course_user_datum)
     @users = {}
-    @usersEncoded = {}
     @course.course_user_data.each do |cud|
       @users[cud.full_name_with_email] = cud.id
-      @usersEncoded[Base64.encode64(cud.full_name_with_email.strip).strip] = cud.id
```
victorhuangwq, line 21: It seems like we have this Base64 encoding here for a reason, perhaps for students who username is not stored in base64 format? I will have to take a further look before deciding what's the appropriate fix here.

app/views/course_user_data/sudo.html.erb
```diff
@@ -1,10 +1,43 @@
...
+      });
+
+      /* track changes in student autocomplete field */
+      $studentAutocompleteField.on('change', function() {
+        $hiddenCUDField.val(userData[$studentAutocompleteField.val()]);
+      })
+    });
+  </script>
+<% end %>
+
```
victorhuangwq, line 28: I have tested the autocomplete out and it works. Thanks :)

app/views/submissions/new.html.erb
```diff
@@ -9,7 +9,7 @@
 
       /* user autocomplete */
       $studentAutocompleteField = $('#student_autocomplete');
-      $hiddenCUDField = $('#extension_course_user_datum_id');
+      $hiddenCUDField = $('#submission_course_user_datum_id');
```
victorhuangwq, line 12: I think this is sound too

victorhuangwq (commented):
I will circle back on the Base64 encoding

victorhuangwq:
https://github.com/cg2v/Autolab/commit/057493b3e518376d54fa62dc74b4ac7ba46353a5 and https://github.com/cg2v/Autolab/commit/752a44f942fe65168de0de003060d073c09817ea seems to be the correct fix, in addressing the issue brought up a while back. https://github.com/autolab/Autolab/pull/1013

damianhxy:
This PR coincides with the latest feature request by Prof Iliano (autocompletion for sudo) and is worth revisiting.

app/controllers/extensions_controller.rb
damianhxy: Encoding is necessary to avoid XSS attacks

damianhxy:
Note that currently, xss is possible via the autocomplete dropdown by crafting a malicious first name / last name

Related issue: https://github.com/materializecss/materialize/pull/49

PR #1444, opened 2021-12-17
[BUGFIX] Fixed sql syntax error

ugogon:
There is a SQL error when querying all announcements.

## Description
"system" is a keyword an needs to be escaped.

## How Has This Been Tested?
Now you can see all announcements.

## Types of changes

## Checklist:

### DIFF
--- a/app/controllers/announcements_controller.rb
+++ b/app/controllers/announcements_controller.rb
@@ -10,7 +10,7 @@ class AnnouncementsController < ApplicationController
   action_auth_level :index, :instructor
   def index
     @announcements = if @cud.user.administrator?
-                       Announcement.where("course_id=? or system", @course.id)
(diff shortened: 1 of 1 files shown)


app/controllers/announcements_controller.rb
```diff
@@ -10,7 +10,7 @@ class AnnouncementsController < ApplicationController
   action_auth_level :index, :instructor
   def index
     @announcements = if @cud.user.administrator?
-                       Announcement.where("course_id=? or system", @course.id)
+                       Announcement.where("course_id=? or `system`", @course.id)
```
victorhuangwq, line 13: definitely seems like a newly added keyword, at least in mysql. https://dev.mysql.com/doc/refman/8.0/en/keywords.html

victorhuangwq (approved):
Thank you for catching this! LGTM

PR #1445, opened 2021-12-17
[IMPROVEMENT] Added a Feedback Button

ugogon:
I added a Feedback button for every submission.

## Motivation and Context
Our students regularly fail to recognize that they are able to see the autograder feedback by clicking on the points. Its much more intuitive to have a separate button.

## How Has This Been Tested?
I clicked on the new button.

## Types of changes

## Checklist:

### DIFF
--- a/app/views/assessments/_submission_summary_row.html.erb
+++ b/app/views/assessments/_submission_summary_row.html.erb
@@ -129,7 +129,10 @@
       <%= link_to 'Destroy', [:destroyConfirm, @course, @assessment, submission], {:class=>"btn small"} %>
       <%= link_to 'Edit', [:edit, @course, @assessment, submission], {:class=>"btn small"} %>
   <% end %>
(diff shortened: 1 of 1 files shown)


xinyis991105:
Hi Ugo! It seems that the feedback button will automatically redirect to feedback of the last problem in the problem sets. It's fine as long as the problems share common autograded feedback. However, if the last problem is manually graded, this might confuse the students. A single feedback button also isn't very consistent with the per-problem feedback notion in the submission score summary. We agree that there should be better ways to signify to students that feedback is available for the autograded problems. We'll come back to you once we have discussed and come up with a solution! Thanks again for your devotion!!!!!!

ugogon:
Ah ok, I'm not familiar with the problem specific feedback and and always thought its a bit weird to have individual links. Now I think I get it. I close the PR.

PR #1454, opened 2022-01-08
Lint GitHub integration

victorhuangwq:
## Description

Ran rubocop on github_integration.rb & github_integrations_controller.rb

## Motivation and Context

Part of linting effots on autolab codebase

## Types of changes

## Checklist:

### DIFF
--- a/app/controllers/github_integrations_controller.rb
+++ b/app/controllers/github_integrations_controller.rb
@@ -10,6 +10,7 @@ class GithubIntegrationsController < ApplicationController
   action_auth_level :get_repositories, :student
   def get_repositories
     return fail_with_reason("User not connected to Github") unless @github_integration
+
     repositories = @github_integration.repositories
     render json: repositories, status: :ok
   rescue StandardError => e
@@ -21,6 +22,7 @@ class GithubIntegrationsController < ApplicationController
   def get_branches
     return fail_with_reason("User not connected to Github") unless @github_integration
     return fail_with_reason("Repository not provided") unless params["repository"]
+
     branches = @github_integration.branches(params["repository"])
     render json: branches, status: :ok
   rescue StandardError => e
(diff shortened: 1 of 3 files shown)


app/controllers/github_integrations_controller.rb
```diff
@@ -21,20 +22,20 @@ def get_repositories
...
 
-  private
+private
 
   def set_github_integration
-    gi = current_user.github_integration
-    if gi && gi.is_connected
-      @github_integration = current_user.github_integration
-    end
+    return if !current_user&.github_integration&.is_connected
```
victorhuangwq, line 36: let me know if this is similar to what you intended.

app/models/github_integration.rb
```diff
@@ -6,27 +6,27 @@ class GithubIntegration < ApplicationRecord
...
+                           query: {
+                             sort: "pushed",
+                             per_page: 30,
+                           })
+      repos.map { |repo|
+        { repo_name: repo[:full_name],
+          clone_url: repo[:clone_url],
+          default_branch: repo[:default_branch] }
+      }
     rescue Exception => e
```
victorhuangwq, line 26: Do we need to catch all exception here? or just standarderror

app/controllers/github_integrations_controller.rb
fanpu: Could the if and negation be changed to an unless?

fanpu (dismissed):
There are a couple of places using if ! <cond>, I think changing it to unless <cond> will be more idiomatic if we are going to change it anyway (https://github.com/rubocop/ruby-style-guide#if-vs-unless). Other than that everything else seems good!

app/controllers/github_integrations_controller.rb
victorhuangwq: yup that make sense

app/models/github_integration.rb
fanpu: Good catch, just StandardError. We don't want to deal with stuff raised internally by Ruby

fanpu (approved):

PR #1477, opened 2022-02-27
Git submission docs

fanpu:
## Description

Docs for Github Submission feature. This was based on feedback from bpr, and also the fact that we only had installation instructions for Github Submission right now.

One minor logic-related change is that I realised that our default handin extension when Git submission was enabled was `.tar` instead of `.tgz` as mentioned in the docs and in what actually happens, so I changed it to set it to `.tgz` to be consistent.

## Types of changes

## Checklist:


ashleyzhang (approved):
LGTM, thanks Fan Pu! I like the best practices and common issues section - we can expand on this as people use it

PR #1497, opened 2022-04-16
Adds shared comments feature

victorhuangwq:
## Description

Adds ability for instructors and CAs to have shared comments. Graders can easily opt in to have their comments be used by other graders.

## Motivation and Context

From interview with 122 TAs.

## How Has This Been Tested?

Pool here will refers to the list of shared comments shown

- Check add to pool, and save. Click edit to see that it is unchecked.
- Remove add to pool and save, able to see that being updated (unchecked, and no longer in the pool in creating new annotations)
- Delete annotation. See that it is no longer in the pool
- Edit annotation. See that it is updated in the pool
- Refresh to see if the change persists

Other cases:
- Act as a course assistant, and I'm still able to see the shared comments
- Change to a different assessment, I should not be able to see the shared comment from other assessments

Note: As of now, it does not appear when you are editing an annotations. To be discussed if that's what we want.

## Types of changes

## Checklist:

- [ ] Question: Do you think we should have a tool tip to tell them what the comment pool means?

### DIFF
--- a/app/assets/stylesheets/annotations.css
+++ b/app/assets/stylesheets/annotations.css
@@ -216,8 +216,10 @@
   padding: 5px;
 }
 
-.code-table .annotation-line .annotation-box .annotation-body p {
+.code-table .annotation-line .annotation-box .annotation-body div {
   margin: 5px 0;
+  font-family: 'Source Sans Pro', sans-serif;
+  white-space: pre-line;
 }
 
 
@@ -811,7 +813,7 @@
   padding-left: 10px;
 }
 
-/* Style dropdown */
+/* Autocomplete dropdown */
 .ui-menu .ui-menu-item {
   font-family: "Source Sans Pro", sans-serif;
   font-size: 16px;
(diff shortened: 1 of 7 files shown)


michellexliu:
The major functionality all seems to be working well! There is a small UI bug where the comment pool doesn't stick to the input box when you click into it and then scroll:
Screen Shot 2022-04-18 at 11 04 56 PM

Also, this isn't a major issue & is something that already wasn't working before this PR, but right now the annotations don't preserve line breaks (low priority issue if we want to ship this out fast).

victorhuangwq:
I definitely didn't notice that the line breaks wasn't preserved. Let me see if I can fix it.

victorhuangwq:
- Adds tooltip
image
- Adds autocomplete to editing
- Allow newlines in annotations

app/assets/stylesheets/annotations.css
```diff
@@ -794,3 +812,53 @@
...
+  max-height: 15rem;
+  max-width: 60%;
+  overflow-y: auto;
+  /* prevent horizontal scrollbar */
+  overflow-x: hidden;
+}
+
+/* Golden Layout */
+
+.lm_content {
```
victorhuangwq, line 839: Here

michellexliu (approved):
lgtm!

PR #1503, opened 2022-04-20
Submission Version Switching

damianhxy:
## Description
- Add buttons to switch to prev / next version of a student's submission that contains a specific file
- Make buttons update when switching between files

## Motivation and Context
When investigating AIVs, instructors might want to track changes to a specific file across the different versions of a student's submission. This PR adds buttons to enable instructors to easily switch between the different versions.

**Note**: For single file handin formats, the arrows will effectively work to cycle through the versions.

## How Has This Been Tested?
- Create a new assessment with handin format `handin.tar`
- Submit `hello.tar`, as well as modified versions (e.g. remove files, add files)
- Ensure that prev / next buttons work as expected (i.e. that they go to the same file, skip versions where the file doesn't exist)
- Ensure that when we change the file selected, the buttons update as expected
- Also check that for single file handin formats (e.g. `.c`, `.pdf`) that the arrow buttons serve to enumerate through the versions

Screen Shot 2022-07-17 at 20 46 57

## Types of changes

## Checklist:


damianhxy:
Considerations: To aid in code review, we want to be able to easily switch between next and previous versions of a particular file.
- Need to add a param to enable loading a specific file, rather than from an offset
- Arrow buttons should operate on the basis of a specific file's versions, i.e. prev / next buttons should be w.r.t. the file versions -- perhaps they should go to the prev / next version of the file? Or go to the prev / next version of the submission and put a placeholder if file does not exist
- Dropdown still shows all versions and clicking on a version loads the first file as usual

victorhuangwq:
@damianhxy could you elaborate what was changed since we last discussed? 

At the same time I do have concerns that people might get confused by the [# student / total students] and version / # of versions
 
image

app/views/submissions/view.html.erb
```diff
@@ -79,7 +79,8 @@
             </option>
           <% end %>
         </select>
-        <strong>/<%= @userVersions.length %>
+        <strong>/<%= @userVersions.first.version %>
```
victorhuangwq, line 82: May I understand the change here?

victorhuangwq (commented):
Functionally it looks good to me as described. I would just like some clarification and I think we are good to go.

app/views/submissions/view.html.erb
damianhxy: The issue with using `@userVersions.length` is that when we delete a submission, the version numbers of the other submissions do not change.

So if we had 5 submissions initially and deleted the first version, the dropdown would now show `version: {2, 3, 4, 5} / 4`.

By using the version of the first (i.e. latest) submission, the dropdown will now show `{2, 3, 4, 5} / 5`.

An alternative would be to update the version numbers of other submissions instead, but it seems overkill.

damianhxy:
> @damianhxy could you elaborate what was changed since we last discussed?

The main difference is that the arrow buttons now work on a per-file basis.

Previously, they would simply bring you to the next or previous version (and reset you to `header_position` 0). They would be grayed out when you are on the first / last version. But as Iliano pointed out, that isn't very useful from an instructor's perspective.

Now, the arrow buttons operate on a per-file basis. That is to say, clicking them will bring you to the next or previous version that **contains the file** (i.e. the same filepath). They will also now be grayed out when there is no older / newer version containing the file.

How this works is that when viewing a file (e.g. from an archive), we look through the other versions and search each one for the same filepath. If it exists, we take note of the `header_position` of the file in that particular version and append it to the querystring of the buttons so that we will jump directly to the same file when switching versions.

Note however that the functionality of the version dropdown remains unchanged, and using it to change versions will reset you to `header_position` 0 as usual.

> At the same time I do have concerns that people might get confused by the [# student / total students] and version / # of versions
> 
> image

Perhaps we can attach a `title` to the `span` so that users will see explanatory text when they hover over it?

Screen Shot 2022-07-22 at 13 01 15

app/views/submissions/view.html.erb
victorhuangwq: That's a sensible approach.

victorhuangwq:
I think adding the title would address the issue - good enough fix for now, because I think we might want to reimagine this page in the future.

At the same time I tested this
- Ensure that prev / next buttons work as expected (i.e. that they go to the same file, skip versions where the file doesn't exist)

It works as mentioned. It doesn't skip if I submitted a single file, and then a tar file afterwards however. But I guess that's an okay behavior.  (i.e. hello.c, then i submit hello.tar containing abc.c & def.c)

damianhxy:
> It works as mentioned. It doesn't skip if I submitted a single file, and then a tar file afterwards however. But I guess that's an okay behavior. (i.e. hello.c, then i submit hello.tar containing abc.c & def.c)

Yup, I don't expect the handin format to change across versions after all

victorhuangwq (approved):
LGTM

app/views/submissions/view.html.erb
```diff
@@ -26,8 +26,6 @@
     <% end %>
 
     const editableUrl = "<%= url_for ([:quickSetScore, @course, @assessment]) %>"
-    const rootFiles = $('.file-list').children()
```
damianhxy, line 29: With the new golden layout, these selectors are run before the elements exist. Remedied this by deferring evaluation until the values are needed.

app/views/submissions/_version_links.html.erb
```diff
@@ -3,13 +3,15 @@
     <%= link_to(url_for([:view, @course, @assessment, @prevVersion[:submission], header_position: @prevVersion[:header_position]]), class: "btn small", title: "Previous version containing this file (shortcut: [)", id: "prev_version_link") do %>
       <i class="material-icons">south</i>
     <% end %>
+    <link rel="prerender" href=<%= url_for([:view, @course, @assessment, @prevVersion[:submission], header_position: @prevVersion[:header_position]]) %> />
```
victorhuangwq, line 6: I don't think we should be prerendering the next/prev version pages - it's a resource intensive suggestion & we are already prerendering the left & right students.

victorhuangwq: Seems like it won't be resource intensive.
https://www.chromium.org/developers/design-documents/prerender/#minimizing-resource-contention
but it means only 2 out of our 4 pages will be prerendered

PR #1507, opened 2022-04-26
Metrics included/excluded categories

ashleyzhang:
Screen Shot 2022-04-26 at 5 56 49 PM

## Description
Add a configurations tab to metrics feature that includes a "Considered Categories" section enabling professors to decide what assessment categories are or are not included in watchlist calculations.

## Motivation and Context
Some courses have optional assessments that do not factor into determining whether or not a student is at risk. Allowing instructors to exclude a category of assessments reduces false positive rates for students appearing in the watchlist.

## How Has This Been Tested?
Navigate to the configurations tab in the student metrics dashboard. Exclude a category in the watchlist, and save the changes. Ensure that the change is reflected in the watchlist. 

## Types of changes

## Checklist:


victorhuangwq:
Tested. It seems like the default behavior for when you resolve and then change the configuration is to archive everything (just like a change of metrics). Is that intended behavior?

db/schema.rb
```diff
@@ -242,11 +244,12 @@
     t.datetime "resolved_at"
     t.integer "resource_owner_id"
     t.string "access_code"
+    t.index ["application_id"], name: "fk_rails_4035c6e0ed"
```
victorhuangwq, line 247: this seems to be a bit odd to push up @ashleyzhang, as this looks like a config?

xinyis991105:
> Tested. It seems like the default behavior for when you resolve and then change the configuration is to archive everything (just like a change of metrics). Is that intended behavior?

Yes this is expected! Currently a change in metrics config has the same effect on watchlist instances as a change in risk conditions, i.e. refreshing the list, archiving old instances.

xinyis991105:
In addition, I've had some misunderstanding of the use of the word "allowlist" and currently allowlist contains all the categories that are not considered in the calculation of metrics. Do we want to change this to blocklist afterwards?

victorhuangwq:
I think that would be confusing - would it be possible to change that in this PR? I'm afraid we might forget that later

xinyis991105:
> I think that would be confusing - would it be possible to change that in this PR? I'm afraid we might forget that later

See the latest changes. I rollbacked and changed the relevant migration file instead of writing another migration file for renaming.

victorhuangwq (approved):
Did a rails db:reset and utilized the new schema. Feature works as expected. Refactor looks fine.
LGTM!

PR #1510, opened 2022-04-27
Consistent annotation among group submissions

xinyis991105:
## Description
This PR ensures annotations are consistent among group members' submissions of the same iteration.

## Motivation and Context
This is in the interest of polishing the group feature of Autolab.

## How Has This Been Tested?
1. Create a group of >= 2.
2. Act as one of the members and hand in a file.
3. Reset user and annotate one of the members' submission. Toggle among the members' submissions and see that the same annotation shows up. (You might want to refresh again if there doesn't appear to be a change due to the latency of database.)
4. Edit this annotation, and see that the change is reflected across members.
5. Delete this annotation and see that it's been deleted for all members.

## Types of changes

## Checklist:

### DIFF
--- a/app/controllers/annotations_controller.rb
+++ b/app/controllers/annotations_controller.rb
@@ -29,12 +29,18 @@ class AnnotationsController < ApplicationController
       respond_with(@course, @assessment, @submission, primary_annotation)
     end
 
-    # All submissions of the iteration in the group, excluding the current one
-    group_submissions = @submission.group_associated_submissions
-
     # Set shared comment to false to avoid duplicates in shared comment pool
     tweaked_params = annotation_params
     tweaked_params[:shared_comment] = false
+
+    # All submissions of the iteration in the group, excluding the current one
+    group_submissions = @submission.group_associated_submissions
+
+    # Set up annotation group key
+    submission_group_key = @submission.group_key
+    annotation_group_key = "#{submission_group_key} + #{tweaked_params[:problem_id]}"
+    annotation_group_key += Time.current.utc.to_s(:number)
+    tweaked_params[:group_key] = annotation_group_key
 
     annotations = [primary_annotation]
 
--- a/app/controllers/annotations_controller.rb
+++ b/app/controllers/annotations_controller.rb
@@ -17,14 +17,39 @@ class AnnotationsController < ApplicationController
   # POST /:course/annotations.json
   action_auth_level :create, :course_assistant
   def create
-    annotation = @submission.annotations.new(annotation_params)
+    # primary annotation
+    primary_annotation = @submission.annotations.new(annotation_params)
 
-    ActiveRecord::Base.transaction do
-      annotation.save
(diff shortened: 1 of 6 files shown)


victorhuangwq:
Tested based on given test case.
Works as expected. Deleting and editing annotations get reflected accordingly. Acted as members of the group and the same annotations show.
Realize that I would prefer to see grade submissions by group as well, but that's not a big issue now.

victorhuangwq:
Could you do a rebase to ensure that all the migrations don't mess things up?

app/controllers/annotations_controller.rb
```diff
@@ -17,22 +17,79 @@ class AnnotationsController < ApplicationController
...
   action_auth_level :create, :course_assistant
   def create
-    annotation = @submission.annotations.new(annotation_params)
+    # primary annotation
+    primary_annotation = @submission.annotations.new(annotation_params)
 
-    ActiveRecord::Base.transaction do
-      annotation.save
-      annotation.update_non_autograded_score
+    if @submission.group_key.empty?
```
victorhuangwq, line 23: seems like based on this implementation, you are:
- adding a group key
- creating a duplicate annotation related to the primary annotation via the group key
- and then deleting and updating all the group key.

I was under the assumption that the implementation be something like:
- when creating an annotation for a group add a foreign key that is the group_id
- when viewing a group submission, retrieve all annotations based on the group_id as well

I think the current implementation works for now, but I'm afraid with future additions it might get confusing. For example, whenever we have something similar to a shared_comments feature, we need to also ensure that the duplicates do not have the shared_comment field checked.

victorhuangwq (changes requested):
Refer to the above comment on my thoughts on this implementation. 
If you are interested in getting this feature shipped, we can discuss further.

app/controllers/annotations_controller.rb
xinyis991105: 1. The current implementation already checks for duplicate shared_comment field across annotations.
2. I'm assuming you still want the group key to be present for the same iteration of submissions? Retrieving annotations by group_id gives you annotations for all submissions of the group. Without the group key, it can be hard to figure out which submissions are associated in the same iteration. Versions can be different across group members, and timestamps might differ depending on the database's performance.

victorhuangwq: I can see that it already checks for duplicate shared_comment. What I was trying to say that if we need to introduce a new feature similar to shared_comment, we need to think about this particular edge case and update group comments accordingly, which might get tricky.

My main thought is that it would be best if there can be a way for us to only store 1 single copy of annotation, without needing to create duplicates.

xinyis991105: I would propose retaining the submission group_key and add group_id to annotations. Should we disable annotating on non-latest submission in the future, we can get rid of group_key as well. Let me know if it sounds ok!

PR #1511, opened 2022-04-27
Improve accessibility of Admin Course button

damianhxy:
## Description
This PR implements the relevant aria code from the assessment view page to the manage course page. It also replaces the obsolete `onkeypress` event with `onkeydown`.

## Motivation and Context
Similar to #1215, adding the `aria-expanded` attribute improves accessibility.

## How Has This Been Tested?
Checked that the dropdowns have the correct initial `aria-expanded` attribute (`true` for admin course and `false` for admin autolab). Also checked that clicking on a dropdown, or tabbing to a dropdown and pressing the enter key, toggles the `aria-expanded` attribute correctly.

## Types of changes

## Checklist:

### DIFF
--- a/app/assets/javascripts/handin.js
+++ b/app/assets/javascripts/handin.js
@@ -34,7 +34,7 @@ function dragExit(e) {
 
 function clickDrag(e) {
   // Ignore shift and tab
-  if (e.key === 9 || e.key === 16) return;
+  if (e.key === "Shift" || e.key === "Tab") return;
   $("#handin_show_assessment input[type='file']").trigger('click');
 }
 
--- a/app/views/assessments/_submission_panel.html.erb
+++ b/app/views/assessments/_submission_panel.html.erb
@@ -15,7 +15,7 @@
   <div class="ui tab active" data-tab="upload" id="upload_tab">
     <div class="row handin-row">
       <div class="valign-wrapper drag-drop-handin" tabindex="0"
-          onclick="clickDrag();" onkeydown="clickDrag()" ondrop="dropHandler(event);"
+          onclick="clickDrag(event);" onkeydown="clickDrag(event);" ondrop="dropHandler(event);"
           ondragover="dragOverHandler(event);" ondragenter="dragEnter(event);" ondragexit="dragExit(event);">
         <p class="center-align" style="color:grey;" ondrop="dropHandler(event);" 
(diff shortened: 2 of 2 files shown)


app/assets/javascripts/handin.js
```diff
@@ -32,7 +32,9 @@ function dragExit(e) {
   $(".drag-drop-handin").get(0).style = undefined;
 }
 
-function clickDrag() {
+function clickDrag(e) {
+  // Ignore shift and tab
+  if (e.key === 9 || e.key === 16) return;
```
najclark, line 37: With the original changes the submission box didn't work for me (neither clicking nor pressing a key worked). The `e` check should ensure `e` is not undefined before accessing `key`, which appears to be the issue. Also, the key codes you had didn't work for me, but "Shift" and "Tab" did.

```suggestion
  if (e && (e.key === "Shift" || e.key === "Tab")) return;
```

app/views/assessments/_submission_panel.html.erb
```diff
@@ -15,7 +15,7 @@
   <div class="ui tab active" data-tab="upload" id="upload_tab">
     <div class="row handin-row">
       <div class="valign-wrapper drag-drop-handin" tabindex="0"
-          onclick="clickDrag();" onkeypress="clickDrag()" ondrop="dropHandler(event);" 
+          onclick="clickDrag();" onkeydown="clickDrag()" ondrop="dropHandler(event);"
```
najclark, line 18:
```suggestion
          onclick="clickDrag();" onkeydown="clickDrag(event)" ondrop="dropHandler(event);"
```
I think `event` needs to be added here to allow `clickDrag` to case on what key was pressed.

najclark (changes requested):
I made some suggested changes regarding the drag/drop handin functionality.

app/views/assessments/_submission_panel.html.erb
damianhxy: Good catch -- fixed

app/assets/javascripts/handin.js
damianhxy: Changed the names, I believe the undefined check is no longer needed now that event is correctly passed to the function.

najclark (approved):
LGTM

xinyis991105 (approved):
LGTM!

PR #1512, opened 2022-04-28
Make releaseAllGrades, withdrawAllGrades, etc. POST to prevent CSRF attacks

20wildmanj:
This PR aims to fix the CSRF vulnerability for 'releaseAllGrades' along with other vulnerable endpoints that previously used GET requests instead of POST.

Thanks to Damien and Fan Pu for the discussion about the security flaw and the useful article here about CSRF and rails:
https://guides.rubyonrails.org/security.html#csrf-countermeasures

## Description
- change releaseAllGrades, withdrawAllGrades, and releaseSectionGrades in the
assessment page to be POST requests, from GET requests.
- turn link_to in the frontend for these actions to be forms with button_to
- changed styling of input classes for danger-zone to look like links, to match
the style of the other actions on the page

## Motivation and Context

We were sent this issue by huntr.dev, where a contributor pointed out this flaw.

## How Has This Been Tested?

Tested mainly through local host, seeing if we can directly access `http://localhost:3000/courses/AutoPopulated/assessments/homework0/releaseAllGrades` from the browser, which should no longer work (should get an error saying there doesn't exist a GET request with that endpoint).

## Types of changes

## Checklist:

@damianhxy

### DIFF
--- a/app/assets/stylesheets/style.css.scss
+++ b/app/assets/stylesheets/style.css.scss
@@ -1078,6 +1078,20 @@ form p:last-child {
   padding-left: 14px;
 }
 
+.danger-side input {
+  background: none;
+  border: none;
+  padding-left: 14px;
+  text-decoration: none;
+  color: #0882af;
+  font-family: "Source Sans Pro", sans-serif;
+  font-weight: 300;
+  font-size: 16px;
+  text-align: left;
+  width: 100%;
+  cursor: pointer;
+}
+
 /* Form Tables */
 
 .verticalTable tr {
--- a/app/controllers/assessments_controller.rb
+++ b/app/controllers/assessments_controller.rb
@@ -64,6 +64,8 @@ class AssessmentsController < ApplicationController
   action_auth_level :set_repo, :instructor
   action_auth_level :import_svn, :instructor
 
+  protect_from_forgery with: :exception
+
   def index
     @is_instructor = @cud.has_auth_level? :instructor
     announcements_tmp = Announcement.where("start_date < :now AND end_date > :now",
@@ -649,7 +651,6 @@ class AssessmentsController < ApplicationController
 
   action_auth_level :releaseAllGrades, :instructor
 
-  protect_from_forgery
   def releaseAllGrades
     # release all grades
     num_released = releaseMatchingGrades { |_| true }
@@ -667,7 +668,6 @@ class AssessmentsController < ApplicationController
 
   action_auth_level :releaseSectionGrades, :course_assistant
 
-  protect_from_forgery
   def releaseSectionGrades
     unless @cud.section? && !@cud.section.empty? && @cud.lecture && !@cud.lecture.empty?
       flash[:error] =
@@ -696,7 +696,6 @@ class AssessmentsController < ApplicationController
 
   action_auth_level :withdrawAllGrades, :instructor
 
-  protect_from_forgery
   def withdrawAllGrades
     @assessment.submissions.each do |submission|
       scores = submission.scores.where(released: true)
--- a/app/controllers/assessments_controller.rb
+++ b/app/controllers/assessments_controller.rb
@@ -649,6 +649,7 @@ class AssessmentsController < ApplicationController
 
   action_auth_level :releaseAllGrades, :instructor
 
+  protect_from_forgery
   def releaseAllGrades
     # release all grades
     num_released = releaseMatchingGrades { |_| true }
@@ -666,6 +667,7 @@ class AssessmentsController < ApplicationController
 
   action_auth_level :releaseSectionGrades, :course_assistant
 
+  protect_from_forgery
   def releaseSectionGrades
     unless @cud.section? && !@cud.section.empty? && @cud.lecture && !@cud.lecture.empty?
       flash[:error] =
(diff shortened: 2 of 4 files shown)


victorhuangwq:
I'm assuming that you have tested to ensure that releaseAllGrades and withdrawAllGrades still work?

app/assets/stylesheets/style.css.scss
```diff
@@ -1065,6 +1065,19 @@ form p:last-child {
   border-left: 6px solid #e53935;
   border-right: 1px solid #e53935;
 }
+.danger-side input{
```
damianhxy, line 1068: Nit: space before {

app/views/assessments/show.html.erb
```diff
@@ -73,9 +73,19 @@
               <li class="collection-item red-text danger-bottom no-hover"><h4>Danger Zone</h4></li>
 
               <li class="danger-side">
-                <%= link_to "Release all grades", {:action => "releaseAllGrades" }, {:title=>  "Make all scores for this assessment visible to students", :class=>"", data: {confirm: "Are you sure you want to release all grades?"}} %>
-                </li>
-              <li class="danger-side"> <%= link_to "Withdraw all grades", {:action => "withdrawAllGrades" }, {:title=>  "Hide all scores for this assessment from students", data: {confirm: "Are you sure you want to withdraw all grades?"}} %></li>
+                <%= form_with :url => { :action => "releaseAllGrades" } do |f| %>
+
+                <%= button_to "Release all grades", {:method => "post" }, {:title=>  "Make all scores for this assessment visible to students", :class=>"", data: {confirm: "Are you sure you want to release all grades?"}} %>
```
damianhxy, line 78: Empty :class can be omitted

damianhxy: Nit: I think it would be good to have only a single space after :title=>

damianhxy (changes requested):
Checked that GET routes no longer exist, and POST routes work as expected.

However, it seems that CSRF protection isn't working -- even after deleting the `authenticity_token` fields in a form (or filling them with garbage) and submitting, the request still succeeds. This needs further investigation

20wildmanj:
> I'm assuming that you have tested to ensure that releaseAllGrades and withdrawAllGrades still work?
Both methods still work, has been tested.


> Checked that GET routes no longer exist, and POST routes work as expected.
> 
> However, it seems that CSRF protection isn't working -- even after deleting the `authenticity_token` fields in a form (or filling them with garbage) and submitting, the request still succeeds. This needs further investigation

I can confirm that CSRF doesn't seem to be working as intended. I first tried using postman to send a POST request, which actually failed (didn't release grades), but by making just an html file with a form and using that to send a POST request, I was able to release all grades without being on autolab. This definitely seems to be a bit more tricky than originally thought.

20wildmanj:
I've added protect_from_forgery to each of the methods (releaseAllGrades etc.) and it now seems that if I try to make a POST request outside of Autolab it should fail:
Screen Shot 2022-04-29 at 0 53 36

app/views/assessments/show.html.erb
```diff
@@ -73,9 +73,19 @@
               <li class="collection-item red-text danger-bottom no-hover"><h4>Danger Zone</h4></li>
 
               <li class="danger-side">
-                <%= link_to "Release all grades", {:action => "releaseAllGrades" }, {:title=>  "Make all scores for this assessment visible to students", :class=>"", data: {confirm: "Are you sure you want to release all grades?"}} %>
-                </li>
-              <li class="danger-side"> <%= link_to "Withdraw all grades", {:action => "withdrawAllGrades" }, {:title=>  "Hide all scores for this assessment from students", data: {confirm: "Are you sure you want to withdraw all grades?"}} %></li>
+                <%= form_with :url => { :action => "releaseAllGrades" } do |f| %>
```
damianhxy, line 76: `form_with` and `button_to` can be replaced with `<%= button_to "Release all grades", { :action => "releaseAllGrades" }, {:title=> "Make all scores for this assessment visible to students", data: {confirm: "Are you sure you want to release all grades?"}} %>`

damianhxy:
> I've added protect_from_forgery to each of the methods (releaseAllGrades etc.) and it now seems that if I try to make a POST request outside of Autolab it should fail: Screen Shot 2022-04-29 at 0 53 36
> 
> But also I'm not sure if this is ideal if it gives out an error like this?
> 
> EDIT: After restarting the server it seems to give a nicer error message: Screen Shot 2022-04-29 at 1 00 39

Confirmed that this fixes the issue. However, it seems weird that `protect_from_forgery` has to be repeated so many times, and there might be some other underlying issue.

Error message shouldn't appear in prod, so that's fine

app/controllers/assessments_controller.rb
```diff
@@ -64,6 +64,8 @@ class AssessmentsController < ApplicationController
   action_auth_level :set_repo, :instructor
   action_auth_level :import_svn, :instructor
 
+  protect_from_forgery with: :exception
```
damianhxy, line 67: Would be good to figure out why `protect_from_forgery` inside `application_controller.rb` isn't sufficient

damianhxy (approved):
LGTM!

PR #1515, opened 2022-05-10
Consistent annotation alternative implementation

xinyis991105:
Serves the same purpose as #1510 but with a neater implementation. See #1510 for details.

## Types of changes

## Checklist:


app/controllers/annotations_controller.rb
```diff
@@ -30,8 +30,11 @@ def create
   # PUT /:course/annotations/1.json
   action_auth_level :update, :course_assistant
   def update
+    tweaked_params = annotation_params
+    tweaked_params.delete(:submission_id)
+    tweaked_params.delete(:filename)
```
victorhuangwq, line 35: May I know why this is modified?

app/controllers/submissions_controller.rb
```diff
@@ -436,6 +436,12 @@ def view
     @problemReleased = @submission.scores.pluck(:released).all?
 
     @annotations = @submission.annotations.to_a
+    unless @submission.group_key.empty?
+      group_submissions = @submission.group_associated_submissions
+      group_submissions.each do |group_submission|
+        @annotations += group_submission.annotations.to_a
```
victorhuangwq, line 442: You are appending to `@annotations` which already has  `@submission.annotations.to_a` assigned to it.
Would that lead to the original submission having double the annotations? I don't see that when testing however.

app/helpers/assessment_handin_core.rb
```diff
@@ -90,11 +90,17 @@ def saveHandin(sub, app_id = nil)
     end
 
     submissions = []
+
+    # group_key = group_name_submitter_email_handin_filename_timestamp
+    group_key = "#{group.name}_#{@cud.user.email}_#{@assessment.handin_filename}_"
+    group_key += Time.current.utc.to_s(:number)
+
```
victorhuangwq, line 97: just to clarify, this is to uniquely identify the group submission, and this is needed because there might be a case where it is like version 1 for group member a, and version 2 for group member b?

victorhuangwq (approved):
Tested based on given test case. Works as expected. 
Implementation is cleaner than the previous approach. 

Left some clarifying questions, but lgtm!

app/helpers/assessment_handin_core.rb
damianhxy: Might there be an issue if any of the fields contain an underscore?

app/controllers/annotations_controller.rb
damianhxy: From #1510: "Delete submission_id and filename to avoid association with the wrong submission"

damianhxy: I believe that the intention was to ensure that when we update an annotation from a different submission than the submission it was originally created in, we do not accidentally override the original values of `submission_id` and `filename` with the new one.

But actually, this doesn't appear to be necessary; the `submission_id` and `filename` that is submitted via Ajax seems to always be the original values (because the data is derived from `@annotations` in `submissions/view.html`).

app/controllers/submissions_controller.rb
damianhxy: `group_associated_submissions` specifically gets all submissions _except_ the current submission (`Submission.where(group_key: group_key).where.not(id: id)`), so we do not duplicate annotations for the original submission.

app/helpers/assessment_handin_core.rb
damianhxy: Yes, that seems to be precisely the case, especially if one student made submissions before joining the group, so the version numbers might differ. This allows us to figure out exactly which "submissions" are part of a certain group submission.

Using underscores as a separator seems a bit iffy to me, but a name collision seems pretty much impossible especially given the timestamp at the end, so this is fine.

damianhxy (approved):
Followed testing plan and code worked as expected. Logic looks good too.

LGTM!

PR #1517, opened 2022-05-12
Hotfix for metrics grace days

xinyis991105:
## Description
On the current master branch, metrics still take into consideration the number of grace days of assessments whose category is blocklisted. This PR fixes the problem by comparing the sum of grace days used on allowlisted assessments to the risk condition's threshold.

## Motivation and Context
To improve metrics feature.

## How Has This Been Tested?
Play around with grace day usage and blocklist a category of assessments ignoring whose grace day usage will invalidate a watchlist instance for that risk condition. Blocklist that category that verify that no watchlist instances appear for them.

## Types of changes

## Checklist:

### DIFF
--- a/app/models/watchlist_instance.rb
+++ b/app/models/watchlist_instance.rb
@@ -468,13 +468,17 @@ class WatchlistInstance < ApplicationRecord
     return unless latest_aud_before_date.global_cumulative_grace_days_used >= grace_day_threshold
 
     violation_info = {}
+    allowlist_asmt_cumulative_gdu = 0
     auds_before_date.each do |aud|
-      violation_info[aud.assessment.display_name] = aud.grace_days_used if aud.grace_days_used > 0
+      if aud.grace_days_used > 0
+        allowlist_asmt_cumulative_gdu += aud.grace_days_used
+        violation_info[aud.assessment.display_name] = aud.grace_days_used
+      end
     end
 
     # The logic might fall through if the grace days were used on blocklisted assessments
     # This ensures no funky "0 grace day used" instance is added
-    return if violation_info.empty?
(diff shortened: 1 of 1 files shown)


damianhxy:
Tested as follows:

Create two assessments with categories `Lab` and `Lab2` respectively.
Have a user who used 63 and 45 grace days on the assessments respectively.

Set watchlist for 100 grace days. Then,
- Test with both categories included
- Test with only `Lab` included
- Test with only `Lab2` included
- Test with no categories included
By saving configuration and reloading after each change.

Before patch
- Both: Appears (108)
- Only `Lab`: Does not appear
- Only `Lab2`: Appears (45) -- this is wrong
- Neither: Does not appear

After patch
- Both: Appears (108)
- Only `Lab`: Does not appear
- Only `Lab2`: Does not appear
- Neither: Does not appear

damianhxy (approved):
Logic seems sound, `allowlist_asmt_cumulative_gdu` correctly computes total number of grace days used across non blocklisted assessments as `auds_before_date` does not contain blocklisted assessments. Also tested and this fix works as expected.

The primary issue seems to be that `global_cumulative_grace_days_used` can sometimes return a larger value than expected due to `cumulative_grace_days_used_before` considering blocklisted assessments. Ideally, we would want to explicitly exclude blocklisted assessments here. But this fix works too since we never get a smaller value than expected.

LGTM!

app/models/watchlist_instance.rb
```diff
@@ -468,9 +468,18 @@ def self.add_new_instance_for_cud_grace_day_usage(course, condition_id,
     return unless latest_aud_before_date.global_cumulative_grace_days_used >= grace_day_threshold
```
damianhxy, line 468: Note: I believe this line is no longer needed after the fix

PR #1522, opened 2022-05-14
Better config reload error handling

### LINKED ISSUE #1516: course reload diagnostics always fail
**Your environment**
This is seen on the production Carnegie Mellon Autolab, which is cg2v@e37e12abd33998c63018c9f90b99b7f53b109e98, based on 2b2164e586490434d12ea8a5e021bc660deea1b1

**Steps To Reproduce**
1. Go to a course
2. create a syntax error in the course.rb
3. Click on 'Manage Course'
4. Open 'Admin Course'
5. Click 'Reload course config file
6. See donkey kong

**Current behavior**
```
#<ActionView::Template::Error: undefined method `backtrace' for nil:NilClass>
Stack Dump
/opt/autolab/app/views/courses/reload.html.erb:7:in `_app_views_courses_reload_html_erb___1636394926029189055_70270092293520'
/opt/autolab/app/controllers/courses_controller.rb:225:in `reload'
```

**Expected behavior**
reload.html should provide help for instructors to fix course.rb problems

**Analysis**
reload.html.erb expects an `@error` variable, but the controller does not set it

damianhxy:
Modify course and assessment config reload code to better handle cases where an invalid config is provided

## Description
- Modified course `reload.html` to be consistent with assignment `reload.html`
- Validate syntax of config (via compilation) before saving
- Correctly catch (certain classes of) exceptions when attempting to load config files
- Make backups of last-known good configs when reloading

## Motivation and Context
Currently, course config reload does not set the `@error` variable, so the error page is not correctly rendered.

Furthermore, config reload does not check the syntax of the config file before saving it, meaning that it's possible to reload a file with invalid syntax and corrupt the config.

Also fixes #1516

## How Has This Been Tested?
- Under normal circumstances, course config reload still works
- Under normal circumstances, assessment config reload still works
- When reloading, observe that a backup copy of the original config is made
- Corrupt course config `courses/<course-name>/course.rb` and reload config. Error should be correctly caught and config in `courseConfig/<course-name>.rb` unaffected
- Corrupt assessment config `courses/<course-name>/<assessment-name>/<assessment-name>.rb` and reload config. Error should be correctly caught and config in `assessmentConfig/<course-name>-<assessment-name>.rb` unaffected

## Types of changes

## Checklist:

### DIFF
--- a/app/controllers/assessments_controller.rb
+++ b/app/controllers/assessments_controller.rb
@@ -607,11 +607,9 @@ class AssessmentsController < ApplicationController
 
   def reload
     @assessment.load_config_file
-    # rubocop:disable Lint/RescueException
-  rescue Exception => e
+  rescue StandardError, SyntaxError => e
     @error = e
     # let the reload view render
-    # rubocop:enable Lint/RescueException
   else
     flash[:success] = "Success: Assessment config file reloaded!"
     redirect_to(action: :show) && return
--- a/app/controllers/assessments_controller.rb
+++ b/app/controllers/assessments_controller.rb
@@ -607,8 +607,11 @@ class AssessmentsController < ApplicationController
 
   def reload
     @assessment.load_config_file
-  rescue StandardError
+    # rubocop:disable Lint/RescueException
+  rescue Exception => e
+    @error = e
     # let the reload view render
+    # rubocop:enable Lint/RescueException
   else
     flash[:success] = "Success: Assessment config file reloaded!"
     redirect_to(action: :show) && return
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -223,11 +223,9 @@ class CoursesController < ApplicationController
   action_auth_level :reload, :instructor
   def reload
     @course.reload_course_config
-    # rubocop:disable Lint/RescueException
-  rescue Exception => e
+  rescue StandardError, SyntaxError => e
     @error = e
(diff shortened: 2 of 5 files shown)


app/controllers/assessments_controller.rb
```diff
@@ -607,8 +607,11 @@ def valid_json?(json)
 
   def reload
     @assessment.load_config_file
-  rescue StandardError
+    # rubocop:disable Lint/RescueException
+  rescue Exception => e
```
michellexliu, line 611: `rescue Exception` should be avoided + best to avoid disabling the linter; try `rescue SyntaxError` instead

app/controllers/courses_controller.rb
```diff
@@ -222,8 +222,13 @@ def users
 
   action_auth_level :reload, :instructor
   def reload
-    render && return unless @course.reload_course_config
-
+    @course.reload_course_config
+    # rubocop:disable Lint/RescueException
+  rescue Exception => e
```
michellexliu, line 227: same as above

michellexliu (changes requested):
see comments

damianhxy:
Note that certain classes of errors (such as `NameError` and `ArgumentError`) will still not be caught. If a user introduces such an error they'll be stuck. There should be a way to gracefully recover from such a mistake (or at the very least, make a backup of last-known good config)

PR #1524, opened 2022-05-14
Hotfix for CSRF protection

damianhxy:
Amendment to PR #1512

## Description
- Remove duplicate `protect_from_forgery` added in #1512
- Correctly specify `only: [action]` for `skip_before_action` statements inside `action_no_auth`

## Motivation and Context
#1512 added CSRF protection by repeating `protect_from_forgery`. However, this overrides the `skip_before_action` inside `action_no_auth`. Since Tango does not have a CSRF token, this means autograding feedback is rejected.

In reality, there is no need to repeat `protect_from_forgery`; it wasn't working because the `skip_before_action` did not specify `only: [action]`.

## How Has This Been Tested?
- CSRF protection for releaseAllGrades, withdrawAllGrades, releaseSectionGrades still work as expected
- Autograding feedback works as expected
- Ensure that admin routes `/admin/github_integration` and `/admin/email_instructors` are only accessible by admins

## Types of changes

## Checklist:

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -76,15 +76,15 @@ class ApplicationController < ActionController::Base
   end
 
   def self.action_no_auth(action)
-    skip_before_action :verify_authenticity_token, raise: false
-    skip_before_action :authenticate_user!, raise: false
-    skip_before_action configure_permitted_paramters: [action]
-    skip_before_action maintenance_mode: [action]
-    skip_before_action run_scheduler: [action]
+    skip_before_action :verify_authenticity_token, only: [action], raise: false
+    skip_before_action :authenticate_user!, only: [action], raise: false
+    skip_before_action configure_permitted_paramters, only: [action]
+    skip_before_action maintenance_mode, only: [action]
+    skip_before_action run_scheduler, only: [action]
 
-    skip_before_action authenticate_user: [action], raise: false
+    skip_before_action authenticate_user, only: [action], raise: false
     skip_before_action :authorize_user_for_course, only: [action]
-    skip_before_action authenticate_for_action: [action], raise: false
+    skip_before_action authenticate_for_action, only: [action], raise: false
     skip_before_action :update_persistent_announcements, only: [action], raise: false
   end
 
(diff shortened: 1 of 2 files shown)


app/controllers/application_controller.rb
```diff
@@ -78,13 +78,13 @@ def self.action_auth_level(action, level)
   def self.action_no_auth(action)
     skip_before_action :verify_authenticity_token, only: [action], raise: false
     skip_before_action :authenticate_user!, only: [action], raise: false
-    skip_before_action configure_permitted_paramters, only: [action]
-    skip_before_action maintenance_mode, only: [action]
-    skip_before_action run_scheduler, only: [action]
+    skip_before_action :configure_permitted_paramters, only: [action]
+    skip_before_action :maintenance_mode?, only: [action]
```
fanpu, line 82: Good catch

fanpu (approved):
The changes looks good, thanks for making the syntax consistent and fixing the bug! Tested that autograding returns results successfully. Did not test the CSRF protection, @20wildmanj could you double check on that front before we merge? Thanks!

20wildmanj (changes requested):
I've verified that previous CSRF protection changes broke autograding feedback, and this PR fixes the issue. Also releaseAllGrades / withdrawAllGrades still have CSRF protection working.

However there seems to be an issue with authentication in other areas of the site now. For example when I try to view my user profile, I get either one of these two errors:
Screen Shot 2022-05-18 at 0 34 17
Screen Shot 2022-05-18 at 0 34 54

@damianhxy could you look into what's going on here?

app/controllers/application_controller.rb
```diff
@@ -65,7 +65,6 @@ def self.action_auth_level(action, level)
 
     if level == :administrator
       skip_before_action :authorize_user_for_course, only: [action], raise: false
-      skip_before_action authenticate_for_action: [action]
```
damianhxy, line 68: Currently, this line of code is silently ignored because `authenticate_for_action` isn't defined at this point.

Even so, this line of code should be removed. Otherwise, we are skipping authentication and allowing non-admins to access admin routes.

For example, if we change this line to `skip_before_action :authenticate_for_action, only: [action], raise: false`, we observe that non-admins can access `/admin/github_integration` and `/admin/email_instructors` routes.

20wildmanj (approved):
Issue with not being able to access user profile is now resolved, functionality remains the same. LGTM!

PR #1532, opened 2022-05-19
Improve extensions + submissions autocomplete

damianhxy:
Improve correctness of extensions + submissions autocomplete, and extract common logic.
Also fix an XSS issue in gradebook, and various visual display issues.

## Description
- `CGI.escapeHTML` on backend
- `escape_javascript` on frontend
- remove `sanitize` from flash
- replace `encode64` with `strict_encode64` to avoid line feeds
- extracted common backend logic to `ApplicationController`
- extracted common frontend logic to `components/_autocomplete.html.erb`
- use `to_csv` for `download_roster`

## Motivation and Context
Currently, there are several issues with autocomplete, including
- XSS in name / email
- Disappearing letters when backslashes exist in name
- base64 breaking on long names
- download roster breaking on names with commas

This PR serves to improve the robustness of extensions + submissions autocomplete

## How Has This Been Tested?
First created the following users:
(invalid emails can be saved by using inspect element and deleting the type from the input field)
| First Name | Last Name | Email | Comments |
| -- | -- | -- | -- |
| `\back` | `\slash` | `back@slash.com` | Test js escaping |
| `` | `alert` | `bad@alert.com` | Test XSS, also tests long base64 encoding | 
| `<strong>first</strong>` | `<strong>last</strong>` | `first@last.com` | Test HTML escaping |
| `qn?` | `mark` | `question@mark.com` | Tests base64 encodings with `/` |
| `Space` (with a leading space) | `Name` | `space@name.com` | Tests that btoa is serving its purpose of handling names with leading / trailing spaces |
| `italics` | `email` | `<em>name</em>@italicsemail.com` | Tests HTML escaping |
| `bad` | `email` | `<strong>name</strong>@bademail.com` | Tests HTML escaping |
| `script` | `email` | `<script>alert("hello")</script>@scriptemail.com` | Tests HTML escaping |
- Ensured that flash displays correctly when creating users (i.e. any tags are escaped)
- Successfully give all users an infinite extension, with no autocomplete display issues
- Successfully create submissions for all users (via Manage Submissions > Create New Submission), with no autocomplete display issues
- Ensured that gradebook displays correctly
- Ensured that download roster produces a csv file that displays correctly

## Types of changes

## Checklist:


app/controllers/submissions_controller.rb
```diff
@@ -35,11 +35,7 @@ def new
         render([@course, @assessment, :submissions]) && return
       end
     else
-      @cuds = {}
-      # TODO: change order
-      @course.course_user_data.joins(:user).order("email ASC").each do |cud|
-        @cuds[cud.full_name_with_email] = cud.id
-      end
+      retrieve_autocompletion_data!
```
damianhxy, line 38: The `.joins(:user).order("email ASC")` appears redundant, so this should be fine

app/form_builders/form_builder_with_date_time_input.rb
```diff
@@ -126,16 +126,19 @@ def date_helper(name, options, strftime, date_format, alt_format)
...
+  def wrap_field(name, field, help_text = nil, display_name = nil)
     @template.content_tag :div, class: "input-field" do
       label(name, display_name, class: "control-label") +
          field + help_text(name, help_text)
     end
   end
 
   def help_text(_name, help_text)
-    @template.content_tag :p, help_text, class: "help-block"
+    if help_text.nil?
```
damianhxy, line 137: Avoid rendering `help-block` unless needed, to save on vertical screen estate

cg2v:
Have you tested with names containing characters that would be entity encoded (I used ' (`&apos;`) to make sure they're not double encoded?

Did you verify that the flash xss occurs without these changes? the flash rendering seems safe already:
```
              <%= content_tag :div, sanitize(msg), id: "flash_#{name}",
                  class: "#{name}" %>
```

damianhxy:
Re: flash, I recall that the success flash did not escape `<em>name</em>@italicsemail.com` before the changes. However, I'll do more testing later and with the error flashes too.

I will also test for double encoding later.

damianhxy:
Examples of flash XSS with the email: `<em>italics</em>@email.com`

Screen Shot 2022-05-20 at 9 35 48 PM
Screen Shot 2022-05-20 at 9 36 13 PM

damianhxy:
It appears `sanitize` isn't sufficient for our purposes as it _strips_ tags, rather than escapes them.

Screen Shot 2022-05-20 at 9 42 52 PM

damianhxy:
OK - Admittedly `sanitize` _does_ prevent XSS by stripping unsafe tags and attributes. But nonetheless, I believe we should still HTML escape in order to display the text as-is.

damianhxy:
Re: double encoding, there does not seem to be any issues. Text displays correctly, and creating extensions, creating submissions, sudo, etc. work.
Screen Shot 2022-05-20 at 10 06 47 PM
Screen Shot 2022-05-20 at 10 07 20 PM
Screen Shot 2022-05-20 at 10 07 32 PM

victorhuangwq:
Is this still WIP or is it ready for review?

cg2v (commented):
it seems pointless to call sanitize() after escape_html(). sanitize won't find anything to filter because escape_html has already encoded all the tag metacharacters.

It appears that if you don't call sanitize, you don't need escape_html either, because rails will automatically escape the inner text of the content_tag.(https://makandracards.com/makandra/2579-how-to-use-html_safe-correctly)

cg2v (commented):
Instead of this, it might make sense to have the controller use strict_encode64 instead of urlsafe_encode64 (my bad there)

cg2v (commented):
I suppose I'll have to test this at some point, but I don't see how this can avoid breaking the autocomplete in some cases. the keys in usersEncoded have to match the keys in users. They differ only in whether or not they're encoded. If the keys don't match, the user can't be autocompleted.

damianhxy:
Re: `sanitize` / `escape_html`, if Ruby indeed escapes everything correctly in the absence of `sanitize`, removing it would seem to be the best solution. The only downside would be that we wouldn't be able to style the flashes -- but I don't think we do that anyway (other than `roster_error`, which is handled specially in `application.html.erb` anyway).

Re: `strict_encode64`, I agree that's a better solution given that we don't use the base64 in urls anyway, in which case I'll also remove the additional `.replace`s

Re: autocomplete, I'm not sure which line of code you're specifically referring to. However, I don't see any immediate issues with the current approach. Could you clarify this?

cg2v:
Say my name is
First Name | Last Name | Email 
-- | -- | -- 
Chask'l | Grundman | cg2v@autolab.io


Then `@users[CGI.escapeHTML cud.full_name_with_email] = cud.id` will use the value
Chask\&apos;l Grundman (cg2v@autolab.io) 
and `@usersEncoded[Base64.urlsafe_encode64(cud.full_name_with_email.strip).strip] = cud.id` will use 
Q2hhc2snbCBHcnVuZG1hbiAoY2cydkBhdXRvbGFiLmlvKQ==

When the javascript calls btoa on Chask\&apos;l Grundman (cg2v@autolab.io), it gets Q2hhc2smYXBvcztsIEdydW5kbWFuIChjZzJ2QGF1dG9sYWIuaW8p which does not match.

However, the existing code is broken if there are tags in the name (materialize strips the tags - there's no xss). So perhaps the correct thing to do is to escapeHTML before encoding. Things are getting very hacky at this point and it leads me to wonder how other software deals with this problem.

damianhxy:
Hmm, this doesn't seem to be an issue (at least on my browser, FF100 Mac) since `$studentAutocompleteField.val()` appears to unescape the characters before the call to `btoa`.
Screen Shot 2022-05-22 at 3 31 46 PM

Perhaps there is no need to change the code then? (However, I'll change `j k` to `j k.html_safe` as per production to avoid double escapes, not that it seems to cause an issue though)

damianhxy:
Also note that I am doing `j k` instead of `j k.html_safe` as we appear to need an additional level of escaping to avoid XSS attacks in the autocomplete dropdown (autocomplete seems to unescape once).

damianhxy:
Alternatively, `CGI.escapeHTML (CGI.escapeHTML cud.full_name_with_email)` together with `j k.html_safe` also seems to work, but I'm not sure which is preferable.

cg2v:
If that's what autocomplete is doing, I will withdraw my objections

app/views/extensions/index.html.erb
```diff
@@ -102,9 +79,8 @@
     <p><b>Create New Extension</b></p>
       <%= form_for @new_extension, :as=>"extension", :url=>{:action=>"create"}, builder: FormBuilderWithDateTimeInput do |f| %>
         <div class="input-field">
-          <label class="control-label active" for="student_autocomplete">Student Name/Email</label>
-          <input type="text" id="student_autocomplete" placeholder="Start typing student name or email" class="autocomplete" autocomplete="off">
-          <p class="help-block"></p>
+          <input type="text" size="3" id="student_autocomplete" class="autocomplete" autocomplete="off"/>
```
damianhxy, line 82: This is to make the style consistent with that of create submission

damianhxy:
@victorhuangwq This PR is ready for review!

app/controllers/submissions_controller.rb
damianhxy: Mainly because order doesn't really matter when we're using the autocomplete widget

app/helpers/gradebook_helper.rb
```diff
@@ -71,11 +71,12 @@ def gradebook_rows(matrix, course, section = nil, lecture = nil)
       grace_days = 0
       late_days = 0
 
+      # CGI.escapeHTML to avoid XSS
```
damianhxy, line 74: Note: We escape here because in `gradebooks/view.html.erb`, we use `<%== gradebook_rows(...).to_json %>`. That is, we output the raw html (`<%== ... %>`) and we don't have the opportunity to escape the string then.

app/controllers/application_controller.rb
```diff
@@ -331,6 +331,20 @@ def make_dlist(cuds)
...
   end
 
+  # Retrieves necessary data for autocompletion purposes
+  # Currently used by manage extensions and create submission
+  def retrieve_autocompletion_data!
+    @users = {}
+    @usersEncoded = {}
+    @course.course_user_data.each do |cud|
+      # Escape once here, and another time in _autocomplete.html.erb (see comments)
+      @users[CGI.escapeHTML cud.full_name_with_email] = cud.id
```
damianhxy, line 341: We escape once here, and once on the frontend because the alternatives are
- Escape twice on backend, but this leads to a yucky `CGI.escapeHTML CGI.escapeHTML ...` on the backend and a `<%= j k.html_safe %>` on the frontend.
- Escape twice on frontend, but there isn't an easy way to do so since `escape_html` sets `html_safe` to true, and so a second `escape_html` would not work to escape the text again.

I think this is the best way to go about it.

app/controllers/extensions_controller.rb
```diff
@@ -14,12 +14,7 @@ class ExtensionsController < ApplicationController
   action_auth_level :index, :instructor
   def index
     @extensions = @assessment.extensions.includes(:course_user_datum)
-    @users = {}
-    @usersEncoded = {}
-    @course.course_user_data.each do |cud|
-      @users[cud.full_name_with_email] = cud.id
-      @usersEncoded[Base64.encode64(cud.full_name_with_email.strip).strip] = cud.id
-    end
+    retrieve_autocompletion_data!
```
victorhuangwq, line 17: I wonder if retrieve_autocompletion_data return the information instead of directly assigning to @users, as someone who is looking at this function would not have known what was being assigned.

victorhuangwq (approved):
Tested the functionality as given by the test case, works as expected.
Good that you abstracted repeated functionality!

PR #1533, opened 2022-05-19
Sudo Autocomplete

damianhxy:
Add autocomplete to Sudo page.

Also make sudo utilize user id instead of email, to facilitate reuse of autocomplete code between extension, submission, and sudo.

## Description
- Add autocomplete logic to frontend
- Add autocomplete logic to backend
- Use cud id instead of email

## Motivation and Context
As mentioned by Prof Iliano, it would be good to have autocomplete on the sudo page to easily select the correct user.

## How Has This Been Tested?
(Same users from #1532)
- Successfully sudoed and unsudoed each user, with no autocomplete display issues

## Types of changes

## Checklist:

### DIFF
--- a/app/views/course_user_data/sudo.html.erb
+++ b/app/views/course_user_data/sudo.html.erb
@@ -6,9 +6,9 @@
 
 <p>You can use sudo to examine how Autolab appears to other users. You can also change your permission levels to see what your course assistants or students generically see.</p>
 
-<div>
+<p>
 <%= form_tag do %>
-  Act as user (email):
+  <p><b>Act as user (email):</b></p>
   <div class="input-field">
     <input type="text" size="3" id="student_autocomplete" class="autocomplete" autocomplete="off"/>
     <label for="student_autocomplete">Start typing student name or email</label>
@@ -16,4 +16,4 @@
   <%= hidden_field_tag(:sudo_id) %>
   <%= submit_tag "Sudo", {:value=>"Sudo", :class=> "btn primary"} %>
 <% end %>
-</div>
+</p>


app/views/course_user_data/sudo.html.erb
```diff
@@ -1,10 +1,51 @@
+<% content_for :javascripts do %>
+
+  <script type="application/javascript">
+    jQuery(function() {
+      /* match user name/email with cud_email */
+      /* escape_javascript prevents issues with backslashes in names, etc. */
+      userData = {
+          <% @usersEncoded.each do |k,v| %>
+            "<%= j k %>": "<%= v %>",
```
damianhxy, line 9: I am doing `j k` instead of `j k.html_safe` as per production Autolab, since that defeat the point of `CGI.escapeHTML`. That is, it unescapes the escaped html.

app/controllers/course_user_data_controller.rb
```diff
@@ -190,20 +190,25 @@ def sudo
       redirect_to([@cud.course]) && return
     end
 
-    return unless request.post?
-
-    sudo_user = User.where(email: params[:sudo_email]).first
-    unless sudo_user
-      flash[:error] = "User #{params[:sudo_email]} does not exist"
-      redirect_to(action: :sudo) && return
+    @users = {}
```
damianhxy, line 193: Replace this chunk with `retrieve_autocompletion_data!` once #1532 merged

app/views/course_user_data/sudo.html.erb
```diff
@@ -1,10 +1,51 @@
+<% content_for :javascripts do %>
+
+  <script type="application/javascript">
```
damianhxy, line 3: Replace this chunk with `<%= render partial: "common/autocomplete", locals: { hiddenCUDField: "#sudo_id" } %>` once #1532 merged

app/controllers/course_user_data_controller.rb
```diff
@@ -190,20 +190,18 @@ def sudo
...
 
-    sudo_user = User.where(email: params[:sudo_email]).first
-    unless sudo_user
-      flash[:error] = "User #{params[:sudo_email]} does not exist"
-      redirect_to(action: :sudo) && return
-    end
+    return unless request.post?
 
-    sudo_cud = @course.course_user_data.where(user_id: sudo_user.id).first
+    sudo_cud = @course.course_user_data.where(id: params[:sudo_id]).first
```
damianhxy, line 197: Previously
- User inputs `sudo_email`
- Search for corresponding user with that email, fail if not found
- Search for corresponding cud for that user, fail if not found
The new way is more direct
- User inputs `sudo_id`
- Search for corresponding cud for that id, fail if not found

damianhxy: After that, we can obtain the corresponding user by finding the user with `id` equal to `sudo_cud.user_id`. Technically, we might needto check if `sudo_user` is non-nil. But practically, this shouldn't happen, and besides we wouldn't really be able to give a meaningful error message.

victorhuangwq (approved):
LGTM. Tested sudo and it works as expected :)

PR #1545, opened 2022-06-02
Fix XSS in metrics watchlist

damianhxy:
## Description
Escape student names and emails on the metrics watchlist page.

## Motivation and Context
Currently, student names and emails are not escaped on the metrics watchlist page. This can possibly lead to XSS attacks.

## How Has This Been Tested?
Create user with a name containing ``. Make user use a grace day and apply the grace day metric to get user to show up on watchlist.

Before changes
Screen Shot 2022-06-01 at 21 51 53
After changes
Screen Shot 2022-06-01 at 21 51 59
## Types of changes

## Checklist:

### DIFF
--- a/app/assets/javascripts/watchlist.js
+++ b/app/assets/javascripts/watchlist.js
@@ -1,6 +1,14 @@
 // Loads all Semantic javascripts
 //= require semantic-ui
 
+const escapeHtml = (unsafe) => {
+  return unsafe.replaceAll('&', '&amp;')
+               .replaceAll('<', '&lt;')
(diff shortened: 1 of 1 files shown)


fanpu (approved):
We discussed offline about the pros and cons of this approach vs other ones like preventing names with special characters in the first place, but came to the conclusion that we should  assume people could actually have names with such characters. The proposed way of escaping the main XSS special characters gives us good enough mileage for now based on our threat model, and we can revisit with a more general solution if we see more of such issues in the future. LGTM!

PR #1549, opened 2022-06-08
Fix Flash links

damianhxy:
## Description
- Pluralize some strings using `pluralize`
- Add `flash[:html_safe] = true` whenever the flash should be displayed verbatim (if we know the whole string is safe and have a link that we want to display)
- Add logic on frontend to set `html_safe` whenever `flash[:html_safe]` is set, hence not escaping `msg`

## Motivation and Context
Currently, there appears to be some attempts to mark flashes as `html_safe` in the controllers (from ~2015)
- In the past, marking a string as `html_safe` in the controller would ensure the flash message would not be escaped in the view
- However, since Rails 4.1, the serialization method was changed, so HTML safety was no longer preserved ([source](https://groups.google.com/g/rubyonrails-core/c/z52zgDgUmbs))
- That said, this seems redundant given the view code of the time, which marked everything as `html_safe` (a terrible idea)

Since #216, `sanitize(msg)` was used as an alternate way to mark messages as `html_safe`. This prevents XSS attacks, but can lead to unexpected styling since tags are stripped but not escaped. This led to the removal of `sanitize` in #1532 altogether, to ensure that emails always appeared as expected.

However, this meant that some legitimate uses of links in flashes would not show up properly. To counteract this, I am proposing the use of a "flag" via `flash[:html_safe]` that we can set to `true` should we need a string to display verbatim. Otherwise, by default, Rails will escape the flash message as usual.

## How Has This Been Tested?
- Batch regrade submissions, observe that link appears as expected (`Regrading x submissions`) -- note "submission" is conditionally pluralized
- Regrade all submissions, observe that link appears as expected (`Regrading the most recent submissions from x students`) -- note "student" is conditionally pluralized
- Make a submission, observe that link appears as expected (`Submitted file <user>_<ver>_<file>.c (Job ID = xx) for autograding. Refresh the page to see the results.`)
- A variety of escaped flashes display as usual (e.g. sudo, unsudo)
- Also check that escaping works (e.g. sudo to a user whose email contains html -- requires inspect element to remove `type=email` from element)
  - For example, add user to course with email `<strong>name</strong>@bademail.com` (bypassing validation by removing type from email field)
  - Then, sudo to the user (also removing type from email field)
- The other code paths are analogous and do not require special testing

## Types of changes

## Checklist:

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -19,7 +19,6 @@ class ApplicationController < ActionController::Base
   before_action :authenticate_for_action
   before_action :update_persistent_announcements
   before_action :set_breadcrumbs
-  before_action :mark_flash_html_safe
 
   rescue_from ActionView::MissingTemplate do |_exception|
     redirect_to("/home/error_404")
@@ -310,19 +309,6 @@ protected
                     else
                       (view_context.link_to @course.full_name, [@course], id: "courseTitle")
                     end
-  end
-
-  # To display flash messages verbatim, set flash[:html_safe] = true
-  # Might want to consider separate flags for different flash types
-  def mark_flash_html_safe
-    return unless flash[:html_safe]
-
-    flash.delete(:html_safe)
-    flash.each do |name, _|
-      # rubocop:disable Rails/OutputSafety
-      flash.now[name] = flash[name].html_safe
-      # rubocop:enable Rails/OutputSafety
-    end
   end
 
   def pluralize(count, singular, plural = nil)
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -19,6 +19,7 @@ class ApplicationController < ActionController::Base
   before_action :authenticate_for_action
   before_action :update_persistent_announcements
   before_action :set_breadcrumbs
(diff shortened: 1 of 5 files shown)


damianhxy:
Alternative approaches include
- Reverting to `sanitize(msg)` and escaping on backend instead -- but this can be error-prone leading to unexpected styling from `msg`, and `sanitize` could inhibit more creative uses of `flash`
- Creating flash classes such as `success_safe`, `error_safe`, etc -- but this seems more troublesome and unnecessarily complicated
- Check for `flash[:html_safe]` on frontend instead of inside `ApplicationController` and manage logic accordingly -- but I feel that the logic is best left to the backend

20wildmanj:
I think for now what you're doing in terms of using `flash[:html_safe]` on frontend is fine, I'm not an expert on where we should be doing this logic but at least this way it's not too complicated, don't need to implement anything extra.

20wildmanj (approved):
- batch regrade link appears as expected and is pluralized, same with regrade all
- autograde submission appears as expected
- flashes still appear to function fine
- verified that escaping for new user with html in email works

PR #1550, opened 2022-06-09
Remove home layout

damianhxy:
## Description
- Delete home layout
- Clean-up application layout

## Motivation and Context
Currently, the home layout is almost a duplicate of the application layout, and is only used when rendering the home page (i.e. `courses#index`). In fact, the only difference when rendering with the application layout is that the tab title is "Autolab" instead of "Autolab Home".

In order to avoid the need to duplicate future code changes across both layouts, this PR deletes the home layout and makes `courses#index` default to the usual application layout.

## How Has This Been Tested?
- Used RubyMine grep to ensure no other route renders with the home layout
- Diffed home layout with new application layout and ensured that any differences are inconsequential (notably, it has additional includes, `roster_error` flash handling, and persistent announcements)
- Ensured that home page looks the same and functions the same (but no surprises there)

## Types of changes

## Checklist:


20wildmanj:
When logged out, when I try to go to home (e.g. localhost:3000) I get a 404 not found error. I couldn't find a simple fix as well (I thought that since you removed `render layout: "home"` in `app/controllers/courses_controller.rb` I tried replacing that line with `render layout: "application"` but that didn't fix it). I also get a 404 error just by going to `http://localhost:3000/auth/users/sign_in` but not for developer log in. (Actually it seems like the redirect to the sign-in causes a 404 as well)

Screen Shot 2022-06-10 at 17 37 32

20wildmanj (approved):
Home page works now for not logged in users, I found a couple more references to "home" in `home_controller.rb` (line 36, 52) but they seem to be unused so it shouldn't affect anything.

LGTM!

PR #1520, opened 2022-05-13
Fix application xss

damianhxy:
## Description
- Remove unnecessary raw tag

## Motivation and Context
- Currently, application name is xss-able

## How Has This Been Tested?
- Go to `/oauth/applications`
- Create application with name ``
- Click on authorize, name should be correctly escaped

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] I have run rubocop for style check. If you haven't, run `overcommit --install && overcommit --sign` to use pre-commit hook for linting
- [ ] My change requires a change to the documentation, which is located at [Autolab Docs](https://github.com/autolab/docs)
- [ ] I have updated the documentation accordingly, included in this PR

PR #1552, opened 2022-06-25
Rails 6.0.5 Migration

damianhxy:
## Description
Migrate Autolab to Rails 6.0.5.

Procedure
1. Delete `Gemfile.lock` and run `bundle update` with Rails 6.0.0
2. Run `rails app:update` with Rails 6.0.0
3. Change `rails` to 6.0.5 in `Gemfile`
4. Delete `Gemfile.lock` and run `bundle update` with Rails 6.0.5
5. Run `rails app:update` with Rails 6.0.5

**Note:** after running `bundle update`, platforms `ruby`, `x86_64-darwin-19`, `x86_64-darwin-20` and `x86_64-linux` were added to `Gemfile.lock` to ensure compatibility on other platforms. (`bundle lock --add-platform xxx`)

Gems with semver major changes
- lockbox (0.6.8 -> 1.0.0)
- marcel (0.3.3 -> 1.0.2)
- psych (new 4.0.4)
- faraday (1.10.0 -> 2.3.0)
- faraday-net_http (1.0.1 -> 2.0.3)
- rack-test (1.1.0 -> 2.0.2)

Notable fixes
- Used `.to_hash` in `risk_condition.rb` to address an error `Psych::DisallowedClass: Tried to load unspecified class: ActiveSupport::HashWithIndifferentAccess` when creating watchlist metrics. This was due to a breaking change in `Psych` whereby YAML `safe_load` was made the new default. The problem manifested due to `.to_h` on `new_metrics_params` (of class `ActionController::Parameters`) yielding a `ActiveSupport::HashWithIndifferentAccess`. From what I can see, `.to_h` is not used in a similar manner anywhere else.

Miscellaneous fixes
- Added some missing `flash[:html_safe]` where `<br>` were used in flashes (See #1549 for reference, but basically, set `flash[:html_safe] = true` whenever we want to display flashes raw)
- Fix link to custom scoreboard docs
- Fix confirmation prompt for delete user button
- Fix incorrect rendering for application name on authorization page (#1520 accidentally escaped the `<strong>` tags as well)
- Add `<th>` element corresponding to the operations column in the scheduler table to improve table appearance
- Align folders with files in speedgrader

Note: The previous migration to 6.0 was somewhat incomplete as we did not migrate to the new defaults. i.e., we ought to eventually include `config.load_defaults 6.0` under `application.rb`. However, this breaks some existing code (especially the new default `active_record.belongs_to_required_by_default = true`) and isn't necessary per se. So I am leaving it commented out for now.

Also, the message ``To use retry middleware with Faraday v2.0+, install `faraday-retry` gem`` pops up when running Autolab. As far as I can tell, this is a non-issue as I do not believe we use retry middleware.

## Motivation and Context
Keeping libraries updated and enabling us to resolve dependabot alerts.

## How Has This Been Tested?
Successfully completed [migration tests](https://docs.google.com/spreadsheets/d/1K89NjIgr_98V6Z0JFV6d4BvoUD8dz5RczY4jfZokrdc/edit#gid=0).

## Types of changes

## Checklist:


Gemfile
```diff
@@ -66,6 +66,7 @@ gem 'oauth2'
 
 # Github API
 gem "octokit", "~> 4.0"
+gem "faraday-retry"
```
damianhxy, line 69: Needed for faraday >= 2, which octokit uses

app/controllers/application_controller.rb
```diff
@@ -211,6 +211,7 @@ def authorize_user_for_course
     @cud.errors.full_messages.each do |msg|
       flash[:error] += "<br>#{msg}"
     end
+    flash[:html_safe] = true
```
damianhxy, line 214: See #1549 for context. Essentially, the PR escaped all flash messages by default (previously, we used `sanitize` which allowed unintended styling from user controlled strings) and set the "flag" `flash[:html_safe] = true` if we wish to display the message verbatim.

Touching up on some flashes which use `<br>` that I missed out on.

app/models/risk_condition.rb
```diff
@@ -25,8 +25,9 @@ def self.create_condition_for_course_with_type(course_id, type, params, version)
             "Make sure your request body fits the criteria!"
     end
 
-    options = { course_id: course_id, condition_type: type, parameters: params, version: version }
-    new_risk_condition = RiskCondition.new(options)
+    options = { course_id: course_id, condition_type: type, parameters: params.to_hash,
+                version: version }
+    new_risk_condition = RiskCondition.new(options.to_hash)
```
damianhxy, line 30: See PR description for more details. Essentially, it seems that we can't serialize `ActiveSupport::HashWithIndifferentAccess`. The problem doesn't seem to manifest anywhere else.

app/views/doorkeeper/authorizations/new.html.erb
```diff
@@ -13,7 +13,7 @@
 <h3>Authorization Required</h3>
 
 <p class="h4">
-  <%= t('.prompt', client_name: "<strong class=\"text-info\">#{ @pre_auth.client.name }</strong>") %>
+  <%= raw t('.prompt', client_name: "<strong class=\"text-info\">#{ h @pre_auth.client.name }</strong>") %>
```
damianhxy, line 16: See #1520 for context. The PR removed the `raw` tag because of the potential XSS via the client name. However, this escapes the `<strong> ... </strong>` tags too. Now, restore the `raw` tag but escape only the client name.

app/views/schedulers/index.html.erb
```diff
@@ -10,6 +10,7 @@
     <th>Next</th>
     <th>Interval</th>
     <th>Visual Run</th>
+    <th>Operations</th>
```
damianhxy, line 13: Adding a `<th>` that corresponds to the operations column as currently there is no column header, which makes the table appear unsightly.

app/views/users/edit.html.erb
```diff
@@ -27,5 +27,5 @@
 <% if current_user.administrator? %>
 <br> <br><hr><br>
 <h3> Admin Only</h3>
-  <%= link_to raw('<span class="btn">Delete User</span>'), user_path(@user), method: :delete, confirm: "Are you sure!?" %>
+  <%= link_to raw('<span class="btn">Delete User</span>'), user_path(@user), method: :delete, data: {confirm: "Are you sure!?"} %>
```
damianhxy, line 30: `jquery-ujs` requires us to set `data-confirm` on the element, so the current code is wrong and results in no confirmation before user deletion.

config/application.rb
```diff
@@ -8,6 +8,11 @@
 
 module Autolab3
   class Application < Rails::Application
+    # Initialize configuration defaults for originally generated Rails version.
+    # config.load_defaults 5.0
```
damianhxy, line 12: Technically, we should strive to flip over to the new Rails defaults eventually by setting `config.load_defaults 5.0` or even `6.0`. However, some existing code breaks (e.g. the new default `active_record.belongs_to_required_by_default = true` causes issues with current code that does not necessarily initialize foreign keys). This isn't necessary per se so it is fine to continue using old defaults for now.

damianhxy (commented):
Annotating the changes.

app/models/risk_condition.rb
```diff
@@ -25,7 +25,8 @@ def self.create_condition_for_course_with_type(course_id, type, params, version)
             "Make sure your request body fits the criteria!"
     end
 
-    options = { course_id: course_id, condition_type: type, parameters: params, version: version }
+    options = { course_id: course_id, condition_type: type, parameters: params.to_hash,
```
damianhxy, line 28: See PR description for more details. Earlier on, `.to_h` was called on `new_metrics_params` (of class `ActionController::Parameters`) which yielded a `ActiveSupport::HashWithIndifferentAccess`. The params here is also a `ActiveSupport::HashWithIndifferentAccess`, and so we use `.to_hash` to convert it to a normal `Hash` before attempting to save. This is due to a breaking change in `Psych` (used somewhere down the stack), where YAML `safe_load` is the new default, which does not support most classes.

app/views/doorkeeper/authorizations/new.html.erb
20wildmanj: Is the h before the `@pre_auth.client.name` for here how to specify that it should be escaped?

damianhxy: Yes, see https://api.rubyonrails.org/classes/ERB/Util.html#method-c-h

20wildmanj:
When trying to download all submissions from a section as a CA, I get a CourseUserDatum::AuthenticationFailed error:

Screen Shot 2022-07-13 at 23 32 49

20wildmanj:
When trying to make a git submission as a student, I get another CourseUserDatum::AuthenticationFailed error:
Screen Shot 2022-07-13 at 23 55 41

damianhxy:
> When trying to download all submissions from a section as a CA, I get a CourseUserDatum::AuthenticationFailed error:
> 
> Screen Shot 2022-07-13 at 23 32 49

Addressed in #1563

> When trying to make a git submission as a student, I get another CourseUserDatum::AuthenticationFailed error: Screen Shot 2022-07-13 at 23 55 41

Addressed in #1564

20wildmanj (approved):
Ran through Autolab features through testing document comprehensively, all issues that were found in the process were fixed, LGTM!

PR #1553, opened 2022-06-29
Improve flash styles

damianhxy:
## Description
- Change all occurrences of `:info` to `:notice`
- Remove unused `#flash_alert.alert` style
- Removed unused `#flash_alert.success` and `#flash_notice.success` styles
- Made flash style code DRY-er

## Motivation and Context
Currently, `flash[:info]` is used in several locations throughout the codebase despite there being no style for `:info`, but only `:notice`. Hence, this PR migrates all uses of `:info` to `:notice`.

Additionally, some style code is unused, such as `#flash_alert.alert` (`:alert` is not used anywhere), and `#flash_alert.success` + `#flash_notice.success` (with the current flash code, the class always mirrors the id). Hence such styles are removed in this PR.

This PR also extracts the styling common to the different flash types to improve DRY-ness of code.

## How Has This Been Tested?
- Grepped codebase to ensure only `:error`, `:notice`, `:success` are used (and not `:alert` nor `:info`). Note that `:roster_error` and `:html_safe` are used for custom logic.
- In `application.html.erb`, added the following code under `div#flashes` and opened home page to observe that flashes were correctly rendered.
```erb
<%= content_tag :div, "success", id: "flash_success", class: "success" %>
<%= content_tag :div, "error", id: "flash_error", class: "error" %>
<%= content_tag :div, "notice", id: "flash_notice", class: "notice" %>
```

## Types of changes

## Checklist:

### DIFF
--- a/app/assets/stylesheets/style.css.scss
+++ b/app/assets/stylesheets/style.css.scss
@@ -1156,16 +1156,19 @@ div.center {
   margin-top: 8px;
 }
 
-#flash_alert.alert,
-#flash_error.error,
-#flash_notice {
-  color: #ce4844;
+#flash_error,
+#flash_notice,
+#flash_success {
   padding: 10px;
   margin-bottom: 0px;
   border-radius: 3px;
   border: 1px solid #eee;
-  border-left: 5px solid #ce4844;
   font-weight: 500;
+}
+
+#flash_error {
+  color: #ce4844;
+  border-left: 5px solid #ce4844;
 }
 
 #flash_error table {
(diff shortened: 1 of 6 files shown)


victorhuangwq:
Does this change the look of the flashes?

fanpu (commented):
Agree with the changing from info to notice, it seems like :alert and :notice are the canonical flashes and others would be custom [based on the flash controller](https://github.com/rails/rails/blob/main/actionpack/lib/action_controller/metal/flash.rb).

app/assets/stylesheets/style.css.scss
```diff
@@ -1156,18 +1156,21 @@ div.center {
   margin-top: 8px;
 }
 
-#flash_alert.alert,
-#flash_error.error,
-#flash_notice {
-  color: #ce4844;
+#flash_error,
```
fanpu, line 1159: I am wondering if it would be more idiomatic to change all these to classes instead of IDs, since IDs are meant to be only used once. What do you think?

fanpu (approved):
LGTM, replicated test plan and changes looks good!

PR #1557, opened 2022-07-11
Fix roster format

### LINKED ISSUE #997: Roster export format differs from import format
I noticed that the roster CSV format that is exported by Autolab appears to differ from the "General Autolab Format" that it imports.

In [uploadRoster.html.erb](https://github.com/autolab/Autolab/blob/master/app/views/courses/uploadRoster.html.erb#L71), the "General Autolab Format" is described as containing the fields `Semester,email,last_name,first_name,school,major,year,grading_policy,courseNumber,courseLecture,section`. This is consistent with the behavior of the [`parse_roster_csv`](https://github.com/autolab/Autolab/blob/master/app/controllers/courses_controller.rb#L586) function, which does not use the `courseNumber` column.

However, the [`downloadRoster`](https://github.com/autolab/Autolab/blob/master/app/controllers/courses_controller.rb#L271) function, instead of leaving the `courseNumber` column empty, writes CSV rows that doesn't include the `courseNumber` column at all. If re-imported back into Autolab, the new values of `courseLecture` would take on the previous values of `section`, and the new values of `section` would be empty.

On another note, the [`detectAndImportRoster`](https://github.com/autolab/Autolab/blob/master/app/controllers/courses_controller.rb#L645) function, when processing the SIO/F16 format, is importing the `Course` column (the course number) as the lecture. I'm not sure whether this is intentional, but it means that the actual information on the lecture a student is in is discarded.

damianhxy:
## Description
- When exporting roster, use `@course.name` for the missing `courseNumber` column
- Make `ROSTER_COLUMNS_F16` and `ROSTER_COLUMNS_F20` use `Lecture` (rather than `Course`) for `courseLecture` field.
- Added entire `ROSTER_COLUMNS_F16` and `ROSTER_COLUMNS_F20` formats in comments
- Map `map[0]` (`Semester`) to `-1` since it is not used
- Fix mapping for `grading_policy` in `ROSTER_COLUMNS_F16` format

## Motivation and Context

The "General Autolab Format" is described by `Semester,email,last_name,first_name,school,major,year,grading_policy,courseNumber,courseLecture,section`.

Inside `parse_roster_csv`, the corresponding fields are extracted (except for `Semester` and `courseNumber`, since the data is stored in `@course`). However, when exporting a roster, the `courseNumber` column is not outputted,  which causes the columns to shift if we reimport the data (`courseLecture` will now take on the value of `section`, and `section` will be blank). This PR fixes this by using `@course.name` for the `courseNumber` column.

Furthermore, when processing Roster formats `ROSTER_COLUMNS_F16` and `ROSTER_COLUMNS_F20`, the `Course` field, instead of the `Lecture` field, is mapped to `courseLecture`, meaning that we are recording the Course name (e.g. `15122`) as the Lecture number (which should be `1`, `2`, etc.). In older format `ROSTER_COLUMNS_S15`, the `courseLecture` was correctly mapped to the `Lecture` field.

Fixes #997.

### Test 1: Ensure General Autolab Format still works correctly

**Roster**

`M22,xho@foo.bar,Ho,Damian,SCS,CS,2025,L,15122,1,A`

Screen Shot 2022-07-12 at 12 42 06

### Test 2: Check that Export + Import works correctly

**Roster (Note the 4 undropped students)**

Screen Shot 2022-07-11 at 14 30 45

**Before PR: Export + Import**

Observe that the section letter shifted into the lecture column.

Screen Shot 2022-07-11 at 14 31 42

**After PR: Export + Import**

Screen Shot 2022-07-11 at 14 32 03

### Test 3: Check that CMU Roster Format works correctly

**Roster**

```
Semester,Course,Section,Lecture,Mini,Last Name,Preferred/First Name,MI,Andrew ID,Email,College,Department,Major,Class,Graduation Semester,Units,Grade Option,QPA Scale,Mid-Semester Grade,Primary Advisor,Final Grade,Default Grade,Time Zone Code,Time Zone Description,Added By,Added On,Confirmed,Waitlist Position,Units Carried/Max Units,Waitlisted By,Waitlisted On,Dropped By,Dropped On,Roster As Of Date
M22,15122,A,1,N,Ho,Damian,,xho,xho@andrew.cmu.edu,SCS,CS,CS,3,F25,0,L,4+,,Test Advisor,,,EST,Eastern Standard Time (GMT-5:00),fake_andrewid,11 Jul 2022,Y,,,,,,,11 Jul 2022 12:00 AM
```

**Before PR**

Observe that Course number appears in the lecture field.

Screen Shot 2022-07-11 at 14 39 39

**After PR**

Screen Shot 2022-07-11 at 14 39 14

## Types of changes

## Checklist:

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -730,19 +730,19 @@ private
     case parsedRoster[0].length
     when ROSTER_COLUMNS_F20
       # In CMU S3 roster. Columns are:
-      # Semester(0), Course(1), Section(2), (Lecture)(3), (Mini-skip)(4),
+      # Semester(0), Course(1), Section(2), Lecture(3), (Mini-skip)(4),
       # Last Name(5), Preferred/First Name(6), (MI-skip)(7), Andrew ID(8),
       # (Email-skip)(9), College(10), (Department-skip)(11), Major(12),
       # Class(13), Graduation Semester(skip)(14), Units(skip)(15), Grade Option(16), ...
-      map = [0, 8, 5, 6, 10, 12, 13, 16, -1, 1, 2]
+      map = [0, 8, 5, 6, 10, 12, 13, 16, -1, 3, 2]
       select_columns = ROSTER_COLUMNS_F20
     when ROSTER_COLUMNS_F16
       # In CMU S3 roster. Columns are:
-      # Semester(0), Course(1), Section(2), (Lecture-skip)(3), (Mini-skip)(4),
+      # Semester(0), Course(1), Section(2), Lecture(3), (Mini-skip)(4),
       # Last Name(5), First Name(6), (MI-skip)(7), Andrew ID(8),
       # (Email-skip)(9), School(10), (Department-skip)(11), Major(12),
       # Year(13), (skip)(14), Grade Policy(15), ...
-      map = [0, 8, 5, 6, 10, 12, 13, 15, -1, 1, 2]
+      map = [0, 8, 5, 6, 10, 12, 13, 15, -1, 3, 2]
       select_columns = ROSTER_COLUMNS_F16
     when ROSTER_COLUMNS_S15
       # In CMU S3 roster. Columns are:
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -278,8 +278,9 @@ class CoursesController < ApplicationController
     @cuds.each do |cud|
       user = cud.user
       # to_csv avoids issues with commas
-      output += [@course.semester, cud.user.email, user.last_name, user.first_name, cud.school,
-                 cud.major, cud.year, cud.grade_policy, cud.lecture, cud.section].to_csv
+      output += [@course.semester, cud.user.email, user.last_name, user.first_name,
+                 cud.school, cud.major, cud.year, cud.grade_policy,
+                 @course.name, cud.lecture, cud.section].to_csv
     end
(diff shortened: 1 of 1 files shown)


fanpu (approved):
Great work catching this and the changes looks good! Could you also update the comments to include the subsequent columns that were elided so people will be able to know the entire roster format in the future?

app/controllers/courses_controller.rb
```diff
@@ -726,34 +730,44 @@ def detect_and_convert_roster(roster)
...
       select_columns = ROSTER_COLUMNS_S15
     else
       # No header row. Columns are:
-      # Semester(0), Email(1), Last Name(2), First Name(3), School(4),
-      # Major(5), Year(6), Grade Policy(7), (skip)(8), Lecture(9),
-      # Section(10), ...
+      # Semester(0 - skip), Email(1), Last Name(2), First Name(3), School(4),
+      # Major(5), Year(6), Grade Policy(7), Course(8 - skip), Lecture(9),
+      # Section(10)
       return parsedRoster
```
damianhxy, line 771: To be fully consistent with what we're doing with SIO rosters (i.e. dropping the Semester and Course columns here, even though they will be ignored later on anyway), we should technically use a "map" here `[-1, 1, 2, 3, 4, 5, 6, 7, -1, 9, 10]`. However, this doesn't make a difference unless for some reason we make use of those columns in the future.

fanpu:
Updated changes looks good, don't worry about the stuff from S15 too, no one should be using those now

PR #1411, opened 2021-11-01
Switch some `after_save` hooks to `after_commit`

oliverli:
## Description

Switched some `after_save` hooks that do not require strict data dependency from the DB (e.g. methods that do not write to the database, only filesystem) to `after_commit`, as recommended by Chaskiel

## Motivation and Context

Chaskiel's email dated Oct 21, 2021, 11:58 AM: MySQL deadlocks

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation, which is located at [Autolab Docs](https://github.com/autolab/docs)
- [ ] I have updated the documentation accordingly, included in this PR

## Other issues / help required

If unsure, feel free to submit first and we'll help you along.

PR #1560, opened 2022-07-11
Do not run :log_entry on :destroy

damianhxy:
## Description
Inside `score.rb`, restrict `:log_entry` to run only on `:create` and `:update`

## Motivation and Context

In #1411, several `after_save` hooks were changed to `after_commit`. Notably, `after_commit` hooks run when a record has been destroyed.

Hence, when destroying a course, `:log_entry` is called when `Score` records are eventually destroyed. However, `:log_entry` includes a logging statement which prints `problem.name` -- but the corresponding `Problem` record has already been destroyed. This results in an exception (although the course seems to be destroyed regardless).

Thus, this PR restricts `:log_entry` to run only on `:create` and `:update` (which was the behavior in the past anyway).

## How Has This Been Tested?
- Create a new course and create a new assessment by importing `hello.tar`
- Make a submission
- Delete the course successfully
- Also, use `byebug` to check that `:log_entry` is called when creating a submission / regrading a submission.

**Before PR**

Screen Shot 2022-07-12 at 01 48 37

**After PR**

Course deletes successfully.

## Types of changes

## Checklist:

### DIFF
--- a/app/models/score.rb
+++ b/app/models/score.rb
@@ -22,7 +22,7 @@ class Score < ApplicationRecord
   validates(:problem_id, uniqueness: { scope: :submission_id })
   validates :grader_id, presence: true
 
-  after_commit :log_entry
+  after_commit :log_entry, on: [:create, :update]
 
   def self.find_with_feedback(*args)
     with_exclusive_scope { find(*args) }
--- a/app/models/score.rb
+++ b/app/models/score.rb
@@ -23,7 +23,6 @@ class Score < ApplicationRecord
   validates :grader_id, presence: true
 
   after_commit :log_entry, on: [:create, :update]
-  after_commit :log_delete, on: [:destroy]
 
   def self.find_with_feedback(*args)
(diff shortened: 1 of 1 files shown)


app/models/score.rb
```diff
@@ -59,6 +60,10 @@ def log_entry
     " #{setter}")
   end
 
+  def log_delete
+    COURSE_LOGGER.log("Score #{id} DESTROYED for #{submission.course_user_datum.user.email}")
```
victorhuangwq, line 64: I don't think we should log this, mainly because it can take a while for courses with more students, and it's not particularly useful.

victorhuangwq (changes requested):
Solves the issue that you mentioned. But I don't think we should log the deletes.

victorhuangwq (approved):
LGTM

20wildmanj (approved):
Verified that fix works when deleting a course, verified :log_entry is being called when creating/regrading a submission.

PR #1562, opened 2022-07-14
Disable user data fields during CUD edit

damianhxy:
## Description
- Disable `:email`, `:first_name` and `:last_name` fields on edit CUD page, but enable on new CUD page

## Motivation and Context
In #877, a hotfix was applied to address a bug which prevented users from adding students to a course. However, the modified partial is also used by the edit CUD page. This means that when editing CUDs, the `email`, `first name`, and `last name` of students can be changed. As commented by @cg2v, it doesn't make sense for the fields to be edited here (`first name` and `last name` can be edited via manage users, `email` shouldn't be edited at all).

This PR adds a local parameter when calling the `fields` partial to determine whether the above mentioned fields should be disabled.

## How Has This Been Tested?
- When adding user to course, `email`, `first name` and `last name` fields should be editable
- When editing course user, `email`, `first name` and `last name` fields should **not** be editable

## Types of changes

## Checklist:


20wildmanj (approved):
- Verified editing first and last name disabled on CUD works properly

Just have a question now, I guess this means instructors wouldn't be able to change first/last names of students (since I assume instructors usually aren't given admin perms), so I wonder if instructors care about not having that ability?

damianhxy:
> Just have a question now, I guess this means instructors wouldn't be able to change first/last names of students (since I assume instructors usually aren't given admin perms), so I wonder if instructors care about not having that ability?

I think it should not matter. After all, when instructors add students, the name fields are auto populated after you enter the email, so it's not really something they should be editing.

Screen Shot 2022-07-17 at 16 10 20

Besides, the name is something that students themselves can edit, so there should be no need for instructors to be involved in the process.

20wildmanj:
> > Just have a question now, I guess this means instructors wouldn't be able to change first/last names of students (since I assume instructors usually aren't given admin perms), so I wonder if instructors care about not having that ability?
> 
> I think it should not matter. After all, when instructors add students, the name fields are auto populated after you enter the email, so it's not really something they should be editing.
> 
> Screen Shot 2022-07-17 at 16 10 20
> 
> Besides, the name is something that students themselves can edit, so there should be no need for instructors to be involved in the process.



> > Just have a question now, I guess this means instructors wouldn't be able to change first/last names of students (since I assume instructors usually aren't given admin perms), so I wonder if instructors care about not having that ability?
> 
> I think it should not matter. After all, when instructors add students, the name fields are auto populated after you enter the email, so it's not really something they should be editing.
> 
> Screen Shot 2022-07-17 at 16 10 20
> 
> Besides, the name is something that students themselves can edit, so there should be no need for instructors to be involved in the process.

Alright that makes sense, sounds good to me!

PR #1572, opened 2022-07-25
Metrics hotfix

damianhxy:
## Description
- Don't subtract one when calculating `@max_consecutive_assessments`
- Use string instead of symbol to access hash

## Motivation and Context
With the recent Rails upgrade, in `risk_condition#get_current_for_course`, `RiskCondition.where(course_id: course_id, version: max_version)` now returns a `Hash` instead of `HashWithIndifferentAccess`. Hence, we can no longer use symbols interchangeably with strings as keys. This PR modifies the Hash accesses to use string keys.

Furthermore, since consecutive assessments are inclusive, we should not subtract one when calculating  `@max_consecutive_assessments`, else we can't set the grades dropping metric to cover all assignments.

## How Has This Been Tested?
Tested workflow for metrics improvement.

Essentially, define categories of 2 assignments to test the following metrics (and considered categories feature)
- 1 grace day used
- 50% drop over 2 assignments
- Did not submit 1 assignment
- 1 assignment below 51%

Tips and reminders
- Remember to set max grace days for assignments where you're making a student use grace days
- Use annotations to award marks (after creating a problem)
- For drop metric, order is by `due_date ASC`

## Types of changes

## Checklist:

### DIFF
--- a/app/controllers/metrics_controller.rb
+++ b/app/controllers/metrics_controller.rb
@@ -10,7 +10,7 @@ class MetricsController < ApplicationController
     @max_consecutive_assessments = if course_max.nil?
                                      0
                                    else
-                                     course_max[1] - 1
+                                     course_max[1]
                                    end
   end
 
--- a/app/models/watchlist_instance.rb
+++ b/app/models/watchlist_instance.rb
@@ -360,8 +360,8 @@ class WatchlistInstance < ApplicationRecord
         case condition.condition_type
 
         when "grace_day_usage"
-          grace_day_threshold = condition.parameters[:grace_day_threshold].to_i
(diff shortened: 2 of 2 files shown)


damianhxy:
For more context: The Rails 6.0.5 migration resulted in the use of Psych 4, which doesn't allow the storage of `ActiveSupport::HashWithIndifferentAccess`. Hence in #1552 I called `to_hash` to store plain `Hash`es. However, the retrieval code was not changed to accommodate for this, and still attempted to use Symbols interchangeably.

victorhuangwq (approved):
LGTM. Changes seems to address the issue mentioned

PR #1578, opened 2022-08-19
Disable version switching feature when displaying autograder output

damianhxy:
## Description
When displaying autograder output, do not attempt to match pathname for version switching purposes

## Motivation and Context
Currently, when displaying the autograder output for autograded archives,
* `@displayFilename` is `Autograder Output`
* However, this is not an actual file in the archive, so we are likely to fail to find any matches (even for the current version) -- unless it happens to be an actual file
* This can lead to `matchedVersions` being empty and `@curVersionIndex` being `nil`, resulting in errors when we attempt to define `@prevVersion` and `@nextVersion`

Hence, this PR skips the matching in such cases, letting `@prevVersion` and `@nextVersion` stay as `nil`

## How Has This Been Tested?
* Created a modified Hello assessment which accepts a tar-ed `hello.c` file (unzip this: [hello.tar.zip](https://github.com/autolab/Autolab/files/9384321/hello.tar.zip))
* Make several submissions of the solution (unzip this: [hello.tar.zip](https://github.com/autolab/Autolab/files/9384326/hello.tar.zip))

Ensure that Autograder Output now displays correctly (note that the arrows will always be grayed out)

Screen Shot 2022-08-20 at 01 34 23

Ensure that normal files still display correctly

Screen Shot 2022-08-20 at 01 34 18

## Types of changes

## Checklist:

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -303,6 +303,10 @@ class SubmissionsController < ApplicationController
       }]
     end
 
+    viewing_autograder_output = params.include?(:header_position) &&
+                                (params[:header_position].to_i == -1) &&
+                                !@submission.autograde_file.nil?
+
     # Adds autograded file as first option if it exist
     # We are mapping Autograder to header_position -1
     unless @submission.autograde_file.nil?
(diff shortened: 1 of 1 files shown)


app/controllers/submissions_controller.rb
```diff
@@ -317,6 +317,7 @@ def view
        !@submission.autograde_file.nil?
 
       file = @submission.autograde_file.read || "Empty Autograder Output"
+      is_autograder_output = true
```
victorhuangwq, line 320: Seems like this variable `is_autograder_output` is initialized only if it is `true`?
If so we are relying on the default behavior of `unless` in line 497 when it is uninitialized, which I think might be risky. Lmk what you think

victorhuangwq (changes requested):
Simple PR, but I have a small question

app/controllers/submissions_controller.rb
damianhxy: imo, it's not really risky since the default behavior is well-defined and intuitive. But I've refactored the code to make the logic more obvious.

victorhuangwq (approved):
LGTM

PR #1579, opened 2022-08-23
Fix remarks for annotations on autograder output

damianhxy:
## Description
When `annotation.position = -1`, set `filename` to `Autograder Output` instead.

## Motivation and Context
A bug noted by the 122 team is that annotations on the autograder output (when `annotation.position = -1`) appear to reference the last file in the file tree instead.

Looking at `viewFeedback.html.erb`, the current logic for choosing `filename` is `@files && annotation.position ? Archive.get_nth_filename(@files, annotation.position) : @submission.filename`.

When we have a single file submission, `@files` is `nil` and so `filename` is `@submission.filename` (i.e. the last file).

Similarly, with an archive submission, `filename` is `Archive.get_nth_filename(@files, annotation.position)`, which evaluates to `@files[annotation.position][:pathname]` where `annotation.position = -1`. In particular, we are accessing index `-1`, i.e. the last file.

This PR remedies the logic by explicitly handling the `annotation.position = -1` case.

## How Has This Been Tested?
- Create an autograded assessment that accepts archives (for simplicity, use the assessment from #1578).
- Add an additional `Style` problem.
- Make a submission, annotate the autograder output, and release grades.
- View feedback page (by clicking on the autograded score on the view assessment page)

**Before changes**
Screen Shot 2022-08-24 at 00 13 17

**After changes**
Screen Shot 2022-08-24 at 00 07 04

## Types of changes

## Checklist:

### DIFF
--- a/app/views/assessments/viewFeedback.html.erb
+++ b/app/views/assessments/viewFeedback.html.erb
@@ -16,19 +16,26 @@ Graded by: <%= if @score.grader && @score.grader.user then @score.grader.user.em
 
 <% end %>
 <!--
-Yes, feedback if per problem but we show all annotations for this submission anyway. While this is not semantically correct, it offers a better UX than we had before. If we just show annotations attached to this problem, then where would we show annotations not attached to a problem at all?
+Yes, feedback is per problem but we show all annotations for this submission anyway. While this is not semantically correct, it offers a better UX than we had before. If we just show annotations attached to this problem, then where would we show annotations not attached to a problem at all?
 -->
 
 <% if @submission.annotations.count > 0 and @problemReleased %>
   <h2 style="margin-top: 1em">Remarks</h2>
 
   <ul>
+  <!-- TODO: Make order of annotations consistent with order of file list  -->
   <% @submission.annotations.each do |annotation| %>
     <li>
       <%= link_to annotation.as_text, [:view, @course, @assessment, @submission, header_position: annotation.position] %>
       <% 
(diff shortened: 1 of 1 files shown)


app/views/assessments/viewFeedback.html.erb
```diff
@@ -16,19 +16,26 @@ Graded by: <%= if @score.grader && @score.grader.user then @score.grader.user.em
...
 <!--
-Yes, feedback if per problem but we show all annotations for this submission anyway. While this is not semantically correct, it offers a better UX than we had before. If we just show annotations attached to this problem, then where would we show annotations not attached to a problem at all?
+Yes, feedback is per problem but we show all annotations for this submission anyway. While this is not semantically correct, it offers a better UX than we had before. If we just show annotations attached to this problem, then where would we show annotations not attached to a problem at all?
 -->
 
 <% if @submission.annotations.count > 0 and @problemReleased %>
   <h2 style="margin-top: 1em">Remarks</h2>
 
   <ul>
+  <!-- TODO: Make order of annotations consistent with order of file list  -->
```
victorhuangwq, line 26: Ah so this is a TODO for the future?

victorhuangwq (approved):
LGTM

app/views/assessments/viewFeedback.html.erb
damianhxy: Yup, it is a TODO to track the second "bug" reported by the 122 team (The order of annotations in the student feedback when they click on the score is not the same as the order when you look at the file page)

PR #1584, opened 2022-09-11
Fix creation of missing submissions

damianhxy:
## Description
- Only run autocomplete code on create submissions page when CUD is `nil`.

## Motivation and Context
As reported in `INC1029200`, an internal server error occurs when one attempts to fill in blank submissions for missing submissions.

This is because the autocomplete js code attempts to run even when CUD is set (so there's no need to autocomplete anything). However, the required data is not initialized in such a case, so we attempt to iterate over `nil`.

**Fill in Empty Submissions feature**
Screen Shot 2022-09-10 at 20 10 02

**Server only initializes autocomplete data when CUD is not nil**
Screen Shot 2022-09-10 at 20 08 08

To remedy this, don't run autocomplete js code unless there is a need to.

## How Has This Been Tested?
- On an assessment with missing submissions, attempt to fill in empty submissions

**Before changes**
Screen Shot 2022-09-10 at 20 12 49

**After changes**
Screen Shot 2022-09-10 at 20 12 39

## Types of changes

## Checklist:

### DIFF
--- a/app/views/components/_autocomplete.html.erb
+++ b/app/views/components/_autocomplete.html.erb
@@ -1,5 +1,5 @@
 <script type="application/javascript">
-  /* requires @usersEncoded and @users to be set, i.e. via retrieve_autocompletion_data! */
+  /* requires @usersEncoded and @users to be set, i.e. via Course.get_autocomplete_data */
   jQuery(function() {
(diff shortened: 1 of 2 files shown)


michellexliu (approved):
lgtm!

PR #1588, opened 2022-09-17
[hotfix] Parameterize category names for allowlist

michellexliu:
## Description
Previously, nothing would happen when a user clicked on the switch icon between the "Included" and "Excluded" list for assignment categories that have spaces in the names (e.g. "Optional Assignments"). This was caused because the category names were being used as ID's, which caused the jQuery to return undefined for categories with spaces. This was fixed by altering the jQuery to look for `[id='included_${category}']`, rather than individual id tags (which looks for id's separated by spaces separately, causing nothing to be returned).

## How Has This Been Tested?
* Toggled included/excluded for both categories with and without spaces and made sure it saved properly.
* Category names with other non-alphanumeric characters (e.g. $, %, etc.)
* Category names with trailing whitespaces or multiple non-alphanumeric characters in a row

## Types of changes

## Checklist:

### DIFF
--- a/app/assets/javascripts/metrics_config.js
+++ b/app/assets/javascripts/metrics_config.js
@@ -2,13 +2,6 @@
 //= require semantic-ui
 
 // metrics api endpoints
-const parameterize = (str) => {
-  return str.trim().toLowerCase()
-          .replace(/[^a-z0-9\-_]+/g, "_")
-          .replace(/_{2,}/g, "_")
-          .replace(/^_|_$/g, "");
-}
-
 const metrics_config_endpoints = {
 	get_category: 'get_watchlist_category_blocklist',
 	update: 'update_watchlist_configuration',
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/metrics_config.js
```diff
@@ -14,9 +21,9 @@ $.getJSON(metrics_config_endpoints['get_category'],function(data, status){
 		excluded_categories = [];
 
 		if (status=='success') {
-			data.forEach(category => {
```
damianhxy, line 17: Nit: Consistent indentation here would be good

app/assets/javascripts/metrics_config.js
```diff
@@ -2,6 +2,13 @@
 //= require semantic-ui
 
 // metrics api endpoints
+const parameterize = (str) => {
```
damianhxy, line 5: Just a note, the [HTML5 spec](https://www.w3.org/TR/2011/WD-html5-20110525/elements.html#the-id-attribute) only requires that IDs be unique and not contain spaces. Parametrizing in this manner (stripping special characters) might be more prone to name collisions (which results in funky behavior).

Perhaps not best practice, but jQuery can actually [select IDs with spaces in them](https://stackoverflow.com/questions/6802765/jquery-dealing-with-a-space-in-the-id-attribute). Would that be a simpler approach?

damianhxy (approved):
LGTM!

PR #1590, opened 2022-09-17
Fix order of annotations on feedback view and file view to be consistent

20wildmanj:
## Description

Previously, the annotations on the "viewFeedback" page were sorted by creation date of the annotation, while the annotations on the file view were sorted by line number, which meant order wasn't maintained among the two views.

Previous feedback and file view:
Screen Shot 2022-09-17 at 15 39 09
Screen Shot 2022-09-17 at 15 38 45

I added sorting to both `viewFeedback.html.erb` and `_annotation_pane.html.erb` to first sort by the filename the annotation was added to, and then to sort by the line number.  The lambda function `get_correct_filename` was added to `assessments_controller.rb` since in multiple instances, we need to get the proper filename / change the filename if it's the autograded output or it's an archived file, in `viewFeedback.html.erb`. `submissions_controller.rb` also received similar code to append the proper filename to `problemSummaries`.

Revised annotation ordering for feedback and file view:
Screen Shot 2022-09-17 at 15 38 18
Screen Shot 2022-09-17 at 15 38 30

## Motivation and Context

This issue was brought up by the 122 team, as they saw that "the order of annotations in the student feedback when they click on the score is not the same as the order when you look at the file page." There was also a TODO in `viewFeedback.html.erb` to make order of annotations consistent with the order of the file list.

## How Has This Been Tested?

- create / use an assessment that has multiple files (an autograded lab with additional problems is preferred)
- as a student account, submit a solution to the assessment, make sure it gets properly autograded
- as instructor, go to "grade submission" and click on the student submission, 
- add multiple annotations to multiple files in the student submission, and try adding some annotations in a random order with regards to line number (for example, add one annotation to the bottom of autograder output, and then add one annotation to the middle of the main submission file, then another at the bottom of the main submission file)
- make sure to add feedback for the problem in "viewGradesheet" for the student
- as a student, go to the assessment, and under the handin with the problem with annotations click on the score like so:
  - Screen Shot 2022-09-17 at 16 08 56
- Verify that the order of the annotations is by file name and then by line number
- Go to the file view, verify that the order of the annotations is the same

## Types of changes

## Checklist:

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -461,10 +461,19 @@ class SubmissionsController < ApplicationController
       value = annotation.value || 0
       line = annotation.line
       problem = annotation.problem ? annotation.problem.name : "General"
-
+      filename = if annotation.position == -1
+                   # position -1 maps to the Autograder Output
+                   "Autograder Output"
+                 elsif @files && annotation.position != 0
+                   # if the submission is an archive, use filename in archive;
+                   # otherwise, use submission filename
+                   Archive.get_nth_filename(@files, annotation.position)
+                 else
+                   @submission.filename
+                 end
       @problemSummaries[problem] ||= []
       @problemSummaries[problem] << [description, value, line, annotation.submitted_by,
-                                     annotation.id, annotation.position]
+                                     annotation.id, annotation.position, filename]
 
       @problemGrades[problem] ||= 0
       @problemGrades[problem] += value
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -461,16 +461,7 @@ class SubmissionsController < ApplicationController
       value = annotation.value || 0
       line = annotation.line
       problem = annotation.problem ? annotation.problem.name : "General"
-      filename = if annotation.position == -1
-                   # position -1 maps to the Autograder Output
-                   "Autograder Output"
-                 elsif @files && annotation.position != 0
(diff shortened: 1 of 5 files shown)


app/controllers/submissions_controller.rb
```diff
@@ -461,10 +461,19 @@ def view
...
+      filename = if annotation.position == -1
+                   # position -1 maps to the Autograder Output
+                   "Autograder Output"
+                 elsif @files && annotation.position != 0
+                   # if the submission is an archive, use filename in archive;
+                   # otherwise, use submission filename
+                   Archive.get_nth_filename(@files, annotation.position)
+                 else
+                   @submission.filename
+                 end
```
victorhuangwq, line 473: I'm curious as to why do we need a duplication of this code in this controller as well?

victorhuangwq (approved):
Tested on nightly and works as expected. LGTM.
Left a comment on one line of the code.

victorhuangwq (approved):
LGTM!

damianhxy:
Screen Shot 2022-09-28 at 17 26 43

There seems to be an issue when annotating archives -- but I'm not sure if this was existing behavior

20wildmanj:
> Screen Shot 2022-09-28 at 17 26 43
> 
> There seems to be an issue when annotating archives -- but I'm not sure if this was existing behavior

Looks like this PR caused this, will make a hotfix, thanks for finding the bug. @michellexliu for visibility since you're requesting computing services to merge this PR into deployment.

PR #1593, opened 2022-09-19
(WIP) Add SAML Authentication functionality

damianhxy:
## Description
TBC

## Motivation and Context
Add SAML authentication to Autolab for universities that use Microsoft IdP.

Note: This currently overrides shibboleth authentication. Work is needed to add some kind of toggle to switch between the two types of authentication. The current changes also appear to comment out a crucial part of Autolab's CSRF protection (inside `verify_authenticity_token`). Ideally, we should be using `action_no_auth` to selectively disable the protection for routes that do not require authentication.

## How Has This Been Tested?
TBC

## Types of changes

## Checklist:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -60,6 +60,8 @@ gem 'omniauth', '>=1.2.2'
 gem 'omniauth-facebook', '>=2.0.0'
 gem 'omniauth-google-oauth2', '>=0.2.5'
 gem 'omniauth-shibboleth', '>=1.1.2'
+gem 'omniauth-saml', '>=1.7.0'
+gem 'omniauth-rails_csrf_protection'
 
 # OAuth2 authentication
 gem 'oauth2'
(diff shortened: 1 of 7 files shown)


cg2v:
Apart from the CSRF thing, the most significant problem I see here is that you are embedding microsoft's (OASIS WS-Security) attribute mappings, which are not the same as the ones used by IdPs in the incommon federation. The attribute uris need to be configurable by the end institution, and there should be some exception handling or other validation. it may make sense to start parameterizing the LDAP lookups too and remove the embedded fixed domain (andrew.cmu.edu)

Gemfile
```diff
@@ -60,6 +60,8 @@ gem 'omniauth', '>=1.2.2'
 gem 'omniauth-facebook', '>=2.0.0'
 gem 'omniauth-google-oauth2', '>=0.2.5'
 gem 'omniauth-shibboleth', '>=1.1.2'
+gem 'omniauth-saml', '>=1.7.0'
+gem 'omniauth-rails_csrf_protection'
```
damianhxy, line 64: Should be unnecessary with omniauth v2.0.0

PR #1599, opened 2022-09-24
Remove success flash for export assessment

damianhxy:
## Description
Remove success flash for export assessment

## Motivation and Context
Currently, we are setting a success flash after exporting an assessment. However, we are not redirecting, so the flash only shows up on the **next** page load.

However, we can't use flash.now either, since we are sending a file, so the user won't see the flash.

Ultimately, it's best to simply remove the flash altogether.

## How Has This Been Tested?
**Before** (Flash only appears after you navigate to another page)
Screen Shot 2022-09-24 at 17 42 51

**After**
(No more flash)

## Types of changes

## Checklist:


20wildmanj (approved):
LGTM - also if you receive the tarball from autolab that pretty much guarantees the export was successful anyways, so makes sense to just remove.

PR #1603, opened 2022-09-24
Changed submit button on Act as user page to 'Act as user'

najclark:
## Motivation and Context

Changed title of submit button on Act as user page from 'sudo' to 'act as user' for clarity.

## How Has This Been Tested?

Before:
Screen Shot 2022-09-24 at 2 51 03 PM

After:
Screen Shot 2022-09-24 at 2 50 12 PM

## Types of changes

## Checklist:


app/views/course_user_data/sudo.html.erb
```diff
@@ -14,6 +14,6 @@
     <label for="student_autocomplete">Start typing student name or email</label>
   </div>
   <%= hidden_field_tag(:sudo_id) %>
-  <%= submit_tag "Sudo", {:value=>"Sudo", :class=> "btn primary"} %>
+  <%= submit_tag "Act as user", {:value=>"Act as user", :class=> "btn primary"} %>
```
damianhxy, line 17: `:value=>...` seems to be redundant, so we can get rid of it.

damianhxy: There also appears to be an unused sudo page under `/users/sudo.html.erb`, we can remove that as well.

damianhxy (approved):
Left some comments, but LGTM!

app/views/course_user_data/sudo.html.erb
victorhuangwq: @najclark

PR #1604, opened 2022-09-24
got rid of unintentional horizontal scroll caused by menu

michellexliu:
Before:
Screen Shot 2022-09-24 at 2 52 53 PM

After: 
Screen Shot 2022-09-24 at 2 52 42 PM


damianhxy (approved):
Confirmed that it fixes the issue, lgtm!

PR #1606, opened 2022-09-24
Overflow scroll within feedback div instead of entire page

michellexliu:
Before:
Screen Shot 2022-09-24 at 3 05 18 PM

After:
Screen Shot 2022-09-24 at 3 05 05 PM


victorhuangwq (approved):
Lgtm

PR #1607, opened 2022-09-24
Make attachment upload more understandable, indicate if attachment unreleased

20wildmanj:
Previously, attachment upload was very confusing, with the buttons not being helpful / informative. Also, if an attachment was unreleased, the instructor wouldn't know unless they went into the edit page for the attachment. I tried to make the upload form give more informative information for each field, and added a (Not Released) to any attachments not released when looking at the assessment page.

Before:
(Earfs is unreleased, but you would never know...)
Screen Shot 2022-09-24 at 15 11 12
Screen Shot 2022-09-24 at 15 11 23

After:
Still kinda ugly, could polish if that is desired:
Screen Shot 2022-09-24 at 15 08 32
New attachment:
Screen Shot 2022-09-24 at 15 17 05
Edit attachment:
Screen Shot 2022-09-24 at 15 16 49


damianhxy:
I think the italics for `(Not Released)` is unnecessary. Also it seems that you swapped the screenshots for New attachment and Edit attachment.

app/views/attachments/_form.html.erb
```diff
@@ -1,30 +1,40 @@
...
-					<%= f.file_field :file, class: "upload_embedded" %>
-			</span>
-		</label>
-		<input type="text" id="upload_file" readonly value="Select to Upload a File.">
-  </div>
-	<% else %>
-		<%= f.text_field :mime_type,
-				help_text: "Note: If you change the Mime type of a file, you might need to clear your browser's cache in order for you to see the change." %>
-	<% end %>
+	<ul>
```
damianhxy, line 4: Is the purpose of `<ul>` to make the lines the same length?

damianhxy (approved):
Left some comments, but otherwise lgtm!

app/views/attachments/_form.html.erb
20wildmanj: Pretty much

PR #1602, opened 2022-09-24
Improve formatting of handin history page

damianhxy:
## Description
Remove unnecessary italics, introduce line breaks, improve handling of infinite extensions and add pluralization.

## Motivation and Context
As per description.

## How Has This Been Tested?
**Before**
2 days extension
Screen Shot 2022-09-24 at 14 39 58
Infinite extension
Screen Shot 2022-09-24 at 14 39 23

**After**
2 days extension
Screen Shot 2022-09-24 at 14 45 31
Infinite extension
Screen Shot 2022-09-24 at 14 45 45

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] I have run rubocop for style check. If you haven't, run `overcommit --install && overcommit --sign` to use pre-commit hook for linting
- [ ] My change requires a change to the documentation, which is located at [Autolab Docs](https://github.com/autolab/docs)
- [ ] I have updated the documentation accordingly, included in this PR

## Other issues / help required

If unsure, feel free to submit first and we'll help you along.

PR #1609, opened 2022-09-24
Compact submission history, increase visibility of "see all submissions"

20wildmanj:
Made submission history more compact (removed lots of unnecessary space and change submission titles from h3 to h4). Also made "See all submission history" an h6 for better visibility

Before:
image
Look at all that white space!
Screen Shot 2022-09-24 at 16 12 52

After:
Screen Shot 2022-09-24 at 16 08 44
Screen Shot 2022-09-24 at 16 09 16


victorhuangwq:
Actually can't we just make this page like a table similar to the handin page

20wildmanj:
> Actually can't we just make this page like a table similar to the handin page

Like this?
Screen Shot 2022-09-24 at 16 20 59

victorhuangwq:
Exactly, cause it's basically the same thing

victorhuangwq:
In fact, just make a partial and link it at both places.

20wildmanj:
That could work yeah

victorhuangwq:
Let me tag the others for a quick feedback @michellexliu @damianhxy @najclark

20wildmanj:
Also ngl we should take a look at the current handin table and see how to make it not look janky when there are a lot of problems:
Screen Shot 2022-09-24 at 16 23 35
Screen Shot 2022-09-24 at 16 23 41
Screen Shot 2022-09-24 at 16 24 58

michellexliu:
Agreed, I think we can just make it a partial; for the handin table, I think the current layout is fine, but we should make it scrollable within the table rather than entire page (i.e. add an `overflow-x: auto` property). I also think we should just make the column width auto instead of fixed so that there's no wrapping. Thoughts?

20wildmanj:
Fair enough, that seems pretty reasonable. I think making the text in the handin table a bit more aligned would be good as well to go along with those changes (if you look at the example above the text is like jumping up and down column by column).

20wildmanj:
Made some progress, but want some preliminary feedback to try to plot a course to polish out the refresh @victorhuangwq  @michellexliu @najclark .

Which alignment do you prefer (top aligned or bottom aligned for headers):
Screen Shot 2022-09-24 at 17 26 57
Screen Shot 2022-09-24 at 17 27 50

Having the table rows be strictly one line or encompass more lines

One line:
Screen Shot 2022-09-24 at 17 26 57
Screen Shot 2022-09-24 at 17 27 01

Multiple lines:
Screen Shot 2022-09-24 at 17 28 04
Screen Shot 2022-09-24 at 17 31 14

victorhuangwq:
Also @damianhxy  seems to have done some adjustments here https://github.com/autolab/Autolab/pull/1602

victorhuangwq:
Personally I prefer:
- Top align
- Multiple lines

20wildmanj:
> Also @damianhxy seems to have done some adjustments here #1602

I see, though this is for a different part of the submission history so I don't think it affects this PR

20wildmanj:
I experimented getting rid of the filename and using the space to make the icons' purposes more clear. My thoughts are that the filename is not really necessary information, since it's literally just the same repeated "#{course_user_datum.user.email}_#{version}_#{assessment.handin_filename}", which doesn't really serve a use to the user, but just takes up space.

Instructor view:
Screen Shot 2022-09-25 at 17 05 56
Instructor view with lots of problems:
Screen Shot 2022-09-25 at 17 06 20

Student view:
Screen Shot 2022-09-25 at 17 06 44

victorhuangwq:
This is a good idea. I like the current look, it's clean and looks good

michellexliu:
Agreed. Perhaps we could save even more space by getting rid of the Ver column and changing "Download Submission" to something like `"Download Version #{version}"` or `"Download Submission #{version}"`?

victorhuangwq:
I think the version number on the left side makes it easy to look through the table though. and it doesn't take up that much space. My personal thought we be somehow better represent the submitted date and time

20wildmanj:
Maybe for now we can keep the version number and evaluate that part at a later date? I feel like this could be a good thing to all go over during the team meeting or sopmething.

But yeah right now the version time shows up as the entire timestamp with the timezone, which I can understand why it might be nice to see, but it is a long string.

victorhuangwq:
Yes. Personally I think we can just merge the current PR in, and decide if we want further modification later. This PR came out of our hacking session which is meant for fast and quick fixes. We can always refine later

app/views/assessments/_downloadSubmission.html.erb
```diff
@@ -1,23 +1,25 @@
 <div class="history-download-pane" style="margin-right: 30px;">
   <% if can_always_download or not (sub.assessment.exam or sub.assessment.course.exam_in_progress) then %>
-    <%= sub.filename %>
+    <!--<%= sub.filename %>-->
```
victorhuangwq, line 3: If this is no longer needed we can remove it

app/views/assessments/show.html.erb
```diff
@@ -318,16 +318,15 @@
...
     No Submissions yet!
   <% else %>
-    <% for submission in @submissions.first(limit=3) do %>
-      <%= render partial: "submission_summary_row",
-        locals: { submission: submission,
-        download_access: download_access } %>
-    <% end %>
-    <%= link_to "See all submissions", {:action => "history"} %></li>
-
+    <%= render partial: "submission_history_table", locals: {download_access: download_access, limit: 3} %>
```
victorhuangwq, line 327: Did GitHub mistook this line as a comment?

victorhuangwq (changes requested):
Some minor changes wrt dead code

app/views/assessments/show.html.erb
20wildmanj: I think so, it's actual code.

victorhuangwq (approved):
LGTM

victorhuangwq:
@20wildmanj

It seems that you didn't label the download button where we force it to be downloaded as a text. Refer to this link
https://nightly.autolabproject.com/courses/1337-f22/assessments/hello

image

Could you do a quick hotfix for this?

20wildmanj:
> @20wildmanj
> 
> It seems that you didn't label the download button where we force it to be downloaded as a text. Refer to this link https://nightly.autolabproject.com/courses/1337-f22/assessments/hello
> 
> image
> 
> Could you do a quick hotfix for this?

Gotcha, thanks for catching, will do!

20wildmanj:
I'm gonna make the text as "Download Submission" for downloading plain / text options to maintain clarity

20wildmanj:
Actually I am encountering an issue where if I submit a plaintext file to an assessment with a .tar format specification, I can upload the file, but downloading will cause an error, not necessarily something to fix here but I'll make an issue about it.

victorhuangwq:
> I'm gonna make the text as "Download Submission" for downloading plain / text options to maintain clarity

Wouldn't that be confusing to have two exactly the same text label of download submission?

Perhaps the one where we are forcing download as plain text can be download as plain/text

20wildmanj:
> > I'm gonna make the text as "Download Submission" for downloading plain / text options to maintain clarity
> 
> Wouldn't that be confusing to have two exactly the same text label of download submission?
> 
> Perhaps the one where we are forcing download as plain text can be download as plain/text
> 
> Let me know if I got the issue completely wrong though

Well, from the user's perspective, I feel as though they are just trying to download the submission, so whether or not they know that it's plaintext doesn't matter too much, especially since we aren't really giving them a choice anyways. 

Being able to download this file also right now depends on the handin specification. We are able to download files as plain when the format requested for handin is something like `hello.c`, in which case when we download the submission as "plain" it's still just `hello.c` so I feel like the added information is not that necessary. Let me know if this makes sense!

victorhuangwq:
If what you are saying is true I agree. Do you mind just checking the commit history for the line of code surrounding the download as plantext to see the commit message behind it (basically if there's any special intention behind why they offered this option in the first place)

20wildmanj:
> If what you are saying is true I agree. Do you mind just checking the commit history for the line of code surrounding the download as plantext to see the commit message behind it (basically if there's any special intention behind why they offered this option in the first place)

Seems like it's been around in some form since the beginning of the commit history:
Screen Shot 2022-09-28 at 17 12 36

So in the end I'm not really sure what the intention was.

PR #1613, opened 2022-09-27
retry loading file viewer with new header position when render error …

### LINKED ISSUE #1594: 500 Error when attempting to view non-supported binary files
Submitted by 122 Head TA. A 500 Error occurs when viewing a submission with a non-supported binary file.

Stack dump: [sub_273_internal_error_info (1).txt](https://github.com/autolab/Autolab/files/9628339/sub_273_internal_error_info.1.txt)

**Steps To Reproduce**
1. Make a submission including a binary file that's unsupported by Autolab (sample: [test.tgz](https://github.com/autolab/Autolab/files/9628451/test.tgz))
2. View the submission with a binary file
3. See error

**Expected behavior**
We should be able to still view the submission interface without being able to view the non-displayable binary file.

Chaskiel's advice: 
The problem seems to be that the first file in the archive is non-displayable binary file (in this case, some MacOS metadata).
You can work around this by editing the url and changing the value of header_position until the interface loads.

20wildmanj:
## Description

- Add an exception handler to `submissions_controller.rb` to catch when `format.html { render(:view) }` fails due to it trying to render a non-supported binary file
- Exception handler increments `header_position` (the file it loads) and redirects to the file view to try again until it either runs out of files it can view, which will redirect back to the assessment page, or it loads the first supported file available.
## Motivation and Context
- Fixes #1594
- Solution based on advice from Chaskiel (see issue)
## How Has This Been Tested?

- Submit an archive to an assessment(use this archive: [brokenarchive.tgz](https://github.com/autolab/Autolab/files/9659410/brokenarchive.tgz))
 - Note: this file contains .DS_Store and ._wifitransmitter.py, which aren't displayable and would normally cause an error
- Click on view source
- see that it loads wifitransmitter.py without issue (might take a couple seconds)
## Types of changes

## Checklist:


damianhxy:
Tested that it works. However, it is still possible to select the file from the speedgrader, which just results in an infinite loading screen.

I think a better fix is to change L358 of `submissions_controller` to `file = "Binary file not displayed" if !mm.present? || (!mm.text? && (mm.subtype != "pdf"))`. This is because MimeMagic is returning `nil` for files such as `._wifitransmitter.py`.

However, we can still keep your current changes too, as it would serve as a good backup in case of any errors

20wildmanj:
> Tested that it works. However, it is still possible to select the file from the speedgrader, which just results in an infinite loading screen.
> 
> I think a better fix is to change L358 of `submissions_controller` to `file = "Binary file not displayed" if !mm.present? || (!mm.text? && (mm.subtype != "pdf"))`. This is because MimeMagic is returning `nil` for files such as `._wifitransmitter.py`.
> 
> However, we can still keep your current changes too, as it would serve as a good backup in case of any errors

Unforunately, this line change by itself doesn't really work b/c this causes for example autograded output to no longer be displayed, as well as a bunch of other file types I tested such as .txt, .csv, .rs, .bf, .rb, .yml, etc (.c files and .py files display fine though). Not really sure why this is the case, will investigate.

Infinite loading is not great either though so there's probably some solution, will take a further look in a bit.

20wildmanj:
I tried experimenting with a lot of stuff, but it seems as though mimemagic just usually misses the file type, and standard exceptions in the controller didn't really work either. It seems as though having to go through the code data before rendering and ensuring it contains only utf-8 characters is the only surefire way of preventing the 500 error while allowing most file types through. 

I'm a little bit concerned about time-wise how long it takes to load the page when every line of code has to be checked, but it seemed like the time difference was negligible for files with hundreds of lines of code. Definitely still a last resort solution though.

app/controllers/submissions_controller.rb
```diff
@@ -549,9 +549,18 @@ def view
...
 
-    respond_to do |format|
-      format.html { render(:view) }
-      format.js
+    begin
+      respond_to do |format|
+        format.html { render(:view) }
+        format.js
+      end
+    rescue StandardError
```
damianhxy, line 557: Is this error handling still required? I suppose it helps with UX but is not strictly required?

app/views/submissions/_code_viewer.html.erb
```diff
@@ -44,24 +44,28 @@
...
-            <% rescue %>
-              </code></pre></div></div>
-              <% success = false %>
-              <% break %>
-            <% end %>
-          </div>
-          <div class="annotation-container" id="annotation-line-<%=index%>" data-line-id="<%=index%>">
-          </div>
+          <% else %>
+            <% success = false %>>
```
damianhxy, line 67: Nit: extra `>`

app/views/submissions/_code_viewer.html.erb
```diff
@@ -44,24 +44,28 @@
...
+            <div class="code-line" id="line-<%=index%>" data-line-id="<%=index%>">
+              <div class="line-sticky">
+                <div class="line-number"><%= index+1 %></div>
+                <div class="code-separator"></div>
+              </div>
+              <div class="add-comment">
+                <% if @cud.instructor? or @cud.course_assistant? then %><div class="add-button-container"><button class="add-button" id="add-btn-<%= index+1 %>"><i class="material-icons">add_comment</i></button></div><% end %>
+              </div>
+              <% begin %>
+              <div class="code"><pre><code class="<%= @displayFilename.split(".")[-1].downcase %>"><%= code %></code></pre></div>
```
damianhxy, line 57: The issue seems to lie with the `<%= code %>` block.

Actually, what we could do is to replace it with `<%= code.encode("UTF-8", invalid: :replace, replace: "") %>`. That way, we can render what's left of the file.

What do you think?

damianhxy (approved):
Verified that bad archive submission can now be viewed. Elegant way of handling non-utf-8 files by reusing existing logic!

I have left some comments and suggestions, but otherwise lgtm!

app/controllers/submissions_controller.rb
20wildmanj: Yeah it's not really required anymore, I'll see if removing it is fine.

20wildmanj: Yeah I guess with the changes to code_viewer we don't need to do the error handling

20wildmanj: Actually I take it back, some files still cause an error when trying to initially view the assessment so this error handling is required.

damianhxy: Hmm, what kind of files are not handled?

20wildmanj: oh nvm nvm actually I'm getting an error due to #1616 I think haha so we can remove

app/views/submissions/_code_viewer.html.erb
20wildmanj: Seems like Japanese works fine, so I'm confused as to what characters specifically cause this UTF-8 error.
This is what it looks like when trying to render what's left of the file:
Screen Shot 2022-10-06 at 21 21 40
Rejecting it entirely:
Screen Shot 2022-10-06 at 21 27 24

Which do you think looks better? Personally rejecting it looks better and I mean the data that is able to be rendered is not very useful.

20wildmanj: Screen Shot 2022-10-06 at 21 29 21
Japanese text rendering on nightly

app/controllers/courses_controller.rb
```diff
@@ -788,7 +788,7 @@ def detect_and_convert_roster(roster)
                "foo.bar"
              end
     (0..(numRows - 1)).each do |i|
-      (0..10).each do |j|
+      11.times do |j|
```
20wildmanj, line 791: don't know why rubocop brought this up now

app/views/submissions/_code_viewer.html.erb
damianhxy: I think we can just reject the files then. However, we might want to update the message - the error is that the file contains non UTF-8 characters, rather than non-ASCII

20wildmanj: Gotcha, I'll update the display.

damianhxy:
LGTM

PR #1615, opened 2022-09-28
Hide view writeup button when no writeup exists

damianhxy:
## Description
* Hide view writeup button when writeup does not exist
* Use flash instead of text to report when writeup does not exist

Note: Defining an analogous check for handouts is harder because the config file might override the default handout behavior

Also note: `flash.now` is necessary because we are rendering directly rather than redirecting (as otherwise the flash would persist for an additional page load). I have updated handout's code to use `flash.now`, but it appears that `flash` works too because the route is defined in a different file and that somehow makes a difference.

## Motivation and Context
Currently, the view writeup button is visible to students even if no writeup exists. Furthermore, the error text is unnecessarily large.

## How Has This Been Tested?
**Before**

Error Text
Screen Shot 2022-09-28 at 16 49 32

Button visible
Screen Shot 2022-09-28 at 16 49 48

**After**

Error Flash
Screen Shot 2022-09-28 at 16 49 58

Button not visible
Screen Shot 2022-09-28 at 16 49 55

## Types of changes

## Checklist:

### DIFF
--- a/app/views/assessments/show.html.erb
+++ b/app/views/assessments/show.html.erb
@@ -126,7 +126,7 @@
             <li><%= link_to "View handin history", {:action => 'history'}, {:title=> "View your submissions, scores, and feedback from the course staff" } %> </li>
 
             <% if @assessment.has_writeup? %>
-            <li><%= link_to "View writeup", {:action => 'writeup'}, {:title=> "View the assessment writeup", :class=>""}%> </li>
+              <li><%= link_to "View writeup", {:action => 'writeup'}, {:title=> "View the assessment writeup", :class=>""}%> </li>
             <% end %>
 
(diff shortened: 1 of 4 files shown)


app/views/assessments/show.html.erb
```diff
@@ -124,7 +124,11 @@
         <div class="collapsible-body">
           <ul class="options">
             <li><%= link_to "View handin history", {:action => 'history'}, {:title=> "View your submissions, scores, and feedback from the course staff" } %> </li>
+
+            <% if @assessment.has_writeup? %>
             <li><%= link_to "View writeup", {:action => 'writeup'}, {:title=> "View the assessment writeup", :class=>""}%> </li>
```
20wildmanj, line 129: nit: indentation

20wildmanj (approved):
LGTM

PR #1616, opened 2022-09-28
[hotfix] Use correct files list for sorting annotations when viewing an archive

damianhxy:
## Description
Use result from `Archive.get_files`, rather than `@files`, when calling `get_correct_filename` in `view`.

Hotfix for #1590

## Motivation and Context
Currently, we are using `@files` when finding the filenames for annotations.

The problem with this is that `@files` is the result of `Archive.get_file_hierarchy`, whose length does not correspond directly to the number of files (subfiles are recursively nested). Rather its length is the number of "top level" files / directories.
Screen Shot 2022-09-28 at 17 46 10

This further results in an error when `get_correct_filename` calls `Archive.get_nth_filename`, potentially on an out of bounds index.

However, `viewFeedback` handles this correctly since `@files` is only set for Archives, and is set to the result of `Archive.get_files`.
Screen Shot 2022-09-28 at 17 48 20

## How Has This Been Tested?
* Create an assessment that accepts tar files
* Upload [hello.tar.zip](https://github.com/autolab/Autolab/files/9669116/hello.tar.zip) (unzip this first)
* Create a problem and annotate a nested file

**Before**
Screen Shot 2022-09-28 at 17 49 31

**After**
Screen Shot 2022-09-28 at 17 49 52

## Types of changes

## Checklist:

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -455,13 +455,16 @@ class SubmissionsController < ApplicationController
       @annotations = []
     end
 
+    if Archive.archive? @filename
+      files = Archive.get_files(@submission.handin_file_path)
+    end
     # extract information from annotations
     @annotations.each do |annotation|
       description = annotation.comment
       value = annotation.value || 0
       line = annotation.line
       problem = annotation.problem ? annotation.problem.name : "General"
-      filename = get_correct_filename(annotation, @files, @submission)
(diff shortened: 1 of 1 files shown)


20wildmanj (commented):
- LGTM, Verified that the fix works, can now view an Archive with annotated files
- Should mention this is a hotfix for #1590

20wildmanj (approved):

PR #1621, opened 2022-10-10
Fixed annotation switching problems not updating score

najclark:
## Motivation and Context
Previously, if an annotation was moved from one problem to another, the effect of the annotation's score would remain on the original problem as well as be applied to the new problem. With this fix, when an annotation moves problems, its impact on the score of the original problem is removed and appropriately applied to the new problem.

## How Has This Been Tested?
Set up an assessment with two problems, create an annotation for problem 1 with nonzero score, then update the annotation to apply to problem 2. Confirm that the total scores for problem 1 and problem 2 are correct.

## Types of changes

## Checklist:

### DIFF
--- a/app/controllers/annotations_controller.rb
+++ b/app/controllers/annotations_controller.rb
@@ -34,6 +34,10 @@ class AnnotationsController < ApplicationController
     tweaked_params.delete(:submission_id)
     tweaked_params.delete(:filename)
     ActiveRecord::Base.transaction do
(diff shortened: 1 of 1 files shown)


victorhuangwq (approved):
Tested and it works. 
Personally I think this is a tad bit hacky because it probably should be done within `update_non_autograded_score`  - but the logic behind this is clear so there shouldn't be an issue!

LGTM

PR #1622, opened 2022-10-13
Streaming output frontend

michellexliu:
## Description
Updates the feedback interface to show more details about a submission:
* Shows the score breakdown floating on the side
* Shows the autograder status (right now is hard-coded as "Completed" since we only show feedback when the autograder is finished running, will eventually have states for "Queued"/"In-Progress")
* Only show "Graded by" if it was graded by someone

Problem with autograder score:
image

Problem without autograder score:
image

Scrolled to bottom of page:
image

Side panel with lots of problem categories:
image

## Motivation and Context
* Clean up visuals of interface
* Allow students to see autograder score breakdown in an easier way
* Set up interface to handle newly introduced feedback states for partial output

## How Has This Been Tested?
* Hello lab autograded feedback
* [Twosum lab](https://github.com/autolab/autograders-examples/tree/main/python-twosumlab) autograded feedback
* Broken feedback (last line of feedback is not formatted in the proper `{"scores": {...}}` format) -> displays output, but doesn't have a score breakdown
* Manual feedback
* Long autograder output
* Released remarks --> remarks load at bottom of page
* Unreleased remarks --> remarks section doesn't appear
* Semantic Feedback --> loads semantic feedback page instead of regular feedback page
* Side panel with lots of problem categories

## Types of changes

## Checklist:


damianhxy:
Looks like the changes here would conflict with #1643 -- I modified the `Graded By` to include `Autograder` and `Error Handler`. If you prefer to hide the `Graded By`, I can remove the changes from my PR.

victorhuangwq:
Seeing the bigger "Correctness" at the top really tripped me up.

image

I understand it's leftover from 
image

But since we are already showing all the autograded scores, do we need to repeat that?

victorhuangwq:
Visually, the remarks and all makes sense.

victorhuangwq:
My Remarks does not have a rounded border like the one you have in your screenshot

image

app/assets/javascripts/feedback.js
```diff
@@ -0,0 +1,14 @@
+document.getElementById("pageBody").style.width = '90%';
+$(window).scroll(function(e){ 
+  const $el = $('.result-fixed'); 
+  const container = $('.result-container')[0];
+  const containerWidth = container.offsetWidth;
+  if (container.getBoundingClientRect().top <= 30){ 
+    // set to be 25% of feedback container width instead of entire page
+    // (which is what happens if you use % for a fixed div)
+    // if the .25 is changed, also change the width for .result-summary class
+    $el.css({'position': 'fixed', 'top': '30px', width: `${containerWidth * .25}px`});
```
victorhuangwq, line 10: May I know what is this fixed positioning for? I'm assuming it's the results box

app/controllers/assessments_controller.rb
```diff
@@ -575,7 +575,7 @@ def viewFeedback
       redirect_to(action: "index") && return
     end
     @jsonFeedback = parseFeedback(@score.feedback)
-    @scoreHash = parseScore(@score.feedback) unless @jsonFeedback.nil?
```
victorhuangwq, line 578: may. I know why is the nil check removed?

app/controllers/assessments_controller.rb
```diff
@@ -594,7 +594,7 @@ def parseScore(feedback)
 
     score_hash = JSON.parse(feedback)
     score_hash = score_hash["scores"]
-    if @jsonFeedback.key?("_scores_order") == false
+    if @jsonFeedback && @jsonFeedback.key?("_scores_order") == false
```
victorhuangwq, line 597: NIT: You could perhaps use ruby's safe navigator:
https://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/

app/controllers/assessments_controller.rb
```diff
@@ -868,6 +868,7 @@ def valid_asmt_tar(tar_extract)
     tar_extract.each do |entry|
       pathname = entry.full_name
       next if pathname.start_with? "."
+      next if pathname.start_with? "PaxHeader"
```
victorhuangwq, line 871: what does PaxHeaders stand for?

app/views/assessments/viewFeedback.html.erb
```diff
@@ -1,24 +1,103 @@
...
+            bFilename = @get_correct_filename.call(b)
+            case
+              when aFilename < bFilename
+                -1
+              when aFilename > bFilename
+                  1
+              else
+                  a.line <=> b.line
+              end
+            end
```
victorhuangwq, line 38: @damianhxy was there a helper function for this?

victorhuangwq (approved):
Tested locally and it works. Left some comments, but I will approve it because it's functionally OK.

app/controllers/assessments_controller.rb
damianhxy: Note: if `@jsonFeedback` is `nil` then `@jsonFeedback&.key?("_scores_order")` would be `nil`, which is not the same as `false`

app/views/assessments/viewFeedback.html.erb
damianhxy: Idts, #1590 introduced this code, which only made a helper function `get_correct_filename` and not one for sorting annotations

app/controllers/assessments_controller.rb
victorhuangwq: Currently it also relies on the behavior of `nil` ANDed, which is also not very ideal.

app/assets/javascripts/feedback.js
michellexliu: Yes, it's to keep it floating at the top; Ideally, the width would be set in the CSS, but there's some weirdness going on with how % is calculated with fixed positioning and it causes it to bug out that way

michellexliu:
> Seeing the bigger "Correctness" at the top really tripped me up.
> 
> image
> 
> I understand it's leftover from image
> 
> But since we are already showing all the autograded scores, do we need to repeat that?

hmmm yeah I at one point did consider just removing it, but I decided to leave it in because for non-autograded problems, there's no other way to see the score for that problem on the page. Would you suggest just hiding it if it's autograded and showing it otherwise?

victorhuangwq:
I'm wondering if we could hide it for autograded problems, and show it for non-autograded problems

app/controllers/assessments_controller.rb
victorhuangwq: could you add a comment on it? so that next time people who look at this code would know

michellexliu:
> I'm wondering if we could hide it for autograded problems, and show it for non-autograded problems

Done! @victorhuangwq

PR #1624, opened 2022-10-14
make submission summary fill up space

20wildmanj:
## Description

Previously, assignments with few problems would have a submission summary where the table only took up part of the horizontal space provided:

Screen Shot 2022-10-13 at 22 20 35

The "show all submissions" button also displayed for any amount of submissions, which was weird in the case where only a couple submissions were made, since the information in history is now exactly the same as submission summary:

Screen Shot 2022-10-18 at 18 19 14

## Change Summary

- Changed the table view to now use css that applies `display: table;` to the actual table, and `overflow-x: scroll` to the wrapper div so that the table will become scrollable if there are too many problems, but take up the space otherwise.
- Changed "Download Submission" and "View Source" into buttons so that the options are easier to click
Screen Shot 2022-10-18 at 18 16 48

- Changed the logic behind the submissions history link so that it only displays when the total submission count exceeds the amount of submissions the summary can display. 
  - The summary table is the exact same as the history, so this change doesn't change much, just makes more logical sense.

Screen Shot 2022-10-14 at 11 39 04

Lots of problems:
Screen Shot 2022-10-18 at 18 15 11

## How Has This Been Tested?

- Submit to an assignment with one problem, see that summary takes up horizontal space
- Submit to an assignment with lots of problems, see that summary scrolls properly
- Submit to an assignment with no previous submissions, see that link to "See all x submissions" doesn't appear until after the fourth submission
- Download Submission / View Submission
- Upload plain text file, see that visually the button looks the same, but hovering over shows tooltip that indicates to download as text
## Types of changes

## Checklist:


20wildmanj:
> nit: can we get the icons and button text to be vertically center-aligned? looks slightly off rn

Screen Shot 2022-10-19 at 10 42 18

app/assets/stylesheets/style.css.scss
```diff
@@ -1347,10 +1347,30 @@ a.skip-main:active {
...
+// used in submission_history_table.html.erb
 table.sub {
-  display: block;
-  overflow-x: scroll;
+  display: table;
 }
 .sub th{
   vertical-align: top;
 }
+.submissions {
```
victorhuangwq, line 1357: nit: add spaces between each style

victorhuangwq (approved):
LGTM! Works as expected.

One other small consideration, do we want view source on top of download submissions? I would imagine people would want to view source more often than downloading a submission.

PR #1639, opened 2022-11-10
Add API Endpoints to create problems and query/update scores for an assessment

najclark:
## Description
Created POST /courses/{course_name}/assessments/{assessment_name}/problems 
* Creates a problem
* required parameters: name, max_score

Created GET /courses/{course_name}/assessments/{assessment_name}/scores
* Gets the scores for all users that made submissions

Created GET /courses/{course_name}/assessments/{assessment_name}/scores/{email}
* Gets the scores for the specified user

Created PUT /courses/{course_name}/assessments/{assessment_name}/scores/{email}
* Updates the scores for a specified user
* if update_group_scores is provided as an argument and true, the score update will be reflected along all group members that the user is a part of

## Motivation and Context
Endpoints created by request to allow more robust types of assessments.

## How Has This Been Tested?
**Setup Oauth Access**
Navigate to /oauth/applications and click 'New Application'
1. Use any name
2. Use 'https://oauth.pstmn.io/v1/browser-callback' for redirect url
3. Use 'user_info user_courses user_scores user_submit instructor_all admin_all' for scope

**Setup Postman**
Screen Shot 2022-11-13 at 1 50 54 PM
1. Create new collection
2. Go to authorization tab
3. Set type to 'Oauth 2.0'

Scroll down to 'Configure New Token'
1. Use 'https://oauth.pstmn.io/v1/browser-callback' for callback url
2. Use 'http://localhost:3000/oauth/authorize' as auth URL
3. Use 'http://localhost:3000/oauth/token' as access token URL
4. Use 'user_info user_courses user_scores user_submit instructor_all admin_all' as scope
5. Click 'Get New Access Token'
6. Log into Autolab then return to Postman
7. Click 'Use Token'

Screen Shot 2022-11-13 at 2 20 42 PM

Create new request for each of the following requests  

**POST /courses/{course_name}/assessments/{assessment_name}/problems** 
* Body: {name: problem1, max_score: 100}
* Verify on Autolab that the assessment has problem1 with max_score 100.

**GET /courses/{course_name}/assessments/{assessment_name}/scores**
* Verify that correct scores for each users' submissions is returned

**GET /courses/{course_name}/assessments/{assessment_name}/scores/{email}**
* Verify that correct scores for each user's submissions is returned

**PUT /courses/{course_name}/assessments/{assessment_name}/scores/{email}**
* Add problem1, and problem2 (both with max_score 100) to this assessment.
* Make a group that includes the user defined by `{email}` and add another student
* Accept the group invitation for the other student
* Make a submission as the user defined by `{email}`
* Make a request with body: {problems: {problem 1: 90, problem 2: 10}}
  * Verify that user defined by `{email}`'s score on most recent submission was updated
  * Verify that the other group member's score was not updated
* Make a request with body: {problems: {problem1: 90, problem2: 10}, update_group_scores: true}
  * Verify that all users that are a part of user's group have the scores on their most recent submission updated
* Ensure there is no problem with name `problem3` for this assessment
* Make a request with body : {problems: {problem1: 1, problem2: 1, problem3: 10}}
  * Verify that the user defined by `{email}`'s score on the most recent submission was _not_ updated
  * Verify that the request response is of the form `"error": "'problem3' not found in this assessment"`
* Make a request with body : {problems: {problem1: 1, problem2: 1, problem3: 10}, update_group_scores: true}
  * Verify that all users that are a part of user's group have the scores on their most recent submission _not_ updated
  * Verify that the request response is of the form `"error": "'problem3' not found in this assessment"`

## Types of changes

## Checklist:


victorhuangwq:
Tested: POST /courses/{course_name}/assessments/{assessment_name}/problems
- Seems like you are allowed on autolab to create multiple problems with the same name
- Tested with some invalid cases. Works

Tested GET /courses/{course_name}/assessments/{assessment_name}/scores
- Works as expected

GET /courses/{course_name}/assessments/{assessment_name}/scores/{email}
- Works as expected

PUT /courses/{course_name}/assessments/{assessment_name}/scores/{email}
- It's allowed to exceed max score, is. that intended behavior?
- **Unable enter score unless it has been "graded"**: I submitted a new file. It's 
image
- I call the endpoint with problem1 and problem2
image
and it returns nothing
- Calling the get endpoint shows that 6, the latest assignment has no score
image
- I grade the assignment using the gradesheet
image
- Now I am able to put the score
image

app/controllers/api/v1/scores_controller.rb
```diff
@@ -0,0 +1,120 @@
...
+    scores = {}
+    submissions.each do |submission|
+      raw_submission_scores = Score.where(submission_id: submission.id)
+      scores[submission.version] = problem_name_to_score(raw_submission_scores)
+    end
+
+    respond_with scores
+  end
+
+  def update
```
victorhuangwq, line 44: This is actually a bit misleading of an update endpoint. The `update` endpoint is associated with an id, so in this case, the update should be accepting the score id as a query param and updating that specific score.

I do understand the preferred use case of just updating the latest score like you did however - but in that case do just create a separate endpoint called update_latest perhaps?

app/controllers/api/v1/scores_controller.rb
```diff
@@ -0,0 +1,120 @@
...
+      submission_scores = Score.where(submission_id: submission.id)
+
+      user_scores = {}
+      submission_scores.each do |score|
+        problem_name = problem_id_to_name[score.problem_id]
+        user_scores[problem_name] = score.score
+        next unless params.key?(problem_name)
+
+        score.score = params[problem_name]
+        unless score.save
```
victorhuangwq, line 78: In this case, it's not good to save the scores within this loop. Because if there's an error, the previous score would already have been saved.

For example, you want to update user 1, 2, 3 & 4's problem 1 score to 100.
user1 and user2 saves, but user3 errors out. So now user1 & user2 score has been changed, but user3 hasn't been changed, and this makes the endpoint's "Error case" have side effects. In general, if an error occurs in a batch update, no changes should be made.

victorhuangwq: What I recommend is to keep track of all the scores to be saved in a list, and then use a activerecord transaction to save all of them. If one of it fails, then it rollsback all of them

https://api.rubyonrails.org/classes/ActiveRecord/Transactions/ClassMethods.html

victorhuangwq (changes requested):
Tested API endpoints locally. Left comments on the scenarios that doesn't work

app/controllers/api/v1/scores_controller.rb
najclark: Will change this to `update_latest`

app/controllers/api/v1/scores_controller.rb
najclark: Will change to use `Score.transaction`

najclark:
The API allows for problems with the same name, and for scores that exceeds the max score because this seems to be the existing behavior using the Autolab website.

app/controllers/api/v1/scores_controller.rb
```diff
@@ -0,0 +1,140 @@
...
+        ).order(version: :desc).first
+
+        user_scores = {}
+        problem_id_to_name = @assessment.problem_id_to_name
+        problem_name_to_id = problem_id_to_name.invert
+
+        # Update scores for problems in problem_params
+        problem_params.each do |problem_name, updated_score|
+          unless problem_name_to_id.include? problem_name
+            raise ApiError.new("'#{problem_name}' not found in this assessment", :bad_request)
```
victorhuangwq, line 81: Do you have a test case that tests whether when aborting here, the API correctly aborts the whole transaction?

spec/api/v1/assessments_api_spec.rb
```diff
@@ -24,22 +24,6 @@
     end
   end
 
-  describe 'GET #problems' do
```
victorhuangwq, line 27: why was this test case removed?

victorhuangwq (commented):
Some initial comments

app/controllers/api/v1/scores_controller.rb
najclark: Yeah I will add to the PR description

spec/api/v1/assessments_api_spec.rb
najclark: Was having issues with it, but those are resolved so will add it back to `problems_api_spec.rb`

config/routes.rb
```diff
@@ -33,6 +33,11 @@
           resources :submissions, param: :version, only: [:index] do
             get "feedback"
           end
+
+          resources :scores, only: [:index, :show],
+                    param: :email, :constraints => { :email => /[^\/]+/ }
+
+          put "scores/:email", :constraints => { :email => /[^\/]+/ }, to: "scores#update_latest"
```
victorhuangwq, line 40: I don't think we should repurpose 
scores:/email to update latest, but perhaps create something like scores/:email/update-latest. 

Because if we are putting scores:/email, we should be going with the default update behavior, not update_latest.

victorhuangwq (approved):
Tested and all endpoints works as expected. Good work.
Approved PR as it is functionally correct, but I left a comment on how we are naming the endpoints.

PR #1640, opened 2022-11-10
 Add bulk import users from emails

victorhuangwq:
## Description

Adds ability to create users from bulk using just email.

Adds:
- UI to add emails and assigned role in bulk
- Frontend validation
- Validation, creation of account (if doesn't exist), and assigning of roles

image

## Motivation and Context

Creating accounts individually for adding CAs and Instructors is tedious for class with of them (~40). 

## How Has This Been Tested?

**Tested Valid Inputs:**
- Single entry: dave@autolab.com (should be added/updated with new role), assigned as instructor
- Multiple line of entries: dave1@autolab.com, greg@autolab.com, assigned as course assistant
- Multiple lines of mixed entries: dave@autolab1.com, Greg &lt;greg1@autolab.com&gt;(first name / last name should be updated accordingly), assigned as student
- Mix of pre-existing & new users: Dave &lt;dave@autolab.com&gt;, Greg &lt;greg@autolab.com&gt;, assigned as instructor (We see that we do not update the pre-existing user's name, just the role)

**Tested Invalid Inputs:**
- Single entry: dave, assigned as instructor (error message of invalid email)
- Single entry: dave@test.com, no assigned role (error message of assigned role)
- All wrong entries: dave, test@@.com, greg, dave <test@g (shows which entries are wrong)
- Mix of correct and wrong: dave, dave@test.com, greg (shows which entries are wrong)

## Types of changes

## Checklist:


app/controllers/courses_controller.rb
```diff
@@ -229,6 +229,63 @@ def users
             end
   end
 
+  action_auth_level :add_users_from_emails, :instructor
+  def add_users_from_emails
```
victorhuangwq, line 233: @michellexliu Since you worked with roster upload, before I continue further, could you take a quick look at the logic below?

app/assets/javascripts/course_users.js
```diff
@@ -0,0 +1,101 @@
...
+
+document.addEventListener('DOMContentLoaded', function () {
+  let elem = document.querySelector('#add-users');
+  let instance = M.Dropdown.init(elem, {
+    constrainWidth: false,
+    coverTrigger: false,
+    closeOnClick: false,
+  });
+
+  // Add authennication token to add form
```
damianhxy, line 59: Nit: typo in `authentication`

damianhxy:
Screenshot 2022-11-16 at 00 32 43

For the first invalid input, I get a different error message

app/assets/javascripts/course_users.js
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    let invalid_inputs = [];
+    inputs.split('\n').forEach(function (input) {
+      // keep track of invalid inputs  
+      if (input.length > 0 && !validateIdentity(input)) {
+        error_free = false;
+        invalid_inputs.push(input);
+      }
+      if (invalid_inputs.length > 0) {
```
damianhxy, line 81: nit: this should be outside the `forEach`

app/controllers/courses_controller.rb
```diff
@@ -229,6 +229,109 @@ def users
...
+    params.require([:user_emails, :role])
```
damianhxy, line 234: I don't think this line is necessary, since you're not using the whole `params` object verbatim

app/controllers/courses_controller.rb
```diff
@@ -229,6 +229,109 @@ def users
...
+    # filter out nil emails
+    user_emails = user_emails.reject(&:nil?)
+
+    # check if email matches regex
+    email_regex = /\A[\w+\-.]+@[a-z\d\-]+(\.[a-z\d\-]+)*\.[a-z]+\z/i
+
+    # raise error if any email is invalid and return which emails are invalid
+    invalid_emails = user_emails.reject { |user| user[:email] =~ email_regex }
+    if invalid_emails.any?
+      flash[:error] = "Invalid email(s): #{invalid_emails.map { |user| user[:email] }.join(', ')}"
```
damianhxy, line 266: Is this for an additional layer of safety? Because I believe this error is already caught on the frontend, so a user would never see this.

app/controllers/courses_controller.rb
```diff
@@ -229,6 +229,109 @@ def users
...
   end
 
+  action_auth_level :add_users_from_emails, :instructor
+  def add_users_from_emails
+    params.require([:user_emails, :role])
+    user_emails = params[:user_emails].split(/\n/).map(&:strip)
+
+    user_emails = user_emails.map do |email|
+      if email.nil?
+        nil?
```
damianhxy, line 239: This should be `nil`

app/controllers/courses_controller.rb
```diff
@@ -229,6 +229,109 @@ def users
...
+    if invalid_emails.any?
+      flash[:error] = "Invalid email(s): #{invalid_emails.map { |user| user[:email] }.join(', ')}"
+      redirect_to([:users, @course]) && return
+    end
+
+    role = params[:role]
+
+    @cuds = []
+    user_emails.each do |email|
+      user = User.where(email: email[:email]).first
```
damianhxy, line 274: nit: you can use `find_by(...)` instead of `where( ... ).first`

damianhxy: same for L297 `@course.course_user_data.where(user_id: user.id).first`

damianhxy (changes requested):
Functionality works as expected, but left some comments.

As an additional nit, I realized that I can't press arrow up / down in the text area -- not sure if that was intended.

app/controllers/courses_controller.rb
victorhuangwq: Yup exactly

victorhuangwq: Generally I like to do validation on both sides. You don't want to assume what is calling the endpoint

damianhxy: Yup, but then the testing plan of `Single entry: dave, assigned as instructor (error message of invalid email)` is wrong because the front end catches the error and states `Invalid input: dave` instead

app/assets/javascripts/course_users.js
```diff
@@ -0,0 +1,104 @@
...
+    let invalid_inputs = [];
+    inputs.split('\n').forEach(function (input) {
+      // keep track of invalid inputs  
+      if (input.length > 0 && !validateIdentity(input)) {
+        error_free = false;
+        invalid_inputs.push(input);
+      }
+    });
+    if (invalid_inputs.length > 0) {
+      error_message = "Invalid input: " + invalid_inputs.join(', ');
```
damianhxy, line 86: Should the frontend error message match the backend?

i.e. `"Invalid email(s): #{invalid_emails.map { |user| user[:email] }.join(', ')}"`

damianhxy (approved):
Thanks for addressing the comments

Other than a comment about error message consistency, LGTM!

victorhuangwq:
Tested with:
firstName <firstname@name.com>
firstName lastName <firstlastname@name.com>
firstName M lastName <firstmlastname@name.com>
image

victorhuangwq:
@damianhxy do a last check review on the cases above. Thanks!

damianhxy:
Tested cases as working.

However, it's a bit weird that we're ignoring middle names altogether. Should we lump it with first names?

victorhuangwq:
Currently the roster upload doesn't support middle name as well. So I think it's logical to just drop the middle name to be consistent.

PR #1645, opened 2022-11-16
Autolab partial feedback endpoints

michellexliu:
## Description
Calls the Tango partial output and live jobs endpoints from Autolab to render more detailed job status and display partial output.

### New feedback view states
**In Progress submissions:**
Screen Shot 2022-11-29 at 1 41 43 AM
Users are able to press a refresh button to get updated partial output, which gets displayed in the feedback section. The refresh button is rate-limited from the frontend such that users may only call a refresh every 5 seconds. If there is no longer any partial output, the refresh will trigger a page reload to check for full output.

**Queued submissions:**
Screen Shot 2022-11-29 at 1 41 11 AM
If a submission is queued, a "queued" page will get displayed, as well as the student's queue position. There is similarly a rate-limited refresh button.

**Completed submissions:**
Once autograding is complete, the feedback is displayed as normal, with no refresh button.

### Notable behavior changes
When the autograder is still running, the `--` in the grades table temporarily becomes clickable. If users click into it, they are led to the view feedback page, where the possible views mentioned above are shown.
image

After the autograder finishes running, the auto-graded scores are updated to numbers and remain clickable. Non-auto-graded scores go back to the original behavior. 
image

If a user has clicked into the viewFeedback page of a non-auto-graded problem while the file was still being autograded, and then refreshes on it after autograding is complete, they will be redirected with a "No feedback for requested score" error.

## Motivation and Context
This change allows students to view feedback as soon as it is updated, rather than just at the end. This is especially important for classes with long test cases, where students may not want to wait until their code is completely done to be able to see individual test case results.

### Setup
1. Run Tango (`redis-server`, `python3 restful_tango/server.py`, `python jobManager.py`)
2. Created assignment with [Python autograder](https://github.com/autolab/autograders-examples/tree/main/python-twosumlab)
3. Create Python files that print every second for 30 seconds, 60 seconds, 90 seconds, etc.

### Test scenarios:
* Standard autograder feedback complete view: displayed as "complete", no refresh button, feedback loads
* Standard autograder feedback complete view with an autograder error: displayed as "complete", no refresh button, feedback loads
* Manual feedback: displayed as "complete", no refresh button, feedback loads
* In-progress submission: loads properly, refresh button works
* Queued submission: successfully shows queue position, refresh button works
* In-progress -> in-progress refresh, In-progress -> finished running, in-progress -> finished running with Autograder error, queued -> finished running with semantic config
* Queued -> updated queued position refresh, queued -> in-progress, queued -> finished running, queued -> finished running with Autograder error, queued -> finished running with semantic config
* Output with semantic configuration
* All of the above in both student view & instructor view

## Types of changes

## Checklist:


victorhuangwq:
I haven't yet look at the codebase, but one suggestion I have right now is that we should now change the page that the users see after submitting from the current handin_history page, to this page if there is an autograder setup.

app/views/assessments/_partial_feedback_js.html.erb
```diff
@@ -0,0 +1,65 @@
...
+            location.reload();
+          }
+        },
+    });
+  }
+
+  const btn = $('#refresh-feedback')[0];
+
+  // Throttling Function
+  const throttleFunction=(func, delay)=>{
```
victorhuangwq, line 46: We could use lodash's throttle function. https://lodash.com/docs/3.10.1#throttle 

We have already included lodash into our codebase as a dependency.  Refer to 
/Autolab/app/helpers/application_helper.rb and 
/Autolab/app/views/metrics/index.html.erb

michellexliu:
> I haven't yet look at the codebase, but one suggestion I have right now is that we should now change the page that the users see after submitting from the current handin_history page, to this page if there is an autograder setup.
> 
> Else it's hard for people to know that this feature exist. Because a clickable "--" doesn't signal to the user that they can look at partial feedback upon clicking. Alternatively, you can still redirect them to the handin history page, but put a banner up at the top to tell them that their latest submission is currently being autograded

@victorhuangwq How does this look? I added a link to view feedback in the flash and changed the link to a clock icon instead of `--` if it's grading.
image

app/controllers/assessments_controller.rb
```diff
@@ -587,6 +594,30 @@ def viewFeedback
     }
   end
 
+  action_auth_level :getPartialFeedback, :student
+
+  def getPartialFeedback
```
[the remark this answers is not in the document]
michellexliu: yeah i had tried getting rack-attack earlier by adding `getPartialFeedback` to the `rack-attack.rb` file, but couldn't quite get it to work :( can try to give it another go though

victorhuangwq:
Yes the clock makes a lot more sense, and I think the banner is good too :)

app/controllers/assessments_controller.rb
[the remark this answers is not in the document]
victorhuangwq: okay we can leave it be for now, and create a new issue

victorhuangwq:
I tested using hello.c lab and submitted a random text file

image

Although results are returned, the clock remains. Clicking into it gets me into a loop

victorhuangwq:
Tested using hello.c with a sleep of 2 second and fflush(stdout).
- Clicking refresh works as expected. However
image, clicking on the refresh button does not give me feedback that it has been registered (no UI feedback)
- It doesn't seem to transit from in progress to completed for me automatically, unless I click refresh

app/views/assessments/_partial_feedback_js.html.erb
```diff
@@ -0,0 +1,50 @@
...
+              <p>
+                Your submission is currently in the autograding queue, waiting to be graded.
+              </p>
+              <p>
+                Press the refresh button to get updates.
+              </p>`);
+          } else if (!res.is_assigned) {
+            /* if there's no partial feedback and no queue position,
+                refresh the page to see if full feedback is available */
+            location.reload();
```
victorhuangwq, line 39: I think this is the line that is causing the refresh loop. Basically I think my tango is acting weird (partialfeedback endpoint was not setup properly, so it doesn't work), and thus there is no partial feedback and no queue position.
I don't think what we should do here is to reload again tho

victorhuangwq: I have figured out how to replicate this bug:
- Set up Tango without the PR that introduces partial feedback
- Submit an assignment normally
- Because partial feedback isn't working (the endpoint doesn't exist), so it's confused causing this bootloop

victorhuangwq: Basically, there should be error handling when the partial feedback endpoint throws an error. 
Here's what currently happens when autograding fails (other parts of it)
image

app/controllers/assessments_controller.rb
```diff
@@ -587,6 +604,30 @@ def viewFeedback
...
+      resp['partial_feedback'] = tango_get_partial_feedback(job_id)
+    rescue AutogradeError
+      # if unable to get partial feedback, check if job is on the queue
+      @job_status = get_job_status(job_id)
+      resp["is_assigned"] = @job_status["is_assigned"]
+      resp["queue_position"] = @job_status["queue_position"]
+      resp["queue_length"] = @job_status["queue_length"]
+    end
+
+    render json: resp.to_json
```
victorhuangwq, line 628: So relevant to the other situation with regards to the refresh loop, you should just throw an error instead of return partial feedback

app/views/assessments/viewFeedback.html.erb
```diff
@@ -1,120 +1,40 @@
...
+  <h2>Autograding Result for 
+      <%= @score.nil? ? @assessment.display_name : "#{@assessment.display_name} - #{@score.problem.name} (#{@score.submission.course_user_datum.user.email})"%> 
+  </h2>
+  <div class="result-container">
+    <div class="feedback-container">
+      <div class="result-pane">
+        <% if @score.nil? %>
+          <div class="feedback-header">
+            <h3>Feedback</h3>
+            <div id="status-refresh">
```
victorhuangwq, line 11: Like I mentioned in the other comment, the refresh button should "spin" when you click on it.
This is important because sometimes when a user refresh, but nothing is getting outputted from the course autograder. So they don't know if the click of the refresh is being registered.

For an example of the refresh button behavior, refer to the refresh button on the metrics page http://localhost:3001/courses/AutoPopulated/metrics/index

victorhuangwq (changes requested):
Tested locally. Works well for the ideal use case.

Requested changes for:
- Spinning behavior of refresh on click
- Better error handling in the Autolab's partial feedback endpoint and in the refreshPartial function

app/views/assessments/viewFeedback.html.erb
michellexliu: done

app/controllers/assessments_controller.rb
michellexliu: flash[:error] & redirect were not working... added an error message from frontend to handle this instead

victorhuangwq (approved):
LGTM. 
Tested locally, with both old tango and new tango (with partial output) and works as expected.

damianhxy:
Tangentially related:

With the removal of the `max-width: 970px` in the previous streaming output PR, I observed the following issue where the handin form is too "narrow"

Screenshot 2022-12-02 at 23 20 55

This is because the div has a `col` style applied without any modifier. We should add a `s12` class as follows in `assessments/_handin_form.html.erb`

Screenshot 2022-12-02 at 23 21 18

So that it renders properly

Screenshot 2022-12-02 at 23 22 37

Note: I did a grep and this seems to be the only instance of a `class="col"`

app/views/assessments/viewFeedback.html.erb
damianhxy: For some reason, the refresh icon at the top rotates off-center (at least, sometimes). The one at the bottom is fine though.

https://user-images.githubusercontent.com/9074856/205424604-c0352319-42a2-445f-8a20-6abf706c3aed.mov

app/views/assessments/_partial_feedback_js.html.erb
```diff
@@ -0,0 +1,66 @@
...
+              </div>`);
+            $('.feedback-status').html("In-Progress");
+            $('.feedback-status').removeClass("feedback-status__queued");
+            $('.feedback-status').addClass("feedback-status__inprogress");
+            $( "#result-body" ).load(window.location.href + " #result-body" );
+
+          } else if (res.queue_position != null) {
+            $('#partial-feedback').html(`
+              <div class="autograding-in-progress">
+                <pre>You are at position ${(res.queue_position) + 1} / ${(res.queue_length)} in the autograding queue.</pre>
```
damianhxy, line 23: Nit: Do students need to know / care about the `queue_length`? Perhaps displaying only `queue_position` would be sufficient.

app/views/assessments/_partial_feedback_js.html.erb
```diff
@@ -0,0 +1,66 @@
...
+              $('#status-badge').html("Queued");
+              $('#status-badge').removeClass("feedback-status__inprogress");
+              $('#status-badge').addClass("feedback-status__queued");
+              $('#result-body').html(`
+              <h4>Submission queued...</h4>
+              <p>
+                Your submission is currently in the autograding queue, waiting to be graded.
+              </p>
+              <p>
+                Press the refresh button to get updates.
```
damianhxy, line 34: Nit: `Click` is more semantically correct than `press` for a software button

app/views/assessments/_results_panel.html.erb
```diff
@@ -0,0 +1,56 @@
...
+  <div class="result-pane result-fixed">
+    <div class="feedback-header">
+      <h3>Results</h3>
+    </div>
+    <div id="result-body">
+      <% if !@scoreHash.nil? %>
+        <h4>Autograder Scores</h4>
+        <table class="ui unstackable table">
+          <tbody>
+          <% @scoreHash.each do |key, value| %>
```
damianhxy, line 11: Nit: Could we do something like `@scoreHash.except("_total")` and remove the `key != '_total'` check below

app/views/assessments/_results_panel.html.erb
```diff
@@ -0,0 +1,56 @@
...
+            <% if key != '_total' %>
+              <tr>
+                <td><%= key %>:</td>
+                <td><%= value %></td>
```
damianhxy, line 15: Nit: value should minimally be rounded to 1 d.p. to be consistent with the submission summary display (`value.to_f.round(1)`)

Same with the autograding total below

damianhxy (changes requested):
Followed testing plan, semantic config works too, code generally looks good

Requesting the following changes
- Add `s12` to handin form `div`
- Update logic for `@problemReleased`
- Remove extraneous schema changes
- Address off-center rotation of refresh button

Commented some other nits too

damianhxy (approved):
Other than nits, LGTM

PR #1647, opened 2022-11-21
Fix codebox highlighting after switching files

damianhxy:
## Description
Moved `#highlightLongLines` event listener into `attachEvents` function.

## Motivation and Context
Currently, the event listener for the "highlight lines longer than 80 characters" checkbox is only attached once on page load.

However, when switching files, the elements of the codebox is replaced, meaning that the listener no longer works.

This PR attaches the event listener inside the `attachEvents` function instead, which is called on
- page load
- switching files (cached)
- switching files (uncached)

## How Has This Been Tested?
- Make a submission with a file containing > 80 characters (for example, submit to `hello` and the first line of the autograder output suffices)
- Load autograder output directly, check box functionality (page load)
- Load another file directly, check box functionality (switching files uncached)
- Load autograder output directly, switch away and switch back, check box functionality (switching files cached) 

## Types of changes

## Checklist:


20wildmanj (approved):
Change looks good to me, and verified that highlighting stays active after switching. I was unable to reproduce the original case of highlighting failing however, if possible could I get a more specific flow to break the highlighting?

20wildmanj:
Ah I see, was able to reproduce. All good to merge in

PR #1657, opened 2022-11-28
LTI Roster Integration

najclark:
## Description
Builds on the  [Implement LTI Advantage NRPS GET request PR](https://github.com/autolab/Autolab/pull/1649)
Implements logic to take the NRPS membership data and updates the roster of a course.

Screen Shot 2022-11-28 at 12 03 55 AM
Created new 'lti settings' modal.

Screen Shot 2022-11-28 at 12 04 32 AM
Created new 'sync course roster' page.

## How Has This Been Tested?
Going through the LTI linking instructions outlined in this [PR](https://github.com/autolab/Autolab/pull/1649)

Navigate to 'Manage Course Users'

**Test roster sync**
Screen Shot 2022-11-28 at 12 05 23 AM

 * Click the refresh button
Screen Shot 2022-11-28 at 12 08 12 AM

 * Press confirm
 * Verify that additional students have been added to the course.

**Test dropping missing students**
 * Click 'Linked Course Settings' on the 'Manage Course Users' page
Screen Shot 2022-11-28 at 12 09 52 AM

 * Check 'Auto drop students not enrolled in linked course', then click 'Save'
 * Ensure the course has students enrolled that don't have email *@example.org (i.e. students manually added or imported via a roster)
Screen Shot 2022-11-28 at 12 05 23 AM

 * Click the refresh button
Screen Shot 2022-11-28 at 12 15 04 AM

 * Verify that students missing in the NRPS roster (non @example.org emails) are red
 * Click 'Confirm' and verify that the students that appeared red are now dropped (via the dropped column on the 'Manage Course Users' page)

**Test unlinking a course**
 * Click 'Linked Course Settings'
Screen Shot 2022-11-28 at 12 17 09 AM

 * Click 'unlink' and verify a confirmation window pops up, click 'OK'
Screen Shot 2022-11-28 at 12 17 59 AM

 * Verify the course was unlinked (via success flash, the 'Linked Course Settings' button, and 'Roster last synced with linked course: ...' text should no longer appear on the 'Manage Course Users' page)

## Types of changes

## Checklist:


app/controllers/lti_nrps_controller.rb
```diff
@@ -64,11 +64,13 @@ def request_access_token
...
   # NRPS endpoint for Autolab to send an NRPS request to LTI Advantage Platform
-  action_auth_level :send_nrps_request, :instructor
-  def send_nrps_request
+  action_auth_level :sync_roster, :instructor
+  def sync_roster
+    params.require(:lcd_id)
+
     lcd = LtiCourseDatum.find(params[:lcd_id])
     if lcd.nil? || lcd.membership_url.nil? || lcd.course_id.nil?
-      raise LtiLaunchController::LtiError.new("Unable to update roster", :bad_request)
```
damianhxy, line 71: This seems to be undoing Joey's change, should be reverted (Leave the `LtiLaunchController::` in front)

app/models/lti_course_datum.rb
```diff
@@ -1,3 +1,10 @@
 class LtiCourseDatum < ApplicationRecord
   belongs_to :course
+  attribute :membership_url, :string
```
damianhxy, line 3: Is there a reason why you are defining all the attributes explicitly? I think it's safe to leave them out.

The defaults on `auto_sync` and `drop_missing_students` should probably be defined in the migration instead, e.g.

```
add_column :lti_course_data, :auto_sync, :boolean, default: false
add_column :lti_course_data, :drop_missing_students, :boolean, default: false
```

app/controllers/courses_controller.rb
```diff
@@ -385,6 +394,7 @@ def upload_roster
       begin
         save_uploaded_roster
         flash.now[:success] = "Successfully updated roster!"
```
damianhxy, line 396: Now that you're redirecting, this should be `flash[:success] = ...`

In general, `flash.now` is only when we are directly rendering a page, rather than redirecting.

app/controllers/lti_nrps_controller.rb
```diff
@@ -64,11 +64,13 @@ def request_access_token
   end
 
   # NRPS endpoint for Autolab to send an NRPS request to LTI Advantage Platform
-  action_auth_level :send_nrps_request, :instructor
-  def send_nrps_request
+  action_auth_level :sync_roster, :instructor
+  def sync_roster
+    params.require(:lcd_id)
```
damianhxy, line 69: This seems to be undoing your previous change, should be reverted (remove `params.require(:lcd_id)`)

app/assets/javascripts/course_users.js
```diff
@@ -49,6 +49,10 @@ const validateIdentity = (input) => {
 };
 
 document.addEventListener('DOMContentLoaded', function () {
+  $('#lti-settings-modal').modal({
+    dismissible: false
```
damianhxy, line 53: Perhaps add `preventScrolling: false,`

Currently, the scrollbar disappears when you open the modal, which causes the background to shift. Setting this option preserves the scrollbar (and allows the user to scroll with the modal open, which I think is okay).

Ref: https://materializecss.com/modals.html

app/views/lti_nrps/sync_roster.html.erb
```diff
@@ -0,0 +1,26 @@
+<h2>Sync Course Roster</h2>
```
damianhxy, line 1: This feels like significant duplication with `upload_roster.html.erb`.

I think it would be worthwhile to define a new partial to be shared across both pages (see: https://guides.rubyonrails.org/layouts_and_rendering.html#passing-local-variables)

app/controllers/lti_nrps_controller.rb
```diff
@@ -84,11 +86,70 @@ def send_nrps_request
...
+      email_to_cud[cud.user.email] = cud
+    end
+
+    cud_view = []
+    members_data.each do |user_data|
+      next unless user_data["roles"].include? "http://purl.imsglobal.org/vocab/lis/v2/membership#Learner"
+
+      cud_data = {}
+      # Normalize email
+      user_data["email"] = user_data["email"].downcase
```
damianhxy, line 111: consider `user_data["email"].downcase!` to do it in-place

damianhxy (changes requested):
Functionality-wise, it works!

However, some previous changes appear to have been accidentally undone.

Also left some comments

20wildmanj:
When a dropped student is reactivated, they are no longer dropped, but they don't show up in the confirmation screen.

app/controllers/lti_nrps_controller.rb
```diff
@@ -84,11 +86,70 @@ def send_nrps_request
...
+    cuds = CourseUserDatum.where(course_id: lcd.course_id, instructor: false,
+                                 course_assistant: false).to_set
+    email_to_cud = {}
+    cuds.each do |cud|
+      email_to_cud[cud.user.email] = cud
+    end
+
+    cud_view = []
+    members_data.each do |user_data|
+      next unless user_data["roles"].include? "http://purl.imsglobal.org/vocab/lis/v2/membership#Learner"
```
20wildmanj, line 107: Can you also check if "status":"Active" in addition to them having the learner role? Seems like canvas just removes them from the query so this works but for other platforms it may be different

20wildmanj:
Tested on nightly, found some issues (see above). But for the most part it works pretty well:

Add students:
Screen Shot 2022-11-28 at 11 57 40

Drop students:
Screen Shot 2022-11-28 at 11 23 54

20wildmanj:
Also with the new settings being added please update the docs for LTI Integration as well. Maybe we should also add a page in "Guide for Instructors"?

app/controllers/lti_nrps_controller.rb
najclark: Will do

app/models/lti_course_datum.rb
najclark: Will remove the explicit attribute definitions and define the defaults in the migration.

app/controllers/courses_controller.rb
najclark: Good call, changed

app/controllers/lti_nrps_controller.rb
najclark: Yeah, removed.

app/assets/javascripts/course_users.js
najclark: Oh nice, will add.

app/views/lti_nrps/sync_roster.html.erb
najclark: I agree, created the `courses/_updateRosterPreview.html.erb` partial

app/controllers/lti_nrps_controller.rb
najclark: Will change

app/views/courses/users.html.erb
```diff
@@ -10,22 +10,51 @@
...
+          <span>
+            <p>Auto drop students not enrolled in linked course
+              <i class="material-icons tooltips tooltipped" data-position="top" data-tooltip="When enabled, students not on the <%= lcd.platform %> roster will be marked as dropped.">info_outline</i>
+            </p>
+          </span>
+        </label>
+      </div>
+      <div class="modal-footer">
+        <span style="float: left;">
+          <%= f.submit "Save" , {class: "btn-flat"} %>
```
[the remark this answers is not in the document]
najclark: Synced offline, seems to be fixed. Will test more.

app/controllers/lti_nrps_controller.rb
najclark: Added

app/controllers/lti_nrps_controller.rb
20wildmanj: Screen Shot 2022-11-29 at 23 57 28
this change may have brought on an error

20wildmanj: nvm probably not this change specifically, will investigate

20wildmanj: Screen Shot 2022-11-28 at 17 20 05

So apparently a "Test Student" got added to the canvas course, and the test student doesn't have an email...

20wildmanj: We should add some code to ensure all fields we import into autolab exist

app/controllers/lti_nrps_controller.rb
```diff
@@ -84,11 +84,72 @@ def send_nrps_request
...
+      # Normalize email
+      user_data["email"].downcase!
+
+      user = User.find_by(email: user_data["email"])
+      cud_data[:color] = if user.nil? || lcd.course.course_user_data.find_by(user_id: user.id).nil?
+                           "green"
+                         else
+                           "black"
+                         end
+      cud_data[:email] = user_data["email"]
```
20wildmanj, line 119: btw if this user exists, we should make sure to populate the other fields (e.g. school, major, year, grade_policy, etc.), because right now these fields are blank when a sync occurs, which then overwrites the previous data, when we should preserve the fields we don't touch.

20wildmanj:
Confirmed roster upload unaffected by change to partial.

app/controllers/lti_nrps_controller.rb
damianhxy: The logic from `courses_controller.rb#write_cuds` can probably be reused here

app/views/lti_nrps/sync_roster.html.erb
```diff
@@ -1,26 +1,2 @@
...
-  <%= render partial: 'courses/uploadTable', locals: {cud_view:@sorted_cuds, sort_status:"sorted_cuds"} %>
-  <br>
-  <%= f.submit "Confirm", {:class => "btn primary"} %>
-<% end %>
-
-<% content_for :javascripts do %>
-<%= javascript_include_tag "sort-toggle" %>
-<% end %>
\ No newline at end of file
+<%= render partial: 'courses/updateRosterPreview', locals: { cuds:@cuds, sorted_cuds: @sorted_cuds} %>
```
damianhxy, line 2: Nit: Add space after `cuds:`. Same for `upload_roster.html.erb`

app/views/lti_nrps/sync_roster.html.erb
damianhxy: Verified that roster displays correctly for both importing and syncing

damianhxy:
Previous comments have been addressed, but as Joey mentioned, we should be preserving unchanged fields of existing users

app/controllers/lti_nrps_controller.rb
najclark: Will add a check for this

app/views/courses/users.html.erb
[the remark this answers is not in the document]
najclark: Yeah, I'll remove this

app/controllers/lti_nrps_controller.rb
najclark: I'll make sure to preserve the other fields. The logic from `courses_controller.rb#wrtie_cuds` I think is already being used when the updateRoster confirm button is pressed.

app/views/lti_nrps/sync_roster.html.erb
najclark: Will add the spaces

20wildmanj:
Unrelated to changes, but one thing I have realized is that we have no way of getting the section of a user with LTI integration. I've been looking through Canvas [forum threads](https://community.canvaslms.com/t5/Canvas-Developers-Group/How-can-an-LTI-app-access-course-section-information/td-p/211361) but there doesn't seem to be any good way of getting the sections (and also ofc course #, lecture), which may be important for classes with multiple sections, and grad/undergrad. 

There are ways of getting sections through the [Canvas API](https://canvas.instructure.com/doc/api/sections.html#Section), but that is out of scope.

I guess ultimately to try to get every field the ultimate integration would be with S3, but we never went that route because of how CMU specific it is. Is not being able to get all fields something worth worrying about?

app/controllers/lti_nrps_controller.rb
```diff
@@ -106,44 +122,37 @@ def parse_members_data(lcd, members_data)
...
 
+      cud_view << cud_data
       email_to_cud.delete(cud_data[:email])
     end
 
     return cud_view unless lcd.drop_missing_students
 
     # Mark the remaining students as dropped
-    email_to_cud.each do |email, cud|
+    email_to_cud.each do |_email, cud|
```
damianhxy, line 151: Nit: If `_email` is not needed, it is cleaner to use `_` as the variable name

damianhxy (approved):
Changes look good, verified that existing data is preserved for imported users.

One minor nit but otherwise LGTM!

app/controllers/lti_nrps_controller.rb
```diff
@@ -84,11 +84,86 @@ def send_nrps_request
...
+      user_data["email"].downcase!
+
+      cud_data = {}
+      user = User.find_by(email: user_data["email"])
+      if user.nil? || lcd.course.course_user_data.find_by(user_id: user.id).nil?
+        cud_data[:color] = "green"
+        cud_data[:email] = user_data["email"]
+        cud_data[:first_name] = user_data["given_name"]
+        cud_data[:last_name] = user_data["family_name"]
+        unless user.nil?
```
20wildmanj, line 138: Added in order to preserve an existing user's Major, School, and Year information on a new CUD

app/controllers/lti_nrps_controller.rb
najclark: Will change

PR #1629, opened 2022-10-27
Global annotations

damianhxy:
## Description
- Allow instructors to create, edit, and delete global annotations
- Display all non-autograded problems in annotation pane to facilitate adding global annotations
- Show a dash if a problem is released with no score (via annotation) associated with it. This avoids an internal server error when we try to format `nil`.
- Update annotations documentation (re: shared comments, global annotations, positive grading)
- Fix off-by-one line number on `viewFeedback` page
- Improve validation for max score upon problem creation
- Update logic for `problemReleased` to additionally check that the grading deadline has not passed
- Automatically destroy annotations when a problem is destroyed

## Motivation and Context
To provide instructors with a method of awarding points for a problem globally, without tying it to a specific line of code.

## How Has This Been Tested?
Test using an autograded single file assessment (e.g. hello) and a non-autograded tar assessment
- Ensure that all non-autograded problems are shown in the annotation pane
- Ensure that you can no longer create a problem without a max score
- Ensure that annotation line number shown on viewFeedback page is no longer off-by-one
- Check documentation for annotations
- Test that clickable chevron toggles collapse of a problem
- Successfully create, edit, and delete global annotations
  - Global annotations only appear in the annotation pane, and not in the code viewer
  - Global annotations still influence a student's score correctly
  - Global annotations are not clickable
- Test that above features work even after switching files or adding / editing / deleting annotations
- Ensure that students are unable to view the annotations when either (1) grades are not released and/or (2) the grading deadline has not passed
  - Ensure that students can view annotation otherwise
- Ensure that the view feedback page handles global annotations correctly
  - Clickable link loads the student's submission

**Creating a global annotation**
Screenshot 2022-11-27 at 16 53 52

Screenshot 2022-11-27 at 16 54 45

**Editing a global annotation**
Screenshot 2022-11-27 at 16 54 08

Screenshot 2022-11-27 at 16 54 52

**View feedback page with global annotations** - clicking the link loads the first file of the submission
Screen Shot 2022-11-02 at 17 49 33

**Student unable to view annotations when unreleased**
Screen Shot 2022-11-02 at 17 53 30

Screen Shot 2022-11-02 at 17 54 25

## Types of changes
- [x] New feature (non-breaking change which adds functionality)

## Checklist:
- [x] I have run rubocop for style check. If you haven't, run `overcommit --install && overcommit --sign` to use pre-commit hook for linting
- [x] My change requires a change to the documentation, which is located at [Autolab Docs](https://github.com/autolab/docs)
[...]

PR #1658, opened 2022-11-29
Correctly display orphaned annotations on viewFeedback page

michellexliu:
## Description
Previously, if you add an annotation for a problem, delete the problem, and go view feedback, there was an error caused by calling `problem.name`. 

If the problem is nil, the header is updated to say "Deleted Problem(s)", to match how orphaned problems are handled in #1629.
image

The issue of annotations for deleted problems was fixed by #1629, but this prevents an error on the view feedback page for pages where a problem was deleted before the patch was implemented. 

Also updated check so that instructors & course assistants can view annotations even if they're not released.

## How Has This Been Tested?
* Tested with annotations for regular problems
* Added an annotation for a problem --> deleted the problem --> clicked view feedback

## Types of changes

## Checklist:


damianhxy:
I would suggest making the following simple change to allow orphaned annotations to be deleted:

Add the lines
```ruby
# Associated problem was deleted
return if score.problem_id && score.problem.nil?
```

inside `annotation.rb#update_non_autograded_score`, after `score = Score.find_or_initialize_by_submission_id_and_problem_id...` and before `return if score.grader_id == 0`

We check for `score.problem_id` to be future-proof, in case we implement (problem-independent) global annotations that presumably will have no `problem_id` at all.

damianhxy (approved):
Verified that course assistants and instructors can now see annotations even when the problem is not released.

Also verified that orphaned annotations show up under `Deleted Problem(s)`.

damianhxy:
For consistency, could you also update `getProblemNameWithId` in `annotations.js` as follows:

```js
function getProblemNameWithId(problem_id) {
  var problem_id = parseInt(problem_id, 10);
  var problem = _.findWhere(problems, { "id": problem_id });
  if (problem == undefined) return "Deleted Problem(s)";
  return problem.name;
}
```
**Before**
Screenshot 2022-11-29 at 01 42 51

**After**
Screenshot 2022-11-29 at 01 42 35

damianhxy:
lgtm!

PR #1671, opened 2022-12-14
Update the error message for failed partial feedback & refactor getPartialFeedback logic

michellexliu:
## Description
* Checks that job is assigned _before_ calling `tango_get_partial_feedback` to prevent possible error that occurs when attempting to access partial feedback for job that is being taken off the queue.
* Tracks last valid partial feedback in localStorage so that there's no error message right before an autograding job completes
* Update error message for when no queue position or partial feedback is available to reduce confusion when job is getting assigned.

## How Has This Been Tested?
* Testing scenarios from #1645 
* Repeatedly spam the refresh button during partial feedback to make sure no error message shows up between the final refresh and completion.
* Terminate Tango server.py in the middle of a job to make sure error is still caught correctly

## Types of changes

## Checklist:


victorhuangwq:
**Tested**
- Single positive job queue
- Multiple submissions
- Modified Tango such that it throws an error when getPartialFeedback is called. Was surprised to see an output given. Then I realized it was a cached output from a different submission.
image

app/views/assessments/_partial_feedback_js.html.erb
```diff
@@ -1,22 +1,26 @@
...
+        <i class="refresh-feedback small material-icons" onclick="rotateIcon(); onClickRefresh()">refresh</i>
+        <pre>Autograding in progress</pre>
+      </div>`);
+    $('.feedback-status').html("In-Progress");
+    $('.feedback-status').removeClass("feedback-status__queued");
+    $('.feedback-status').addClass("feedback-status__inprogress");
+    $( "#result-body" ).load(window.location.href + " #result-body" );
+  }
   function refreshPartial(clicked) {
+    const lastPartialFeedback = localStorage.getItem(<%= @job_id %>);
```
victorhuangwq, line 16: job_id does not uniquely identify a job - this counter resets whenever Tango server resets. 
You probably need the submission number (and maybe the user).

victorhuangwq: At the same time, I believe it isn't as appropriate to use localStorage for storing the feedback as the they do not expire (and won't be deleted automatically) So a students localstorage will get bloated after many submissions to autolab.

Either somehow set an expiry, or explore sessionStorage
https://www.digitalocean.com/community/tutorials/js-introduction-localstorage-sessionstorage

victorhuangwq:
A less scary message when partial feedback is blank
image

victorhuangwq:
Maybe we should reword it to There was an _error getting your feedback while autograding_ . Check back once autograding is complete.

Basically we should be conveying that it's only the partial feedback not working, not the whole autograding.

image

app/controllers/assessments_controller.rb
```diff
@@ -608,27 +608,25 @@ def viewFeedback
...
-    resp = {}
+
     # User requested to view feedback on a score
     if job_id.nil?
       flash[:error] = "Invalid job id"
       redirect_to(action: "index") && return
     end
 
-    begin
+    resp = get_job_status(job_id)
```
victorhuangwq, line 618: What happens now if the it throws an error here? Does it bubble up and get returned as an error to the frontend?

app/controllers/assessments_controller.rb
```diff
@@ -608,27 +608,25 @@ def viewFeedback
...
+
+    if resp["is_assigned"]
       resp['partial_feedback'] = tango_get_partial_feedback(job_id)
```
victorhuangwq, line 621: Similarly this?

app/views/assessments/_partial_feedback_js.html.erb
victorhuangwq: job_id only uniquely identifies on the Tango side - basically there will not be two jobs running that has the same job_id. So you won't get a situation where you get someone else's partial feedback

app/controllers/assessments_controller.rb
```diff
@@ -608,27 +608,25 @@ def viewFeedback
...
-      resp["is_assigned"] = @job_status["is_assigned"]
-      resp["queue_position"] = @job_status["queue_position"]
-      resp["queue_length"] = @job_status["queue_length"]
     end
 
     render json: resp.to_json
   end
 
   def parseScore(feedback)
+    return if feedback.nil?
```
victorhuangwq, line 628: May I ask what is this for?

victorhuangwq (changes requested):
Tested and left some comments

app/controllers/assessments_controller.rb
michellexliu: yep it bubbles up— I tested by stopping tango in the middle of autograding so that get_job_status would throw an error and it gets caught

app/controllers/assessments_controller.rb
michellexliu: in the error logs chaskiel sent, it looks like there's some cases where @score.feedback doesn't get properly set even when autograding is complete, which causes an error in this function.

app/views/assessments/_partial_feedback_js.html.erb
```diff
@@ -1,22 +1,33 @@
 <%= external_javascript_include_tag "lodash" %>
 <script>
+  const sessionKeyJson = {
+    submission: <%= @submission_id %>,
+    job: <%= @job_id %>,
+    user: <%= @cud&.user_id %>,
+  }
```
victorhuangwq, line 7: This seems like a good key - issue resolved while testing

app/views/assessments/_partial_feedback_js.html.erb
```diff
@@ -1,22 +1,33 @@
...
+        <i class="refresh-feedback small material-icons" onclick="rotateIcon(); onClickRefresh()">refresh</i>
+        <pre>Autograding in progress</pre>
+      </div>`);
+    $('.feedback-status').html("In-Progress");
+    $('.feedback-status').removeClass("feedback-status__queued");
+    $('.feedback-status').addClass("feedback-status__inprogress");
+    $( "#result-body" ).load(window.location.href + " #result-body" );
+  }
   function refreshPartial(clicked) {
+    const lastPartialFeedback = sessionStorage.getItem(sessionKey);
```
victorhuangwq, line 23: SessionStorage use here resolved the issue of persistence previously.

app/views/assessments/_partial_feedback_js.html.erb
```diff
@@ -1,22 +1,33 @@
...
+    $('#partial-feedback').html(`
+      <pre>${partialFeedback}</pre>
+      <div class="autograding-in-progress">
+        <i class="refresh-feedback small material-icons" onclick="rotateIcon(); onClickRefresh()">refresh</i>
+        <pre>Autograding in progress</pre>
+      </div>`);
+    $('.feedback-status').html("In-Progress");
+    $('.feedback-status').removeClass("feedback-status__queued");
+    $('.feedback-status').addClass("feedback-status__inprogress");
+    $( "#result-body" ).load(window.location.href + " #result-body" );
```
victorhuangwq, line 20: May I ask what's going on in this line of code? I don't quite get what this line does.

victorhuangwq (approved):
Tested - LGTM! 
Left few comments, but PR functional

app/views/assessments/_partial_feedback_js.html.erb
michellexliu: Updates the red "in-progress" badge into a blue "queued" badge 
image

PR #1672, opened 2022-12-22
Catch invalid assessment config on import asmt tar

### LINKED ISSUE #1529: Invalid assessment config SyntaxError not caught on create
**Steps To Reproduce**
1. Create a lab with a .rb config file that has a syntax error
2. Create a tarfile of the lab
3. Click install assessment
4. Click "browse" under "Import from tarball"
5. Select the tarfile for the lab

**Current behavior**
An error page is shown on the deployed version, locally this is shown:
image
The assignment is also created but shows an error whenever you try to access it.

**Expected behavior**
Similar behavior to when you edit an existing config and click reload config (see #1522). No assignment should be created.
image

damianhxy:
## Description
- Validate syntax of assessment config file upon tar import
- Add error handling logic for any `SyntaxError` caught

## Motivation and Context
Currently, if we import the tarfile of an assessment whose config file has a syntax error, the error is not caught. Furthermore, the assessment is created but throws an error every time we attempt to access it.

This PR adds checking to catch syntax errors and also avoids creating assessments if an error is found.

Closes #1529

Notes
- This only catches `SyntaxError`, as it's not really possible to catch errors such as `NameError` and `ArgumentError`
- There is still no checking for import from file system, but this can be the scope of a future PR
- I considered a few ways of displaying the error message
  - We can't just set `@error = e` and use `@error` inside `install_assessment.html.erb`. If we redirect to `install_assessment`, the instance variable disappears. If we render it directly, we miss out on logic set by the controller
  - We could create a separate page just for displaying the error, but then there will be significant duplication with `reload.html.erb`, and it detracts from the workflow
  - We can't use `reload.html.erb` directly either, since it has assessment specific logic (a button to `reload`). It wouldn't really make sense to create a partial just for this either
- Ultimately, I decided to display the error directly inside the flash message. However, this means that I can't show extra info such as the backtrace, as that would cause a cookie overflow error due to its length

## How Has This Been Tested?
- Modify an existing assessment's config file (`<asmt>.rb`) and make it invalid, for example by changing the module's name to be all lowercase
- Export the assessment tar

Before PR
- Create another course and import the assessment tar
- Observe that the assessment is created but we get an error every time we try to access it
- Also observe that the assessment folder is created (`courses/<course>/<asmt>`) 

After PR
- Delete the assessment folder
- Import the assessment tar again
- Observe that the error is caught and displayed to the user, and that the assessment is **not** created
- Also observe that the assessment folder is **not** created

Screenshot 2022-12-21 at 23 31 14

- Finally, import a valid assessment tar and ensure that it still works

## Types of changes

## Checklist:


20wildmanj (approved):
Verified that syntax error catching works and import functionality still works as expected. LGTM

PR #1685, opened 2023-01-12
Change overflow-x on .submissions from scroll to auto

damianhxy:
## Description
As per title.

## Motivation and Context
Currently, `.submissions` uses `overflow-x: scroll`. However, that means a disabled scroll bar displays when there is no overflow. Instead, `overflow-x: auto` should be used, which adds scrollbars only when necessary.

## How Has This Been Tested?
**Before** (disabled scroll bar at the bottom)

Screenshot 2023-01-11 at 22 00 13

**After**

Screenshot 2023-01-11 at 22 00 24

## Types of changes

## Checklist:

### DIFF
--- a/app/assets/stylesheets/style.css.scss
+++ b/app/assets/stylesheets/style.css.scss
@@ -1463,7 +1463,7 @@ table.sub td, th {
 }
 
 .submissions {
-  overflow-x: scroll;
+  overflow-x: auto;
 }
 
 .submission-history-button {


20wildmanj (approved):
LGTM. I'm not sure how the behavior works but on my end making it `scroll` or `auto` doesn't really change anything. This is how my submissions look with `scroll` for example:

Screen Shot 2023-01-15 at 10 51 29

If changing it to `auto` makes the scrollbar disappear properly in some cases though then it's worth to change.

damianhxy:
At least for me, using `scroll` results in the scrollbar appearing on both Firefox and Chrome. Changing it to `auto` properly hides it so I'll merge this PR

PR #1689, opened 2023-01-18
Input pre-validation on sign-up page

evanyeyeye:
## Description
- Removed centralized error reporting from top of page. 
- Added hint text below each field. 
- Added HTML pre-validation for each field. The only error message that should be possible without the user intentionally disabling checks is for reuse of an existing email address.
- All errors now appear under their corresponding field. 

## Motivation and Context
It is easy for users to make errors when trying to sign up because of the lack of instruction. In addition. the default error printing is not entirely intuitive and also results in huge (ugly) font-sizes. 

## How Has This Been Tested?
Tried signing up with various valid and invalid input combinations.

## Before
chrome_S5qDXnktyC

## After 
image
image
image

## Types of changes

## Checklist:


app/assets/stylesheets/signup_form.css
```diff
@@ -0,0 +1,5 @@
+.field-hint {
```
damianhxy, line 1: This style isn't necessary, we should reuse the `.help-block` style from `style.css.scss` to be consistent with other forms

app/views/devise/registrations/new.html.erb
```diff
@@ -1,31 +1,72 @@
...
+<% end %>
 
+<% content_for :javascripts do %>
+  <%= javascript_include_tag "signup_form" %>
+<% end %>
+
+<%= form_for(resource, :as => resource_name, :url => registration_path(resource_name)) do |f| %>
 	<div class="input-field">
-	  <%= f.text_field :first_name, :autofocus => true %>
+		<%= f.text_field :first_name, :title => "", :pattern => ".+", :required => true, :autofocus => true %>
```
damianhxy, line 13: If you are setting `:title => ""` to suppress the hover message, then you should be consistent with it for the other fields

damianhxy: Also, what is the purpose of `:pattern => ".+"`? If `required` is `true`, I think that would be implied, and if `required` is `false`, then it serves no purpose?

Likewise for `last_name`

app/views/devise/registrations/new.html.erb
```diff
@@ -1,31 +1,72 @@
...
 		<%= f.label :first_name %>
+    <div class="field-hint">
```
damianhxy, line 15: There seems to be mixing of tabs and spaces in this file. I know that the file was already like that, but could you switch indentations to spaces?

damianhxy (changes requested):
Tested that functionality works, great javascript code!

Left some comments

damianhxy (approved):
LGTM

PR #1693, opened 2023-01-20
Centralize Sass variables into single file

20wildmanj:
Move Sass variables into `_variables.scss` in order to centralize variables into a file that can be imported into other scss / css files, without having duplicating css in production from importing. Duplicating css code was introduced after #1649 and subsequently #1657. Other features also duplicated the variables in separate CSS files.

Solution derived from the following [post](https://stackoverflow.com/a/20924703).

Note:  The recommended option suggested to import variables in Sass is to use the `@use '<file>'` syntax, but `sass-rails` which is now deprecated doesn't support this functionality. We should probably move to [dart-sass](https://github.com/rails/dartsass-rails) which is still actively being supported.

## Description
- Create `_variables.scss` to hold application-wide sass (don't use `application.scss` as it imports a ton of CSS from materialize and bootstrap)
- Remove `import style.css.scss` from `usersTable.css.scss`
- replace duplicated variables from github integration, metrics stylesheets with imported variable

## How Has This Been Tested?

- run autolab on a production environment (autolab-docker for example)
- check public/assets/usersTable-<long string of chars>.css, see that it only contains css from the original file and doesn't have extra style duplicated
- can also run `RAILS_ENV=development bundle exec rake assets:precompile` on local development server and check the above

## Types of changes

## Checklist:


app/assets/stylesheets/_variables.scss
```diff
@@ -0,0 +1,10 @@
+/********************************************************
+* Autolab Specific CSS variables *
+********************************************************/
+$autolab-red: rgba(153, 0, 0, 0.9);
+$autolab-subtle-red: #c87b7e;
+$autolab-subtle-gray: #eee;
+$autolab-highlight-gray: #ccc;
+$autolab-black-text: #212121;
+$autolab-blue-text: #0882af;
```
damianhxy, line 9: Nit: this style is used in a couple of places, could we replace the occurences with the variable?
- `instructor_gradebook.css:109`
- `annotations.css:515`
- `style.css.scss:1467`
- `gradesheet.css.scss:310`

app/assets/stylesheets/_variables.scss
```diff
@@ -0,0 +1,10 @@
+/********************************************************
+* Autolab Specific CSS variables *
+********************************************************/
+$autolab-red: rgba(153, 0, 0, 0.9);
```
damianhxy, line 4: Nit: style in `quiz.css.scss:73`, we should replace it with the variable

damianhxy (approved):
Tested with precompile method, LGTM

app/assets/stylesheets/beta.scss
```diff
@@ -195,13 +197,13 @@ td.HiddenGradeCell {
 }
 
 table.announcements {
-  border: 1px solid #cccccc;
+  border: 1px solid ;
```
damianhxy, line 200: I think you meant to put `$autolab-highlight-gray` here

damianhxy (approved):
Other than the one missing style, LGTM

PR #1698, opened 2023-01-26
Set up Capybara + Selenium testing, create basic submission tests

20wildmanj:
- Set up Selenium so Autolab can run browser-based automated tests
- Add rake task for simpler test flow
- Create simple tests for submission feature
  
## Description

- Set up Capybara to use Selenium as well as the `webdrivers` gem in order for it to be used by Capybara for browser-based automated testing. `launchy` is also being added to facilitate saving and displaying HTML pages when using capybara with `save_and_open_page`.
- Add `setup_test_env` rake task in order to simplify test environment setup. Previously all four commands would need to be inputted one after the other to set up the testing environment
- Add basic render tests based on user type for `SubmissionsController` (these will likely be expanded at a later date, but would rather not do now, as the testing framework changes should be pushed out quickly)
- Add controller macros
  - `create_ca_for_course`
  -  `get_first_course`
  -  `get_first_cud_by_uid`
  -  `get_first_aid_by_cud`
- Add integration test using Selenium under `manage_submissions_spec.rb`
  
## Motivation and Context

As part of push to revitalize automated testing on Autolab.

## How Has This Been Tested?

To run the individual tests (assuming in `/Autolab`):

```
RAILS_ENV=test bundle exec rails autolab:setup_test_env
rake spec SPEC=./spec/features/manage_submissions_spec.rb
```
```
RAILS_ENV=test bundle exec rails autolab:setup_test_env
rake spec SPEC=./spec/controllers/submissions_controller_spec.rb
```

To run all tests:

```
RAILS_ENV=test bundle exec rails autolab:setup_test_env
bundle exec rails spec
```

## Types of changes

## Checklist:

( I can update the documentation to include the new rake task if desired)

### DIFF
--- a/spec/controllers/submissions_controller_spec.rb
+++ b/spec/controllers/submissions_controller_spec.rb
@@ -51,7 +51,8 @@ RSpec.describe SubmissionsController, type: :controller do
 
     context "when user is course assistant" do
       cid = get_first_course
-      user_id = create_ca_for_course(cid, "courseassistant@test.com", "course", "assistant", "12345678")
+      user_id = create_ca_for_course(cid, "courseassistant@test.com", "course", "assistant",
+                                     "12345678")
       cud = get_first_cud_by_uid(user_id)
       cid = get_course_id_by_uid(user_id)
       course_name = Course.find(cid).name
--- a/spec/controllers/submissions_controller_spec.rb
+++ b/spec/controllers/submissions_controller_spec.rb
@@ -13,26 +13,54 @@ RSpec.describe SubmissionsController, type: :controller do
       assessment_id = get_first_aid_by_cud(cud)
       assessment_name = Assessment.find(assessment_id).name
       it "renders successfully" do
-        get :index, params: {course_name: course_name, assessment_name: assessment_name}
+        get :index, params: { course_name: course_name, assessment_name: assessment_name }
         expect(response).to be_successful
         expect(response.body).to match(/Manage Submissions/m)
       end
     end
 
-    context "when user is Autolab normal user" do
-      login_user
+    context "when user is Instructor" do
+      user_id = get_instructor
+      login_as(user_id)
+      cud = get_first_cud_by_uid(user_id)
+      cid = get_course_id_by_uid(user_id)
+      course_name = Course.find(cid).name
+      assessment_id = get_first_aid_by_cud(cud)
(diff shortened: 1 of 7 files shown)


damianhxy:
Could you update the `Set up database` workflow (inside `rubyonrails.yml`) to use the new initialization method?

spec/features/manage_submissions_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+        click_on "Homework 0"
+
+        click_on "Manage submissions"
+        first(:link, "Edit the grading properties of this submission").click
+        fill_in("submission_notes", with: "test notes")
+        fill_in("submission_tweak_attributes_value", with: "1.0")
+
+        click_on("Update Submission")
+
+        # TODO: check values are okay
```
damianhxy, line 48: Update comment once #1740 merged

spec/support/controller_macros.rb
```diff
@@ -50,6 +50,40 @@ def get_course_id_by_uid(uid)
     CourseUserDatum.where(user_id: uid).first.course_id
   end
 
+  def get_first_cud_by_uid(uid)
+    CourseUserDatum.where(user_id: uid).first
```
damianhxy, line 54: Semantically, this should return `... .first.id` (although it seems to work even without the id)

spec/controllers/submissions_controller_spec.rb
```diff
@@ -1,4 +1,68 @@
...
+      it "renders with failure" do
+        get :index, params: { course_name: course_name, assessment_name: assessment_name }
+        expect(response).not_to be_successful
+        expect(response.body).not_to match(/Manage Submissions/m)
+      end
+    end
+
+    context "when user is course assistant" do
+      cid = get_first_course
+      user_id = create_ca_for_course(cid, "courseassistant@test.com", "course", "assistant",
```
damianhxy, line 54: Perhaps we can update `autolab.rake` to generate course assistants too?

damianhxy (approved):
Tests work, LGTM

(10/10 would break my Autolab install again)

spec/controllers/submissions_controller_spec.rb
20wildmanj: That would be a good idea, I'll leave it to a future PR though since we may wanna update the autopopulated DB in general

Issue #1665, opened 2022-12-02
Annotations Frontend Issues


20wildmanj:
**Your environment**
v 2.9.0

## Summary of issues

This issue compiles multiple issues with the frontend of the Annotations feature, found by @20wildmanj and @damianhxy.

## 1. Annotation closing unexpectedly

Clicking "update annotation" on an existing annotation when another annotation is also open removes the other open annotation

![image](https://user-images.githubusercontent.com/25730111/205388218-a1473ce0-06ca-43ca-8f71-fd94f7ee67cd.png)

### Context

Known issue, it's because of how `annotations.js` liberally calls `displayAnnotations()` after some events (change file,  cancel, update), which basically rerenders all the annotation boxes and thus closes the existing ones.

### Possible Solution

We need a more targeted approach of only editing the required annotation.

## 2. Duplicating Autofilled comments

When autofilling, you can add the same comment that got autofilled back to the shared comment pool, and it'll become a new shared comment.

### Context
Given the current design of shared comments (`shared_comment` flag on annotations), there isn't really a way around this atm.

### Possible Solution
Have a table of shared annotations where each annotation points to the shared annotation it is utilizing and perhaps the shared annotation can store a refcount. This could be a little complicated though.

## 3. Multi-line comments

Multi-line comments don't seem to open all the way when you re-open / autofill

![image](https://user-images.githubusercontent.com/25730111/205389715-33a6239d-fc9d-4e84-aaa7-b5a6b77119b8.png)

### Context

Known issue, some previous observations:
![Screenshot 2022-12-02 at 15 52 44](https://user-images.githubusercontent.com/9074856/205384317-ae4d9695-2870-4f30-8d2b-2804f04c5303.png)

## 4. Shared comments drop-down

adding newlines to an annotation makes the shared comments drop-down get misaligned

![image](https://user-images.githubusercontent.com/25730111/205389964-835aae8b-cbff-4171-a399-2cda5daf9b99.png)

## 5. Scrolling to an Annotation from Sidebar

When we click on an annotation on the right side bar, we start scrolling from the top of the code viewer each time, even if the annotation is already in view. This can result in janky animation.

## 6. Autofilling score from shared comment

**Feedback from 122:** Point values should stay with a shared comment added to the pool. In the screenshot below, "A1: Has contract”’s point values when I created it, +1.0, should be automatically applied when I choose this comment from the shared pool. This issue comes up not only when I attach it to another Problem, but within the same Problem as well.

### Possible solution

[...]

PR #1700, opened 2023-01-26
Scrolling Annotation UI Bug

lykimchee:
## Description
Changed Autolab/app/views/submissions/_annotation_pane.html.erb to fix the scroll animation jumping around when clicking on an annotation that was already in view. (Issue #1665 No. 5)

## Motivation and Context
Previously, when we clicked on an annotation on the right side bar, we start scrolling from the top of the code viewer each time, even if the annotation was already in view, which resulted in janky animations.

## How Has This Been Tested?
Create a multiline file submission, then write an annotation for any line (preferably on a line not initially visible when the file is loaded, such that you have to scroll to view the line).
Test by clicking annotations on the right side bar when:
- The line is not in view (is below your current position on the file)
- The line is not in view (is above your current position on the file)
- The line is in view
- Immediately after clicking the annotation (so the annotation should've already been scrolled to)
The expected behavior would be a gradual scroll such that the annotation + associated line in the submission file moves to the top of the code pane, unless it was already at that position.

## Types of changes

## Checklist:


michellexliu (approved):
auto-scroll works for all conditions, lgtm!

damianhxy (changes requested):
There are at least two immediate issues:

1. Style of annotations have changed
**Before**
Screenshot 2023-01-29 at 03 45 10

**After**
Screenshot 2023-01-29 at 03 44 57

2. Clicking on an annotation from a different file no longer brings you to that file. Instead, it attempts to scroll on the current file.

damianhxy:
To properly address the issue, I would suggest taking a look at `submissions/view.js.erb`, which is the code that is executed when you click on an annotation (via the `remote: true` flag).

You'll see that it calls a function `scrollToLine`, which is defined in `app/assets/javascripts/annotations.js` and uses the `scrollTo` api (https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTo).

The problem with `scrollTo` is that it doesn't check if the element is already in view.

.gitignore
```diff
@@ -57,3 +57,5 @@ out.txt
 # Rails credentials
 config/master.key
 config/credentials.yml.enc
+
+dump.rdb
```
damianhxy, line 61: Instead of adding `dump.rdb` to `.gitignore`, it would be best for you to simply delete the file / not commit it. Redis should not be outputting anything to the Autolab directory.

app/views/submissions/_annotation_pane.html.erb
```diff
@@ -106,4 +104,4 @@
       <% end %>
     </ul>
   </div>
-</div>
+</div>
```
20wildmanj, line 107: Nit: add a newline :((

PR #1750, opened 2023-01-31
Backend throttling for getPartialFeedback

### LINKED ISSUE #1661: Throttle getPartialFeedback endpoint from backend
**Current behavior**
Currently, the refresh button for partial feedback is only throttled from the frontend. This should be fine for the most part, but there is a slight risk that a student may find some way around this and abuse the endpoint from the backend. 

**Expected behavior**
There should be throttling for the getPartialFeedback endpoint via rack-attack (see `config/initializers/rack-attack.rb`). Have attempted adding
```
throttle("getPartialFeedback", :limit => 1, :period => 5.seconds) do |req|
    if req.path.include?("getPartialFeedback")
      req.ip
    end
  end
```
However, it seems to not be working... will need to find a way to fix this.

michellexliu:
Resolves #1661 

## Description
Updates `rack-attack.rb` to throttle `getPartialFeedback` from the backend & fixes error that occurs when throttling all other endpoints.

## Motivation and Context
Previously, `getPartialFeedback` was only throttled via the frontend. Students would be able to get past this frontend throttling by writing a script, or by doing an entire webpage refresh rather than an interface refresh. If abused, this could overload Tango with too many requests, thus necessitating backend throttling.

Additionally, previously, if `env['rack.attack.match_data']` returned `nil`, an error was raised in the overloaded throttled_response method. This error would prevent the endpoints from actually being throttled.

## How Has This Been Tested?
Note: must first activate cache on the development build using `rails dev:cache`. This can be confirmed by checking for the existence of a `tmp/caching-dev.txt` file.
- Calls getPartialFeedback normally after waiting 5 seconds
- 429 error is returned after doing successive page refreshes
- 429 error is returned properly when frontend throttling is disabled and refresh button is spammed
   To disable frontend throttling, replace `onClickRefresh` with `const onClickRefresh = () => refreshPartial(true);` in `app/views/assessments/_partial_feedback_js.html.erb`

## Types of changes

## Checklist:


victorhuangwq:
I tested the following:
> 429 error is returned properly when frontend throttling is disabled and refresh button is spammed
To disable frontend throttling, replace `onClickRefresh` with const `onClickRefresh = () => refreshPartial(true)`; in app/views/assessments/_partial_feedback_js.html.erb

Specifically this is a code snippet of what I have changed:
```

  refreshPartial(false);

  const onClickRefresh = () => refreshPartial(true);

  const rotateIcon = () => {
    const icon = $('.refresh-feedback');
    if(!icon.hasClass("loading")) {
	    icon.addClass('loading');
      setTimeout(() => icon.removeClass('loading'), 1000);
    }
  }

  $('.refresh-feedback')[0].addEventListener("click", onClickRefresh);
  $('.refresh-feedback')[0].addEventListener("click", rotateIcon);
  ```
  
  However, I don't think I see any throttling, or 429 errors popping up for me?
  
image

victorhuangwq (approved):
LGTM.

Tested by activating cache on the development build using `rails dev:cache`

Deactivated frontend throttling, and backend throttling kicks in. 5 seconds does feels a bit longer than expected between wait - but I guess the front end throttling is also 5 seconds too?

image

PR #1759, opened 2023-02-02
Update Dockerfile to use passenger-ruby27

20wildmanj:
Update dockerfile to use version of passenger that supports Ruby 2.7.7

## Motivation and Context

Necessary for Autolab docker container to build due to update to Ruby 2.7.7 in #1698.

## How Has This Been Tested?

phusion/passenger-ruby27 works on Nightly.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@
 #     https://github.com/phusion/passenger-docker
 #
 #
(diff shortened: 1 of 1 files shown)


damianhxy (commented):
LGTM

damianhxy (approved):

Issue #1705, opened 2023-01-29
Multi-line comments don't open all the way


michellexliu:
Multi-line comments don't seem to open all the way when you re-open / autofill

![image](https://user-images.githubusercontent.com/25730111/205389715-33a6239d-fc9d-4e84-aaa7-b5a6b77119b8.png)

### Context

Known issue, some previous observations:
![Screenshot 2022-12-02 at 15 52 44](https://user-images.githubusercontent.com/9074856/205384317-ae4d9695-2870-4f30-8d2b-2804f04c5303.png)

PR #1660, opened 2022-12-01
Annotations improvements and bugfixes

damianhxy:
## Description
**Bug fixes**
- Prevent selection of autograded problems when editing annotation
- Persist shared comments checkbox for global annotations
- Preserve newlines in descriptions
- Add handling for `nil` problem `max_score`

**New features**
- Display current max score on edit problem page
- Add problem max score to annotation panel
- Allow multiple annotations per line

**Breaking changes**
- Make shared comments operate on a per-problem basis

**Others**
- Optimize performance of updating grades and annotation panes

## Motivation and Context
Addresses the following feedback from 122:

1. Keep newlines in the global annotations when displayed. In the screenshot below, I input A, B, C, D on separately lines, but to students, it’ll be all on one line, which is undesirable
2. Make comments added to the shared pool specific to a problem (or not, should be user defined). I currently see that if a comment is added to one problem, it’ll show up in the shared pool for other problems too. This might introduce a lot of clutter if we want problem-specific shared comments.
3. Add the max number of points for each problem to the annotations panel
4. When I hover a line that already has a comment, the green + button appears, making me assume that I can add another comment, but clicking the button doesn’t do anything. It’d be nice to have multiple comments per line. This is depicted in the screenshot above as well.

**Note**
Feedback included the following, to be addressed in a future PR:

Point values should stay with a shared comment added to the pool. In the screenshot below, "A1: Has contract”’s point values when I created it, +1.0, should be automatically applied when I choose this comment from the shared pool. This issue comes up not only when I attach it to another Problem, but within the same Problem as well.

## How Has This Been Tested?

**1. Prevent selection of autograded problems when editing annotation**
- On an autograded assessment, make an annotation (ensure that we can't select the autograded problem)
- [**previously**] Edit the annotation, we are able to select the autograded problem
- [**now**] Edit the annotation, we are **not** able to select the autograded problem

**2. Persist shared comments checkbox for global annotations**
- Make a global annotation, mark it as a shared comment
- [**previously**] Edit the annotation, the shared comment checkbox is not checked
- [**now**] Edit the annotation, the shared comment checkbox **is** checked

**3. Preserve newlines in descriptions**
- Create a multi-line global annotation, check that newlines are preserved in the annotation pane
- Create a multi-line normal annotation, check that newlines are preserved in the annotation pane
- Also observe that annotations descriptions start to the immediate right of the point badge as usual

Screenshot 2022-12-02 at 00 38 18

**4. Display current max score on edit problem page**
[...]

PR #1760, opened 2023-02-03
Autofill scores for shared comments

### LINKED ISSUE #1708: Autofilling score from shared comment
**Feedback from 122:** Point values should stay with a shared comment added to the pool. In the screenshot below, "A1: Has contract”’s point values when I created it, +1.0, should be automatically applied when I choose this comment from the shared pool. This issue comes up not only when I attach it to another Problem, but within the same Problem as well.

### Possible solution

It would necessitate the writing of custom code (see https://jqueryui.com/autocomplete/#custom-data) to attach extra attributes to each dropdown element (i.e. the score) and to fill in the score fields when we select a value.

michellexliu:
## Description
* Upon selecting a shared comment, the score gets autofilled along with the comment.
* Auto-resizes the textarea when you click into it or select a shared comment

Resolves #1708, #1705

## Motivation and Context
Classes, such as 15-122, tend to pre-populate a rubric bank of responses, so that their TA's can easily select common responses. By filling in the scores along with the comments, there's one less step of TA's having to type in the scores themselves.

## How Has This Been Tested?
* Create shared annotation with a non-default score --> check that the shared annotation displays properly in the autocomplete dropdown and fills comment/score correctly
* Create shared annotation with a non-default score in global annotations --> check that the shared annotation displays properly in the autocomplete dropdown and fills comment/score correctly
* Create shared annotation with a non-default score --> check that the shared annotation displays properly in the autocomplete dropdown for global annotations and fills comment/score correctly
* Create shared annotation with a non-default score in global annotations --> check that the shared annotation displays properly in the autocomplete dropdown for global annotations and fills comment/score correctly
* Works across different submissions
* Works after editing score
* Works with shared comments created before this PR
* Autoresizes when long comment is selected
* Autoresizes when long comment is selected in global annotation

## Types of changes

## Checklist:

## Future Steps
* 15-122 also requested for the problem category to be auto-filled, but right now the shared comments show up per problem anyways. We could change this such that the shared comment pool is per assignment & then auto-fill problem, but we might need to discuss if this is a good idea.
* Could find some way to display score on autocomplete label


damianhxy:
Screenshot 2023-02-04 at 22 39 49

Doesn't automatically resize for a long annotation when you open-close-open an annotation.

(There is probably something funky going on during the redrawing process that breaks something)

damianhxy:
Regarding the shared comments being per-problem, that was a change I implemented in #1660 due to 122 feedback:

> Make comments added to the shared pool specific to a problem (or not, should be user defined). I currently see that if a comment is added to one problem, it’ll show up in the shared pool for other problems too. This might introduce a lot of clutter if we want problem-specific shared comments.

So I think the current way is fine

damianhxy:
Screenshot 2023-02-04 at 22 45 09

When you open - close - open an annotation (basically forcing a redraw), the box doesn't resize until you click on it.

damianhxy:
When a long annotation text is deleted while editing, the box does not resize until we click out and back in


https://user-images.githubusercontent.com/9074856/216800474-7d4fba18-1e67-4f1a-8c01-5bddaf69e2ba.mov

app/assets/javascripts/annotations.js
```diff
@@ -745,7 +782,11 @@ function newAnnotationBox(annotation) {
     const problem_id = $(this).val();
 
     // Update autocomplete to display shared comments for selected problem
-    box.find("#comment-textarea").autocomplete("option", "source", localCache["shared_comments"][problem_id] || []);
+    box.find("#comment-textarea").autocomplete({
+        source: getSharedCommentsForProblem(problem_id) || [],
```
damianhxy, line 786: Nit: Could we use a consistent ordering for the source, select, and focus options?

app/assets/javascripts/annotations.js
```diff
@@ -745,7 +782,11 @@ function newAnnotationBox(annotation) {
     const problem_id = $(this).val();
 
     // Update autocomplete to display shared comments for selected problem
-    box.find("#comment-textarea").autocomplete("option", "source", localCache["shared_comments"][problem_id] || []);
+    box.find("#comment-textarea").autocomplete({
```
damianhxy, line 785: Note that it is unnecessary to specify select and focus here. This code is triggered when we change the problem, so we only need to update the source. Same for L526.

damianhxy (approved):
Verified that functionality works.

Left some nits, and also feedback, but approving first as they are not major issues

app/assets/javascripts/annotations.js
```diff
@@ -727,13 +759,18 @@ function newAnnotationBox(annotation) {
     e.preventDefault();
     box.find('.annotation-box').hide();
     box.find('.annotation-form').show().css('width', '100%');
+    
+    M.textareaAutoResize(box.find('#comment-textarea'));
```
damianhxy, line 763: Could you do the same thing for global annotations? 

Screenshot 2023-02-05 at 14 16 39

PR #1490, opened 2022-03-29
Force download of attachments

fanpu:
## Description

Cause attachments to be downloaded instead of being opened with the default MIME handler.

I had to also fix the rubocop-rails dependency in the Gemfile.lock as the previous config broke rubocop for me.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] I have run rubocop for style check. If you haven't, run `overcommit --install && overcommit --sign` to use pre-commit hook for linting
- [ ] My change requires a change to the documentation, which is located at [Autolab Docs](https://github.com/autolab/docs)
- [ ] I have updated the documentation accordingly, included in this PR

## Other issues / help required

If unsure, feel free to submit first and we'll help you along.

PR #1810, opened 2023-02-12
[Breaking] Use ActiveStorage for attachments, update attachments tests

### LINKED ISSUE #617: attachments should be stored in the course directory
There seems to be no good reason for attachments to live in a single toplevel directory. Why aren't they in the course or assessment directory?

damianhxy:
## Description
- Use ActiveStorage to store and serve attachment files
- Update existing tests

**This is a breaking change, and old attachments will have to be re-uploaded.**

## Motivation and Context
Currently, all attachments are stored in a single top-level directory, where collisions are resolved by appending numbers to the end of the filename. This PR switches to ActiveStorage, which handles the file storage logic for us.

Using ActiveStorage also makes it easy to switch to other services such as S3 in the future.

Resolves #617 (albeit with a different approach)

More info about ActiveStorage: https://guides.rubyonrails.org/active_storage_overview.html

Note: Undoes the changes made in #1490 (in consultation with the author) in preference for a better solution to the XSS issue. 
In that PR, the mime_type was set to `application/octet-stream` to force downloads. A better solution is to remove `disposition: inline` since the default is `disposition: attachment`, which forces downloads.

More details here: https://huntr.dev/bounties/90701766-bfed-409e-b3dd-6ff884373968/

## How Has This Been Tested?
ActiveStorage

- Either clear attachments folder or remember what it looks like (or delete your attachments folder)
- Create an attachment
  - Observe that `/attachments` remains unchanged
  - Observe that attachment is stored in `/storage` (might be several directories down, with a randomly generated file name)
- Download the attachment
- Edit and download the attachment
- Delete attachment, observe that file is deleted from `/storage`
- If you want, you can test that the released / unreleased options work, but this PR doesn't touch that logic

XSS fix
- Upload example file from huntr.dev report: something
- When attachment is clicked, ensure file is downloaded rather than displayed

Tests
- Run `rake spec SPEC=./spec/controllers/attachments_controller_spec.rb` and ensure they pass
- Also run the full test suite

## Types of changes

## Checklist:

### DIFF
--- a/app/controllers/attachments_controller.rb
+++ b/app/controllers/attachments_controller.rb
@@ -67,8 +67,8 @@ class AttachmentsController < ApplicationController
     if @cud.instructor? || @attachment.released?
       # Default disposition is "attachment" which forces download
       send_data attached_file.download,
-                :filename => @attachment.filename,
-                :type => @attachment.mime_type
+                filename: @attachment.filename,
+                type: @attachment.mime_type
       return
     end
 
--- a/app/controllers/attachments_controller.rb
+++ b/app/controllers/attachments_controller.rb
@@ -65,10 +65,14 @@ class AttachmentsController < ApplicationController
       redirect_to([@course, :attachments]) && return
     end
     if @cud.instructor? || @attachment.released?
-      # Default disposition is "attachment" which forces download
-      send_data attached_file.download,
-                filename: @attachment.filename,
-                type: @attachment.mime_type
+      begin
+        send_data attached_file.download, filename: @attachment.filename,
+                                          type: @attachment.mime_type
+      rescue StandardError
+        COURSE_LOGGER.log("Error viewing attachment '#{@attachment.name}'")
+        flash[:error] = "Error viewing attachment '#{@attachment.name}'"
+        redirect_to([@course, @assessment])
+      end
       return
     end
 
--- a/app/controllers/attachments_controller.rb
+++ b/app/controllers/attachments_controller.rb
@@ -57,19 +57,19 @@ class AttachmentsController < ApplicationController
 
   action_auth_level :show, :student
   def show
-    filename = Rails.root.join("attachments", @attachment.filename)
-    unless File.exist?(filename)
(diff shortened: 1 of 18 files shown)


damianhxy:
Just a note: WSUV's fork did end up making the attachments per-course

https://gitlab.encs.vancouver.wsu.edu/autolab-project/m6/autolab/-/commit/8ece02fb7f5b05a1f25e0cfc4ca04c152d687b0c

app/controllers/attachments_controller.rb
```diff
@@ -57,19 +57,19 @@ def create
...
+      flash[:error] = "No file attached to attachment '#{@attachment.name}'"
       redirect_to([@course, :attachments]) && return
     end
     if @cud.instructor? || @attachment.released?
-      # Set to application/octet-stream to force download
-      send_file(filename, disposition: "inline",
-                          type: "application/octet-stream",
-                          filename: @attachment.filename) && return
+      # Default disposition is "attachment" which forces download
+      send_data attached_file.download,
```
michellexliu, line 69: bit of an edge case, but if an instructor goes into /storage and deletes a file manually from there, this happens:
image

I don't think this is a huge deal though since it's pretty unlikely & showing the donkey error is reasonable for this.

michellexliu (approved):
Tested to verify that everything works; lgtm except for the small comment i left, but that can be addressed later. I'm not sure if we should merge this to master though since it's a breaking change. Maybe we should leave this as its own branch for now / create a branch for all breaking changes if there are more in the future?

damianhxy:
Note to self: Instead of downloading blobs, make use of redirects

https://onrails.blog/2020/12/10/protecting-activestorage-uploads/

damianhxy:
Note: ActiveStorage URLs are permanent and bypass authentication checks. They are hard to guess though, so this is fine.

In future Rails versions, they can be disabled via `config.active_storage.draw_routes = false`

app/controllers/attachments_controller.rb
damianhxy: Added error handling

Screenshot 2023-02-26 at 14 07 29

PR #1812, opened 2023-02-13
Jump to currently enrolled course

lykimchee:
## Description
Upon loading Autolab, if the user is only currently in one course, the user will automatically be redirected to that course's page. However, the user will still be able to navigate to and access the home page (with all courses, including previously completed courses) if desired.

This change was made by adding a conditional for the controller associated with the `/` path, redirecting to the course assessments page if there was only one item in the list of currently enrolled courses.

However, this meant that from within a course, pressing "Autolab" or the home icon in the breadcrumb would also redirect to the course assessments page, preventing users from being able to see anything but their current course. Previous courses and the home page in general would not be accessible. As such, the links for "Autolab" and the breadcrumb home icon were changed to go to `/courses`, which would still load `index` in the controller, and a route for `/courses` was added in the `routes.rb` files to account for this change.

## Motivation and Context
This change improves the user experience of users who are only enrolled in a single course.

## How Has This Been Tested?
Perform the following actions a) when only currently enrolled in one course; b) when currently enrolled in multiple courses; and c) from within a completed course:
- Opening Autolab for the first time
- Going from a course to the home page (of courses) by pressing "Autolab" in the top lefthand corner
- Going from a course to the home page (of courses) by pressing the home icon in the breadcrumbs

## Types of changes

## Checklist:

Currently, some of the changes I added in the `courses_controller` file seem like repetitive code, so I would appreciate any pointers on how to improve that, if possible.


damianhxy:
Could you update the tests to test both the 1 course and > 1 course variants

KesterTan (approved):
LGTM

PR #1821, opened 2023-02-22
Fix bug in GitHub api rate limit check

evanyeyeye:
## Description
`Hash` values cannot be accessed like methods, making the code in the `rescue` clause of `check_github_authorization` invalid. This code is usually unreachable, even with incorrect GitHub key information or none at all. But when reached, it breaks a bunch of things including assessment creation since many pages do a check for whether GitHub integration is enabled.

## How Has This Been Tested?
Before:

image
After:

image

## Types of changes

## Checklist:
- [x] I have run rubocop for style check.

### DIFF
--- a/app/models/github_integration.rb
+++ b/app/models/github_integration.rb
@@ -134,7 +134,8 @@ class GithubIntegration < ApplicationRecord
     begin
       limit = client.rate_limit!
     rescue StandardError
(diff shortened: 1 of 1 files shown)


damianhxy (approved):
Tested via `rails c`, lgtm

PR #1832, opened 2023-03-13
Adds modifySubmissionScores hook

### LINKED ISSUE #1806: Need a more powerful raw_score hook
**Your environment**
- Carnegie Mellon's Autolab at https://autolab.andrew.cmu.edu/ 
- which says "v2.10.0" in the lower right hand corner

**Current behavior**

As documented, the only thing the `raw_score` hook can do is adjust the calculation of the *total score* for an assessment.  CMU's 15-213 has two labs ("bomblab" and "attacklab") that could benefit from being able to do more at the point where this hook is executed, i.e. in between when Ruby parses the JSON score report coming back from the autograder, and when anything gets written to the database.  (Note: I recognize that the behavior of the existing hook probably needs to be preserved for compatibility, so I might really be asking for a _new_ hook or four.  When I say "`raw_score`" below, I mean "`raw_score` or whatever the new mechanism ends up being.")

I have workarounds for the absence of all of these features. My preferred delivery date for them would be "no later than the beginning of summer term" but it isn't any more urgent than that.

**Desired behavior**

In decreasing order of priority:

1. Allow `raw_score` to adjust scores for each _problem_ and not just the total.  Hypothetically this could look something like this:

    ```
    # The autograder returns scores in [0, 1] for each problem.  Scale them to the desired point totals.
    def raw_score(score)
        score["phase1"] *= 10
        score["phase2"] *= 25
        score["phase3"] *= 25
        score["phase4"] *= 35
        score["phase5"] *= 5
        return score.values.sum
    end
    ```

1. Give `raw_score` a documented mechanism for accessing the scores for _previous submissions_ by the same student for the same assessment.  This _may_ currently be possible by mucking around with `assessment_user_data` but I haven't actually tried it and it seems like it involves knowing too much about Autolab's guts.

    I need this because grading for both bomblab and attacklab is cumulative over all submissions.  For attack lab, each submission is only for one problem, so it's necessary to find the most recent submission that touched each problem in order to pull together a complete grade.  For bomb lab, there's a count of "explosions" that should never go down.

1. Allow `raw_score` to generate the scoreboard array.

    Right now the scoreboard array has to come from the autograder itself, but, as described above, the raw score for these labs is incomplete and therefore the scoreboard array is also.  To do the job right, I would need to calculate the scoreboard array _after_ computing the complete cumulative grade, which means in `raw_score` or later. It would also, IMNSHO, be better separation of concerns if the autograder just emitted scores and it was assessment.rb's job to calculate the scoreboard array.

victorhuangwq:
## Description

Adds a modifySubmissionScores hook that allows instructors to override submission scores. 
This does not modify autograded score, but rather the submission score.

## Motivation and Context

Resolves #1806, in a way that allows for a way that does not affect autograded score.

## How Has This Been Tested?

This is tested using a modified hello.rb that modify scores based on previous submissions
[hello_20230319.tar.gz](https://github.com/autolab/Autolab/files/11011979/hello_20230319.tar.gz)
note that it is renamed as gz because github does not allow .tar files

Specifically, the file that was changed is the `hello.rb` file with:
```
  def assessmentVariables
    variables = {}
    variables["previous_submissions_lookback"] = 1000
    variables
  end

  def modifySubmissionScores(scores, previous_submissions, problems)

    scores["Test"] = -(previous_submissions.length)
    
    # Get Test score for previous submission
    scores["Test2"] = previous_submissions[0].scores.find_or_initialize_by(:problem_id => problems.find_by(:name => "Test").id).score

    # Get Test 2 score for previous submission
    scores["problem2"] = previous_submissions[0].scores.find_or_initialize_by(:problem_id => problems.find_by(:name => "Test2").id).score
    
    scores
  end
```

With a lookback of 1, I submitted `hello.c` three times, and this are the submissions scores:
Screenshot 2023-03-19 at 2 27 50 PM

Removing the lookback (thus defaulting to 1000). (Make sure to `Reload config file` on the assessment page), I submitted `hello.c` three times, and this are the submissions scores:
Screenshot 2023-03-19 at 2 31 49 PM

At the same time observe that the autograded scores are not affected:
image

## Types of changes

## Checklist:


app/helpers/assessment_autograde_core.rb
```diff
@@ -412,10 +412,26 @@ def saveAutograde(submissions, feedback)
...
+                                 assessment_id: @assessment.id).where.not(id: submission.id)
+
+          begin
+            scores = @assessment.config_module.modifySubmissionScores(scores, previous_submissions)
+          rescue StandardError => e
+            raise AutogradeError.new("Error in modifySubmissionScores: " + e.message)
+          end
+        end
+
+        scores.keys.each do |key|
```
victorhuangwq, line 432: note that I flipped the loop (it used to be scores being in the outer loop, and the submissions in the inner loop).

Because I want the scores to be modified per submission

victorhuangwq:
I will update the docs with this PR after @michellexliu reviews it

app/helpers/assessment_autograde_core.rb
```diff
@@ -412,10 +412,32 @@ def saveAutograde(submissions, feedback)
       @autograde_prop = @assessment.autograder
 
       # Record each of the scores extracted from the autoresult
-      scores.keys.each do |key|
-        problem = @assessment.problems.find_by(name: key)
-        raise AutogradeError.new("Problem \"" + key + "\" not found.") unless problem
-        submissions.each do |submission|
+      submissions.each do |submission|
```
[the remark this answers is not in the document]
20wildmanj: Indeed, I think it would be pretty substantial

michellexliu:
If the user-created modifySubmissionScores fails, there should be a fallback such that the un-modify score gets saved rather than a 0.
image

Current behavior: correctness displays as a 0 here, even though the unmodified correctness should be 100
image

michellexliu (changes requested):
code looks good, just address the pervious comment about adding a fallback if hook fails

victorhuangwq:
image

app/helpers/assessment_autograde_core.rb
```diff
@@ -412,13 +412,37 @@ def saveAutograde(submissions, feedback)
...
+          # Get previous submissions of the same assessment by the user, excluding the current submission
+          previous_submissions = Submission.where(course_user_datum_id: submission.course_user_datum_id, 
+                                 assessment_id: @assessment.id).where.not(id: submission.id).order(created_at: :desc).limit(previous_submissions_lookback)
+
+          begin
+            scores = @assessment.config_module.modifySubmissionScores(scores, previous_submissions, @assessment.problems)
+          rescue StandardError => e
+            # Append the error message to the feedback
+            feedback_str = "An error occurred while modifying submission scores:\nError message: #{e}\n\n---\n"
+            feedback = feedback_str + feedback
```
victorhuangwq, line 436: here I directly append error message to feedback

app/helpers/assessment_autograde_core.rb
```diff
@@ -412,13 +412,37 @@ def saveAutograde(submissions, feedback)
...
+        if @assessment.overwrites_method?(:modifySubmissionScores)
+          
+          previous_submissions_lookback = if (@assessment.assessment_variable.key?("previous_submissions_lookback"))
+                                            @assessment.assessment_variable["previous_submissions_lookback"]
+                                          else
+                                            1000 # default to 1000
+                                          end 
+                                          
+          # Get previous submissions of the same assessment by the user, excluding the current submission
+          previous_submissions = Submission.where(course_user_datum_id: submission.course_user_datum_id,
```
michellexliu, line 428: Should we filter the submissions such that only submissions that have scores are included? Rn, if you create a submission that finishes grading before the submission before it, when you access the previous submission, it'll set it as 0 instead of using the score from the last-graded submission. We could also leave it up to the lab author to filter this out in their code though, but I guess it depends on whether you think there's any use-cases where lab authors want to use submissions that are still grading in progress.

image

michellexliu (approved):
lgtm other than optional change i commented about

app/helpers/assessment_autograde_core.rb
victorhuangwq: I can add that as a boolean flag, `include_current_grading_submissions` to assessment_variable, defaulting to `false` how does that sound?

victorhuangwq:
```
require "AssessmentBase.rb"

module Hello
  include AssessmentBase

  def assessmentInitialize(course)
    super("hello",course)
    @problems = []
  end

  def assessmentVariables
    variables = {}
    variables["previous_submissions_lookback"] = 1000
    variables["exclude_autograding_in_progress_submissions"] = false
    variables
  end

  def modifySubmissionScores(scores, previous_submissions, problems)

    scores["Test"] = -(previous_submissions.length)
    
    # Get Test score for previous submission
    scores["Test2"] = previous_submissions[0].scores.find_or_initialize_by(:problem_id => problems.find_by(:name => "Test").id).score

    # Get Test 2 score for previous submission
    scores["problem2"] = previous_submissions[0].scores.find_or_initialize_by(:problem_id => problems.find_by(:name => "Test2").id).score
    
    scores
  end

end
```

michellexliu (approved):
tested w/ new assessment .rb file; lgtm!

PR #1833, opened 2023-03-14
Fix annotated PDF download when global annotation is present

damianhxy:
## Description
When rendering an annotated pdf, use `next`, rather than `return` when `annotation.coordinate` is `nil`.

## Motivation and Context
Currently, the "download annotated" button doesn't work for PDF submissions if a global annotation is present. This is because global annotations do not have coordinates associated, and the current logic is to `return` from the loop.

Screenshot 2023-03-13 at 23 21 59

This is not what we want to do, because rails then attempts to render a erb template that does not exist (by default, when control reaches the end of a controller, rails renders the erb template of the same name). Instead, we should simply skip the particular annotation.

Note: Global annotations are purposely not rendered on the pdf, since imo that's not what they're meant for.

## How Has This Been Tested?

1. Create an assessment that accepts PDFs (remember to set handin filename).

2. Make a submission, and then make a PDF annotation (click on the pdf) + a global annotation.

Screenshot 2023-03-13 at 23 23 08

4. Click the "Download Annotated" button (top right corner, next to "Download File").

Screenshot 2023-03-13 at 23 23 40

**Before**
Screenshot 2023-03-13 at 23 16 41

**After**
Screenshot 2023-03-13 at 23 16 47

## Types of changes

## Checklist:

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -230,7 +230,7 @@ class SubmissionsController < ApplicationController
 
       Prawn::Document.generate(@filename_annotated, template: @filename) do |pdf|
         @annotations.each do |annotation|
-          return if annotation.coordinate.nil?
+          next if annotation.coordinate.nil?
 
           position = annotation.coordinate.split(",")
           page  = position[2].to_i


najclark (approved):
Change looks good. Followed testing plan to confirm this fixes the issue.

PR #1839, opened 2023-03-20
Autolab config UI smtp

najclark:
## Description

Activates the SMTP config tab under the Configure Autolab page.

## Motivation and Context

Instead of having to configure SMTP through `development.rb` and `production.rb` files this change provides a UI option to do so.

## How Has This Been Tested?
Feel free to test using a gmail account, but it will likely be easier to just use mailtrap

### Setup SMTP configuration
1. Create a mailtrap account 
2. Go to 'Email Testing', click 'Start Testing'
3. Select 'Ruby on Rails' under 'Integrations'
4. Log into Autolab as an admin
5. Under 'Manage Autolab' click 'Configure Autolab'
6. Go to the SMTP tab and copy the corresponding  credentials from mailtrap into the form
 * communication security protocol: None
 * host: localhost:3000
 * protocol: http
8. Fill in the 'From:' and 'To:' fields under 'Configuration Testing' and click 'Send Test Email'
9. Verify on mailtrap that an 'Autolab Test Email' was sent and that the link in the email navigates to Autolab.

### SMTP still setup after server restart
1. Restart the Autolab server
2. Under 'Manage Autolab' click 'Bulk Email Instructors'
3. Create an email then press 'Send Email'
4. Verify that the email was successfully sent (hopefully one of the instructor emails is real)

### SMTP settings are still in setup form
1. Under 'Manage Autolab' click 'Configure Autolab'
2. Go to the SMTP tab and copy the corresponding  credentials from mailtrap into the form
3. Verify that the settings from mailtrap are still there

## Types of changes

## Checklist:


victorhuangwq:
Why are all the LTI files appearing again in the diff?

app/controllers/smtp_config_controller.rb
```diff
@@ -0,0 +1,113 @@
...
+    old_default_url_options = ActionMailer::Base.default_url_options
+    old_smtp_settings = ActionMailer::Base.smtp_settings
+
+    ActionMailer::Base.perform_deliveries = true
+    ActionMailer::Base.raise_delivery_errors = true
+    ActionMailer::Base.delivery_method = :smtp
+    ActionMailer::Base.default_url_options = default_url_options
+    ActionMailer::Base.smtp_settings = smtp_settings
+
+    @email = CourseMailer.test_email(params[:from_email], params[:to_email])
```
victorhuangwq, line 70: Since the main purpose is to use this setting once for this email, would it be better to use this instead?
https://guides.rubyonrails.org/action_mailer_basics.html#sending-emails-with-dynamic-delivery-options

spec/controllers/admins_controller_spec.rb
```diff
@@ -33,4 +33,76 @@
       end
     end
   end
+
+  describe "#autolab_config" do
```
victorhuangwq, line 37: Wasn't this file merged in in the previous PR?

victorhuangwq (changes requested):
Quick visual scan of the code

victorhuangwq:
Could you perhaps resolve the conflicts that exist first?

app/controllers/admins_controller.rb
```diff
@@ -36,9 +41,17 @@
...
-
-    @lti_config_hash =
-      YAML.safe_load(File.read("#{Rails.configuration.lti_config_location}/lti_config.yml"))
+    if File.exist?("#{Rails.configuration.config_location}/lti_config.yml")
+      @lti_config_hash =
+        YAML.safe_load(File.read("#{Rails.configuration.config_location}/lti_config.yml"))
+    end
+
+    if params[:tmp_smtp_config].present? && File.exist?(params[:tmp_smtp_config])
+      @smtp_config_hash = YAML.safe_load(File.read(params[:tmp_smtp_config]))
```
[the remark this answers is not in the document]
victorhuangwq: @najclark

victorhuangwq (changes requested):
Testing it locally and here are the comments:
- It clears my input if I inputted something wrong and clicked save configurations. That would be annoying as the user have to retype everything if they filled it in wrong the first time
- Didn't quite understand `Default Url Options` and had to google it up. Could we give more description / tooltip?

victorhuangwq:
I got an error when I sent a test mail, probably because I filled it in wrongly. 
image

But I think we should catch that error and show that instead of just throwing it out

victorhuangwq:
Other than that, updating the configuration works.

Tested the following:
- [X] Sending bulk emails 
- [ ] Doing a registration - doesn't work 
image
Seems like there must be a default `from` address field for situations where we don't define a `from` address

app/controllers/smtp_config_controller.rb
najclark: Changed to use dynamic deliver options, however some settings such as `default_url_options` cannot be specified in the dynamic delivery options, so they are still specified using `ActionMailer::Base.default_url_uptions =`.

app/mailers/course_mailer.rb
```diff
@@ -35,4 +35,13 @@ def bug_report(subject, text, user, course)
       sent_on: Time.now
     )
   end
+
+  def test_email(sender, to)
+    mail(
+      subject: 'Autolab Test Email',
+      from: sender,
+      to: to,
+      sent_on: Time.now
```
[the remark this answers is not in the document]
najclark: Not sure what you mean here, the content for the test email comes from `course_mailer/test_email.html.erb`

victorhuangwq:
Tested:
- Registering
- Bulk Email
- Test Email.

Functionality works :)

Small nit:
image
Maybe we make this a section of it's own rather than under `Autolab Url`? And add an example of when the default email address would be used (e.g. in like registration).

PR #1841, opened 2023-03-23
Update MacOS install docs to use Homebrew

20wildmanj:
## Description

Update MacOS docs for an easier installation process, by recommending using homebrew to install rbenv and ruby-build, instead of from cloning the github source. Homebrew is more convenient, and less likely to cause build errors due to weird dependency issues.

Also add a section in troubleshooting for an OpenSSL related LoadError, due to paths mismatch.

## Motivation and Context

Recently had to reinstall Autolab, found that using Homebrew just worked very easily.

## How Has This Been Tested?

Run docs server locally, checked that `installation/troubleshoot/` and `installation/osx` look okay with new changes.

## Checklist:


jlge (approved):
Changes show up when run on local, LGTM! 🥶🥶🥶

PR #1842, opened 2023-03-26
Update Docs to use bash to run init-letsencrypt.sh instead of sh

umar221b:
This is my first PR, please let me know if something does not look right.

## Description
I changed the command to run the LetsEncrypt Initialization script to `sudo ./ssl/init-letsencrypt.sh` so it would run in the specified shell at the top of the file  (`#!/bin/bash`) instead of `sh`.

## Motivation and Context
The LetsEncrypt guide In the docker installation docs uses the command `sudo sh ./ssl/init-letsencrypt.sh` to generate the certificates. This runs the script using `sh`, which breaks on line 10:
```
domains=(example.com) # Update this to your domains
```
with
```
Syntax error: "(" unexpected
```
because `domains=(example.com)` is bash syntax that does not work in `sh`.

## Checklist:

### DIFF
--- a/docs/installation/docker-compose.md
+++ b/docs/installation/docker-compose.md
@@ -101,7 +101,7 @@ There are three options for TLS: using Let's Encrypt (for free TLS certificates)
 2. Ensure that port 443 is exposed on your server (i.e checking your firewall, AWS security group settings, etc)
 3. In `ssl/init-letsencrypt.sh`, change `domains=(example.com)` to the list of domains that your host is associated with, and change `email` to be your email address so that Let's Encrypt will be able to email you when your certificate is about to expire
 4. If necessary, change `staging=0` to `staging=1` to avoid being rate-limited by Let's Encrypt since there is a limit of 20 certificates/week. Setting this is helpful if you have an experimental setup.
-5. Run your modified script: `sudo sh ./ssl/init-letsencrypt.sh`
+5. Run your modified script: `sudo ./ssl/init-letsencrypt.sh`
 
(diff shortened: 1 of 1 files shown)


damianhxy:
Hi @umar221b, thanks for the fix! We did not notice this earlier because we usually run `sudo bash ./ssl/init-letsencrypt.sh`.

Just curious: are you planning to use Autolab at an academic institution? We would love to keep in touch (and add to the list on https://autolabproject.com/)!

damianhxy (approved):
Tested that the original command fails, and that the new command works.

umar221b:
Thanks @damianhxy!

> Just curious: are you planning to use Autolab at an academic institution? We would love to keep in touch (and add to the list on https://autolabproject.com/)!

Yes, I am trying to setup Autolab and figure out assignment creation to use it in our programming introductory level course at [HTU](https://www.htu.edu.jo/) in Jordan. I just filled the form on the website. Thanks for the platform!

damianhxy:
Thank you for filling up the form -- feel free to join our slack (from our readme) if you require any further assistance!

I saw your comment regarding `sassc`, I'll make a note in the documentation about the memory requirement, thanks for bringing the issue to our attention 😄

umar221b:
I would love to join Slack. Do I need to be invited? The link in the docs points to a workspace that requires an `@andrew.cmu.edu` email address.

damianhxy:
Ah I see, I didn't realize that there's no easy way to let anyone join a Slack channel -- I've invited you!

PR #1863, opened 2023-03-31
Make moss page for current course only

### LINKED ISSUE #1861: Moss Cheat Checker page crashes if there are no assessments in the database
**Your environment**
I am using the latest Docker image.

**Steps To Reproduce**

1. Create a new Autolab instance or delete all the assessments on your current instance.
2. Go to "Manage Course" on any course.
3. Click on "Run Moss cheat checker".

**Current behavior**
Page crashes with a 500 error inside `views/courses/moss.html.erb` at `<div id='<%= a.id %>OptionsDiv' style="margin-right: 10px;" >`

**Expected behavior**
Page does not crash.

**Screenshots**
Screenshot 2023-03-30 at 2 10 50 AM

The reason of the crash is that the variable `a` is used in a couple of locations outside the loop that defines it. Not quite sure these parts it is used in should be in the loop or the usages should be removed (I believe the latter), but if the usages are not needed we could just remove them from the ids of the divs. Happy to submit a PR that removes the usages of `a` outside the loop, and maybe displays a message "to add assessments to get started with plagiarism checking"?

Let me know what you think.

umar221b:
## Description
1. Remove usages of undefined `a` variable in `views/courses/moss.html.erb`.
2. If the course does not have any assessments, block the moss page with a message and a link to install an assessment.
3. Do not show all courses and their assessments in this page, only the current course and its assessments.

## Motivation and Context
Fixes #1861 in addition to making the page for the current course only, which makes more sense as the page is opened within a course's context.

## How Has This Been Tested?
Tested locally on the latest docker images.

Here is how the page looks when there is at least 1 assessment:

Screenshot 2023-03-31 at 3 26 50 AM

And here is how it looks when there are no assessments:

Screenshot 2023-03-31 at 3 26 20 AM

I also submitted a few submissions to moss confirm nothing broke there.

## Types of changes

### DIFF
--- a/app/views/courses/moss.html.erb
+++ b/app/views/courses/moss.html.erb
@@ -6,7 +6,7 @@
 	<ul class="collection" style="text-align: center; font-size: 18px;">
 		<li class="collection-item add">
 			<p>You do not seem to have any assessments in this course to be able to run the Cheat Checker</p>
-			<a title="Install an assessment" class="btn btn-large red darken-3" href=<%= install_assessment_course_assessments_path(@course) %> style="font-size: 18px;">Add an Assessment</a>
+			<%= link_to "<i class='material-icons left'>assignment</i> Install Assessment".html_safe, install_assessment_course_assessments_path(@course), { title: "Install Assessment", class: "btn btn-large red darken-3" } %>
 		</li>
 
 	</ul>
@@ -77,7 +77,7 @@
 							<span> Language </span>
 						</label>
 						<br>
-						<div id='<%= a.id %>OptionsDiv' style="margin-right: 10px;" >
+						<div style="margin-right: 10px;" >
 							<div class="input-field">
 								<%= select_tag :language_selection, options_for_select([
 																																				 ["ASCII", "ascii"],
@@ -116,7 +116,7 @@
 							<span>Max Lines </span>
 						</label>
 						<br>
-						<div id='<%= a.id %>OptionsDiv' style="padding-left:10px;" >
+						<div style="padding-left:10px;" >
 							<%= number_field_tag "max_lines", "10", placeholder: "If you don't know what this is please read the Moss documentation before altering this value.", min: 1 %>
 						</div>
 					</li>
@@ -125,7 +125,7 @@
 							<%= check_box_tag "box_basefile", "1",false, :class => "filled-in" %>
 							<span>Base File Tar </span>
 						</label>
-						<div id='<%= a.id %>OptionsDiv' style="padding-left:10px;" >
+						<div style="padding-left:10px;" >
 							<div class="file-field input-field">
 								<div class="btn">
 									<span>Base File</span>
--- a/app/views/courses/moss.html.erb
+++ b/app/views/courses/moss.html.erb
@@ -1,177 +1,158 @@
 <% @title = "Moss Cheat Checker" %>
 
 <% content_for :javascripts do %>
-  <script type="text/javascript">
-    /**
-     * toggles the "style.display" of the element with the given id.
-     **/
-    function toggleOptions(id) {
-      var show = document.getElementById(id);
-      show.style.display = (show.style.display == "none")? "block" : "none";
-    }
+	<script type="text/javascript">
+		/**
+		 * toggles the "style.display" of the element with the given id.
+		 **/
+		function toggleOptions(id) {
+			var show = document.getElementById(id);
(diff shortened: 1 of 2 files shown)


damianhxy:
What do you think about making the "Add an assessment" button consistent with the "Install assessment" button (from the course home page?)

Screenshot 2023-04-02 at 15 58 13

umar221b:
> What do you think about making the "Add an assessment" button consistent with the "Install assessment" button (from the course home page?)
> 
> Screenshot 2023-04-02 at 15 58 13

Yes, that is a good idea. I will do it.

app/views/courses/moss.html.erb
```diff
@@ -1,177 +1,168 @@
...
+				</h5>
+				<ul id="<%= @course.id %>Assessments" class="moss-inner-list">
+					<% for a in @course.assessments do %>
+						<li>
+							<label>
+								<%= check_box_tag "assessments[#{a.id}]","1",false, :class => "filled-in" %>
+								<span> <%= a.display_name %> </span>
+							</label>
+							<br>
+							<div id='<%= a.id %>OptionsDiv' style="padding-left:10px;" >
```
damianhxy, line 44: Looks like you accidentally undid your removal of the `a.id`s.

I would suggest removing all four occurrences of `id='<%= a.id %>OptionsDiv'` (instead of just the `a.id`) -- it doesn't make sense to have multiple elements with the same `id`

app/views/courses/moss.html.erb
```diff
@@ -1,177 +1,168 @@
...
-      });
-    }
-  
-  </script>
-<% end %>
+<% if @course.assessments.empty? %>
+	<ul class="collection" style="text-align: center; font-size: 18px;">
+		<li class="collection-item add">
+			<p>You do not seem to have any assessments in this course to be able to run the Cheat Checker</p>
+			<a title="Install an assessment" class="btn btn-large red darken-3" href=<%= install_assessment_course_assessments_path(@course) %> style="font-size: 18px;">Add an Assessment</a>
```
damianhxy, line 9: I would suggest wrapping the href attribute in quoates

umar221b: I changed it to use the same line in the index page, but used the url helper for the link to pass the course.

damianhxy (approved):
Verified that MOSS still works, changes work as expected on course with assessments / no assessments.

Left a few nits, but otherwise lgtm!

umar221b:
@damianhxy, resolved all the nits, let me know if anything else needs addressing!

app/views/courses/moss.html.erb
```diff
@@ -1,177 +1,166 @@
...
+				</ul>
+			</li>
+		</ul>
+
+		<br>
 
-  <br>
+		<h4>Step 3:</h4>
+		<p> (Optional) Upload an archive containing additional files you'd like Moss to compare against.</p>
+		<p>(Note: the archive must contain only regular files.  Nested archives wil not be extracted.)</p>
```
damianhxy, line 148: One last thing -- I realized that there's a typo "wil" here. While we're on the topic of MOSS, I figured this might be a good time to fix it.

damianhxy:
LGTM. Thanks again for the PR!

umar221b:
Hi @damianhxy, sorry to go back to this, but I just had a thought that a very usual use case in subsequent runs for a course is to run plagiarism against previous runs of that course. Do you think my work here:

>Do not show all courses and their assessments in this page, only the current course and its assessments).

should be reverted?

damianhxy:
That's also a thought that I had. However, I don't think that's necessary, files from previous iterations can be uploaded via the additional files tar for comparison

damianhxy:
@umar221b On second thoughts, we had an instructor today request this previously ability to compare across semesters. 
Also, it might be tricky to use the additional files tar feature since it doesn't recursively expand archives.

Would it be difficult to revert this change?

umar221b:
> @umar221b On second thoughts, we had an instructor today request this previously ability to compare across semesters. Also, it might be tricky to use the additional files tar feature since it doesn't recursively expand archives.
> 
> Would it be difficult to revert this change?

It should not be, let me try to work on something over the weekend.

damianhxy:
This has been resolved by #2015

umar221b:
> This has been resolved by #2015

Terribly sorry I did not get a chance to work on this.

damianhxy:
No worries, thank you for all your contributions!

PR #1878, opened 2023-04-13
Extension metrics

KesterTan:
## Description
Previously, an instructor had to go through every homework and manually track the number of extensions requests a student had per course. This change allows instructors to filter students in a course by the number of extensions that was given in Student Metrics. 

## How Has This Been Tested?
1. Create a new course or use an existing course
2. Upload a few assessments if the course does not have any already
3. Click into an assessment, select manage extensions, and give extensions to any student enrolled in the course. 
4. Test it out with different assessments, different students, and different extension lengths (especially try giving the same student an extension in multiple assessments).
5. Select Student Metrics, type in the amount of extensions requests you want to filter for, and click save
6. Check the watchlist to see if students appear in the watchlist. If not, try refreshing the list using the refresh button or manually refreshing the page.

## Types of changes

## Checklist:

### DIFF
--- a/app/models/risk_condition.rb
+++ b/app/models/risk_condition.rb
@@ -1,7 +1,7 @@
 class RiskCondition < ApplicationRecord
   serialize :parameters, Hash
   enum condition_type: { no_condition_selected: 0, grace_day_usage: 1, grade_drop: 2,
-                         no_submissions: 3, low_grades: 4 }
+                         no_submissions: 3, low_grades: 4, extension_requests: 5 }
 
   has_many :watchlist_instances, dependent: :destroy
   belongs_to :course
@@ -11,6 +11,7 @@ class RiskCondition < ApplicationRecord
   # :grade_drop => :percentage_drop, :consecutive_counts
   # :no_submissions => :no_submissions_threshold
   # :low_grades => :grade_threshold, :count_threshold
+  # :extension_requests => :extension_count
 
   def self.create_condition_for_course_with_type(course_id, type, params, version)
     # parameter check shouldn't surface to user and is for sanity check during development only
@@ -21,6 +22,7 @@ class RiskCondition < ApplicationRecord
        (type == 3 && (params[:no_submissions_threshold].nil? || params.length != 1)) ||
        (type == 4 && (params[:grade_threshold].nil? ||
                       params[:count_threshold].nil? || params.length != 2))
+      (type == 5 && (params[:extension_count].nil? || params.length != 1))
       raise "Invalid update parameters for risk conditions! "\
             "Make sure your request body fits the criteria!"
     end
--- a/app/models/risk_condition.rb
+++ b/app/models/risk_condition.rb
@@ -21,8 +21,8 @@ class RiskCondition < ApplicationRecord
                       params[:consecutive_counts].nil? || params.length != 2)) ||
        (type == 3 && (params[:no_submissions_threshold].nil? || params.length != 1)) ||
        (type == 4 && (params[:grade_threshold].nil? ||
-                      params[:count_threshold].nil? || params.length != 2))
-      (type == 5 && (params[:extension_count].nil? || params.length != 1))
+                      params[:count_threshold].nil? || params.length != 2)) ||
+       (type == 5 && (params[:extension_count].nil? || params.length != 1))
       raise "Invalid update parameters for risk conditions! "\
             "Make sure your request body fits the criteria!"
     end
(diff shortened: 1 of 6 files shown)


app/views/metrics/index.html.erb
```diff
@@ -232,9 +232,24 @@
...
+        <span class="padded-span">
+          <input type="text" placeholder="1" id='extension_requests_count'>
+        </span>
+        extensions.
+      </div>
+    </div>
+
     <br>
     <% if @cud.instructor? %>
-      <button class="ui disabled submit button" id="save">SAVE</button>
```
damianhxy, line 237: Why are you removing `disabled`?

In metrics.js, it is removed if there was a change to the form.
Screenshot 2023-04-16 at 15 19 14

victorhuangwq:
Testing locally

**Hover over error**
1.  Gave User0 1 day extension on Lab0, and User1 2 day extension on Homework2
2. Selected Students who have greater than or equal to 1 extensions.
3. Both students appeared, but the hover over are both wrong

image
image

**Did not clear out previous metrics / Regression**
1. Students who did not submit at least  1 assignment(s)
2. Look at metrics, many students should appear
3. Select Students who have greater than or equal to 1 extensions, and unselect Students who did not submit at least  1 assignment(s)
4. Now you get students who have made 0 extension requests
image.

app/views/metrics/index.html.erb
KesterTan: I did this because, the default now is that when you enter into metrics, a checkbox is marked but the save button is disabled even when you fill in the field, which is quite non-intuitive. But now thinking about it the solution should be to not have any boxes checked initially instead of removing the disabled class.

victorhuangwq:
On the front end, I am able to set 
image

although it seems to be clamped to 1 in the backend

app/models/risk_condition.rb
```diff
@@ -21,6 +22,7 @@ def self.create_condition_for_course_with_type(course_id, type, params, version)
        (type == 3 && (params[:no_submissions_threshold].nil? || params.length != 1)) ||
        (type == 4 && (params[:grade_threshold].nil? ||
                       params[:count_threshold].nil? || params.length != 2))
```
victorhuangwq, line 24: it seems like you have missed out a || - this might not have been caught because this function isn't used?

app/models/watchlist_instance.rb
```diff
@@ -506,6 +513,39 @@ def self.add_new_instance_for_cud_grace_day_usage(course, condition_id,
                           violation_info: violation_info)
   end
 
+  def self.add_new_instance_for_cud_extension_requests(course, condition_id, cud, extension_count)
+    asmts = Assessment.where(course_id: course.id).pluck(:id)
+    violation_info = {}
+    num_of_extensions = 0
+
+    asmts.each do |asmt|
+      extensions = Extension.where(course_user_datum_id: cud.id, assessment_id: asmt)
```
victorhuangwq, line 522: `.count` should be used instead of using a each loop https://api.rubyonrails.org/classes/ActiveRecord/Calculations.html#method-i-count

app/models/watchlist_instance.rb
```diff
@@ -506,6 +513,39 @@ def self.add_new_instance_for_cud_grace_day_usage(course, condition_id,
...
+        num_of_extensions += 1
+      end
+    end
+
+    if num_of_extensions >= extension_count
+      asmts.each do |asmt|
+        extensions = Extension.where(course_user_datum_id: cud.id, assessment_id: asmt)
+        aud = AssessmentUserDatum.find_by(course_user_datum_id: cud.id, assessment_id: asmt)
+        indiv_extensions_count = 0
+        extensions.each do
```
victorhuangwq, line 533: similarly here, an `.count` should be used https://api.rubyonrails.org/classes/ActiveRecord/Calculations.html#method-i-count

victorhuangwq (changes requested):
Issue with ability to set to more than or equals to 0 on front end. As well as other nits.

victorhuangwq:
I don't think you took into account Considered categories when making your calculations:
image
image

I excluded Labs, and I get lab categories extensions

app/models/watchlist_instance.rb
victorhuangwq: nit: I don't think this has been addressed

victorhuangwq (approved):
Functionality works as expected. Small nit that was not addressed previously.

LGTM

app/models/watchlist_instance.rb
KesterTan: Fixed.

app/models/watchlist_instance.rb
KesterTan: Changed the part about counting the number of extensions.

PR #1885, opened 2023-04-18
Delete unused destroy html erb file

20wildmanj:
## Description

- Delete `/app/views/assessments/destroy.html.erb` as it is an unused view file which hasn't been modified since the initial open source commit:
Screen Shot 2023-04-18 at 11 51 23
- Note: The `assessments#destroy` action is a `DELETE` method, which redirects to the courses page, so the action doesn't use this page
## Motivation and Context

- Remove dead code so it doesn't need to be linted with the upcoming erb-linter
## How Has This Been Tested?

- Delete an assessment, see it still works
- Review `assessments#destroy` action, as well as routes to see that there is no route that uses this page


damianhxy (approved):
Tested that deleting an assessment still works, found no evidence that view is ever used (checked `routes.rb`, grepped for `render` etc.)

LGTM

PR #1888, opened 2023-04-21
Fix hovering over top navbar to be colored white instead of blue for Metrics page

20wildmanj:
## Description

When in the metrics page, hovering over the top navbar changes the color of the items' text to turn blue, when it should stay white:
Screen Shot 2023-04-21 at 15 37 29

Developer console shows that this color originates from the metrics scss file, but no such definition can be found in the file itself, so I believe this is a side-effect of `semantic-ui`, which gets imported in `metrics.scss`.

Screen Shot 2023-04-21 at 15 37 51

To fix, set the hover color for `a` to be white in `metrics.scss`

## How Has This Been Tested?

- Hover over top navbar and see that items' text color stays white
- browse through rest of page to see that other hover colors stay the same

## Types of changes

### DIFF
--- a/app/assets/stylesheets/metrics.scss
+++ b/app/assets/stylesheets/metrics.scss
@@ -235,3 +235,7 @@ span.instances .segment {
   display: flex;
   align-items: center;
 }
+
+a:hover {
+  color: $autolab-white;
+}


jlge (approved):
It is beautiful LGTM

PR #1870, opened 2023-04-09
Google OAuth Config UI

najclark:
## Description
Adds the OAuth integration UI page
image

## Motivation and Context
Resolves #1869 as part of Autolab 3.0

### Standard OAuth still works
1. Log out of Autolab
2. Confirm ability to log back in using standard auth

### Configure Google OAuth
1. Follow  [these instructiosn](https://developers.google.com/identity/protocols/oauth2) to obtain a `Client ID` and `Client Secret`
 * Use `http://localhost:3000/auth/users/auth/google_oauth2/callback` for Authorized redirect URIs
2. Click on 'Manage Autolab' > 'Configure Autolab' > 'OAuth Integration' tab
3. Fill out the form with the OAuth credentials from the Google Cloud console
4. Click 'Update Configuration'
5. Verify that a check now appears next to 'Google OAuth'
6. Log out of Autolab
7. Confirm a 'Sign in with Google' button is now available
image
8. Click 'Sign in with Google' and confirm that you are able to log in with a google account

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have run rubocop for style check. If you haven't, run `overcommit --install && overcommit --sign` to use pre-commit hook for linting
- [x] My change requires a change to the documentation, which is located at [Autolab Docs](https://github.com/autolab/docs)
- [ ] I have updated the documentation accordingly, included in this PR

PR #1892, opened 2023-04-23
OAuth flow fix

### LINKED ISSUE #1873: [UI Config] Fix Google OAuth flow

najclark:
## Description
These changes fix the oauth flow to be more consistent with the expected behavior seen on other websites. Previously, after signing in with an OAuth provider, the user would be redirected to the new user page, where they were prompted to create a password. 

Now when a user signs in with an OAuth provider, they are redirected to a page where they can see all of their courses.
Resolves #1873 

## How Has This Been Tested?
Follow the testing steps in #1870 to configure Google OAuth. 

### New user OAuth
1. Log out of Autolab
2. Click 'Sign in with Google' and select an account that **is not** currently registered on Autolab
3. Verify that the new user is redirected to the 'welcome to autolab' page

### Existing user OAuth
1. Log out of Autolab
2. Click 'Sign in with Google' and select an account that **is** currently registered on Autolab
3. Verify that the existing user is redirected to their courses page
4. Log out of Autolab again
5. Log in with the existing user's email and password to verify that standard auth still works

## Types of changes

## Checklist:


app/models/user.rb
```diff
@@ -99,28 +99,60 @@ def self.find_for_shibboleth_oauth(auth, _signed_in_resource = nil)
...
-                                 uid: data["uid"])
       elsif (data = session["devise.shibboleth_data"])
         user.email = data["uid"] # email is uid in our case
-        user.authentications.new(provider: "CMU-Shibboleth",
-                                 uid: data["uid"])
+      else
+        return # early exit if not called from an oauth callback
       end
+
+      User.assign_random_password user
```
victorhuangwq, line 128: So does this mean that a user is assigned a random password if they log in using oauth? does this happen once or every single time?

victorhuangwq (approved):
This is amazing.
Functionality wise it works well, but I am having difficulty understanding what exactly you have done to the code.

Could you explain to me two cases?:
1. What happens when a completely new user logs in, how is the account created etc?
2. What happens when I log in again?

At the same time, I'm going to tag @20wildmanj to have a second eye on this

najclark:
1. What happens when a completely new user logs in, how is the account created etc?
When a completely new user logs in, a new autolab user is created with the details provided by the OAuth provider (email, first name, last name) and a random password. Additionally a new `Authentication` record is created to mark that this user has been authenticated with the OAuth provider.

2. What happens when I log in again?
When the completely new user logs in again with OAuth, `User.find_for_google_oauth2_oauth` finds the previously created `Authentication` record and the user is signed in.
If the completely new user wants to log in with standard auth, they will need to reset their password because they have no way of knowing the auto-generated password.

victorhuangwq:
nit: could you make the text white on hover
image

victorhuangwq:
So I tested out a funky user flow:

- I tried registering myself normally using my andrew email using the normal registration flow
- I don't verify myself from the email that was sent
- I then tried to log in using OAuth using my andrew email
- It tells me that I have to confirm my account
image

I'm wondering if a student who signed up normally, and then used OAuth, should just be allowed to skip the verification email?

victorhuangwq:
Question:
Is using OAuth equivalent to checking the "remember me" button?

najclark:
Will fix the flow you mentioned to skip confirmation when an oauth authentication method is added to a user.

Right now OAuth is equivalent to the standard sign in without _remember me_ checked (a session is generated). It's a pretty easy change to make it equivalent to _remember me_ (a session and remember me cookie is generated) but I wasn't sure what the standard practice with OAuth was.

app/models/user.rb
```diff
@@ -99,28 +99,64 @@ def self.find_for_shibboleth_oauth(auth, _signed_in_resource = nil)
...
-                                 uid: data["uid"])
+        User.assign_random_password user
+        User.add_oauth_if_user_exists session
       end
     end
   end
 
+  def self.add_oauth_if_user_exists(session)
+    email, provider, uid = "", "", ""
+    if (data = session["devise.facebook_data"])
```
victorhuangwq, line 136: are we still intending to do facebook in the future?

victorhuangwq (approved):
LGTM

20wildmanj:
Retroactively LGTM

PR #1898, opened 2023-05-01
Fix 500 error from going to users/id for non-existent user-id

20wildmanj:
## Description

Currently, navigating to `users/:id` with a non-existent user-id results in a `ActiveRecord::RecordNotFound` error because the users controller uses `User.find(params[:id])` instead of finding by id specifically. 
Screenshot 2023-05-01 at 2 54 48 PM

This PR changes this line to use `User.find_by id: params[:id]` so that the nil check afterwards works correctly and the user is redirected to the user path correctly.

## How Has This Been Tested?

Navigate to `users/:id` with a non-existent user-id on nightly/cmu deployment, see 500 error

Screenshot 2023-05-01 at 2 54 32 PM

On this branch, navigate to `users/:id` with a non-existent user-id, see that you are redirected to `/users` and a error flash pops up

Screenshot 2023-05-01 at 3 04 50 PM

## Types of changes

## Checklist:

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -33,7 +33,7 @@ class UsersController < ApplicationController
   # based on current user's role
   action_auth_level :show, :student
   def show
-    user = User.find(params[:id])
+    user = User.find_by id: params[:id]
     if user.nil?
       flash[:error] = "User does not exist"
       redirect_to(users_path) && return


damianhxy (approved):
Tested that error message shows correctly now.

LGTM

PR #1916, opened 2023-06-03
Improve Scheduler UI, visual run isolation, add until and disable options

### LINKED ISSUE #955: scheduler job disable
scheduler jobs should have an enable checkbox, so they can be disabled without being deleted or setting a bogus next time.

A "don't run after" time value may also be useful

### LINKED ISSUE #1004: Visual run can break scheduled jobs
When scheduled jobs run (and assuming the visual run feature has not been used since the web process was last restarted), the rails process forks before the module is loaded, so on each run:
- The action code is always loaded, because the action path has not previously been require'd
- The `Updater` object is not defined before the action code is loaded, so there are no issues with redefining it.

Once the Visual Run feature is used to execute a scheduled job action, neither of those things is true, because the `require` (and `Updater.update`) are run in the normal rails handler process. Furthermore, this affects future background invocations forked from the same process!

The lack-of-forking has two significant effects
- The instructor action code is not reloaded on future runs (`require` is "require once"), so edits of the action script are ignored until the rails process restarts
- The type of `Updater` cannot change (some scripts declare `Updater` to be a `Class`, others make it a `Module`. rails allows classes and modules to be redefined, but it does not allow you to define a `Module` when a `Class` of the same name exists or vice versa.)

damianhxy:
### Summary generated by Reviewpad on 03 Jun 23 09:41 UTC
This pull request contains various changes across the project. The changes include modifying the formatting of various files, adding new files, updating database schema, modifying controller actions, and adding a new item to the navigation list. Specifically, changes have been made to the scheduler files, such as adding a new description file, updating the view file, introducing a new status method in the controller, and modifying the run and edit actions. In addition, help text has been added to various fields, and the header of the form has been updated. Finally, a new migration file was added that includes a new "disabled" column with a default value on the scheduler table; the Until column was added and its default value set to the current timestamp.

## Description

- General improvements to scheduler UI
- Allow scheduler actions to return output
- Use fork for visual run
- Add scheduler docs, touch-up some headers in other docs
- Adds ability to disable a scheduler
- Adds ability to set an "until" date for a scheduler
- Convert scheduler path to string before require

## Motivation and Context

Currently, visual run only loads a scheduler file once, and does so outside of a forked context. This leads to issues where changes are not reflected in visual runs. This might also cause issues to scheduled runs.

This PR also adds the ability to set an "until" date for a scheduler, beyond which point the scheduler will stop running. It also adds the ability to disable a scheduler.

Fixes #955
Fixes #1004 (subsumes https://github.com/autolab/Autolab/commit/17a6923cb83e3beb7ba570f0a09f002a62d9922c)
Subsumes https://github.com/autolab/Autolab/commit/fc4b59c3fce89d1305b93f8d0845010c9575bae4

## How Has This Been Tested?

Place the file `test.rb` inside the Autolab root directory, with the following contents:
```ruby
module Updater
    def self.update(course)
        out = ""
        out << "my output\n"

        out
    end
end
```
**General functionality**
- Create a scheduler with action `test.rb`, set the `until` date to be in the future.
- Note that the status on the home page is `running`
- Note that the `next` time is in the future (since the scheduler just ran)
- Perform a visual run ("play" icon) ensure you see the output `my output`
- Modify the file (e.g. change the module name), perform another visual run, ensure you see an error

**Note:** your path will be different from my screenshots; I placed mine inside the course folder and named the scheduler accordingly

Screenshot 2023-06-03 at 22 52 48
Screenshot 2023-06-03 at 22 54 24

- Click on the show button ("info" icon), ensure information is correct
- Edit the scheduler, ensure it is populated with the correct info
- Successfully edit the scheduler
- Successfully delete the scheduler

**Disabled / Stopped**
- Create a disabled scheduler, with `until` in the future and `next` in the past
- Note that the status on the home page is `disabled`, and the `next` time is unchanged
- Now un-disable it, and set `until` in the past
- Note that the status on the home page is `stopped`, and the `next` time is still unchanged

Screenshot 2023-06-03 at 22 51 57

**Others**
- Take a look at the documentation
- Note: the use of headers were modified for some files, because using h2 at the top results in two headers displaying: a default h1 header and the h2 header from the file. But if we use a h1 header, there is only one header. Other header sizes have been adjusted with this in mind.

## Types of changes

## Checklist:


damianhxy:
@cg2v This PR resolves two long-standing scheduler issues that you've raised, as well as incorporates changes from one of your patches. I hope this addresses them in a satisfactory manner!

docs/features/schedulers.md
```diff
@@ -0,0 +1,45 @@
+# Schedulers
+
+Schedulers are instructor scripts that run periodically. They can be managed via `Manage Course > Manage schedulers`.
+
+!!! info "Interval Guarantees"
```
KesterTan, line 5: Nit: Maybe make this a note?

KesterTan:
Screenshot 2023-06-10 at 12 27 04 AM

Another Nit, is it possible to make the disabled button be bold when the checkbox is selected?

KesterTan:
LTGM, all scheduler functions work as expected.

docs/features/schedulers.md
damianhxy: Resolved offline -- the convention in the docs are to use admonitions

damianhxy:
> Another Nit, is it possible to make the disabled button be bold when the checkbox is selected?

Resolved offline -- this is not the convention, and there is sufficient notice since the index page displays scheduler status

PR #1936, opened 2023-07-07
Make problem names be unique within assessments

cg2v:
### Summary generated by Reviewpad on 07 Jul 23 19:00 UTC
This pull request makes problem names unique within assessments. Problem names are used as lookup keys, so it is important for them to have the uniqueness property of a key in both the model and the database. The patch adds a validation to ensure the uniqueness of problem names in the Problem model, and creates a migration to add a unique index on the `problems` table in the database. This ensures that each problem name within an assessment is unique.

## Description
Update the problem model and database schema to apply a uniqueness constraint on the names of Problem objects (within the scope of an assessment)

## Motivation and Context
Problem names are used as lookup keys, and so should have the uniqueness property of a key. Duplicate problem names could result in multiple scoreboard columns with the same name and different values, or some submissions having a value in one column and others using a different one.

## How Has This Been Tested?
Unit tests have been run with this patch applied

## Types of changes

- [ X ] Bug fix (non-breaking change which fixes an issue)

## Checklist:

If an existing installation already has duplicate problem names, the database migration will fail.

### DIFF
--- a/app/models/problem.rb
+++ b/app/models/problem.rb
@@ -11,6 +11,7 @@ class Problem < ApplicationRecord
   has_many :annotations, dependent: :destroy
 
   validates :name, :max_score, presence: true
+  validates :name, uniqueness: { case_sensitive: false, scope: :assessment_id }
   validates_associated :assessment
 
   after_commit -> { assessment.dump_yaml }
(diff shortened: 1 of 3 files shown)


damianhxy (approved):
Verified that
- (non-duplicate) problems can be created
- duplicate problems can not be created
- uniqueness check occurs within the scope of an assessment

Screenshot 2023-07-13 at 00 50 44

Note that we get two errors instead of just one, because it displays the error from the assessment containing the problem (the assessment is invalid by extension of problems being invalid). That said, it's not really an issue worth addressing

PR #1945, opened 2023-07-18
Prevent failing import from file system from deleting assessment files

### LINKED ISSUE #1944: Assessment directory deleted as side effect of "Error loading yaml: Name in yaml doesn't match"
**Your environment**
CMU's Autolab (v2.10.0)

**Steps To Reproduce**

1. Create an assessment from scratch
2. Have something slightly wrong with the assessment .yml file: I saw this happen because I'd given the wrong name for the assessment *inside* the .yml file, but I wouldn't be surprised if there are other ways to trigger it.
3. Attempt to install the assessment
4. You get an error (good)
5. The entire assessment directory is deleted (VERY BAD)

**Current behavior**

At least some errors in parsing the assessment .yml file will cause the entire assessment directory to be deleted. There is no notification that this has happened.

**Expected behavior**

Autolab should *never* delete the entire assessment directory, no matter what. Why is it even *capable* of doing this? Someone must've written actual code to make it do this. What were they thinking?

(yes I'm a little grumpy right now, I just lost an hour of work and I'm lucky it wasn't *days* of work)

20wildmanj:
### Summary generated by Reviewpad on 25 Jul 23 04:32 UTC
This pull request includes three patches. 

The first patch (PATCH 1/3) makes changes to the assessments_controller.rb file. It adds a check to ensure that each assessment has an associated yaml file with a name field that matches the filename. It also handles errors when displaying existing assessments.

The second patch (PATCH 2/3) also modifies the assessments_controller.rb file. It updates the comments to clarify the requirements for each assessment's associated yaml file, and it fixes some formatting issues.

The third patch (PATCH 3/3) addresses comments from a previous pull request. It renames the isTarImport parameter to cleanup_on_failure and updates the corresponding code. It also improves error handling and removes unnecessary code.

Overall, these patches enhance the functionality and error handling of the assessments_controller.rb file.

## Description

- use `isTarImport` parameter to`importAssessment` to determine whether import is from tarball (aka from `importAsmtTar`) or from file system
- delete assessment files only is import was from a tarball import, otherwise leave files as-is
- Add pre-check in `install_assessment` to check for invalid name field in yaml to prevent installation of invalid assessment (and thus prevent yaml from getting wiped out, which happens if the error was only caught during deserialization)

## Motivation and Context

- Closes #1944

## How Has This Been Tested?

- Delete existing assessment (go to "Edit Assessment"->"Delete Assessment")
- Find assessment in file system, navigate to yaml, change `name` field to not match filename
- Go to "Install Assessment," see following error appears and that assessment doesn't show up under "Import from File System"
  Screenshot 2023-07-17 at 9 43 24 PM
- Fix the yaml name, then change config rb module name so that it doesn't match the filename / assessment name
- Try installing the assessment, see following error:
  Screenshot 2023-07-17 at 9 46 42 PM
- verify that assessment directory is still there

Regression:
- import from tarball a buggy assessment (such as `spec/fixtures/assessments/homework02-yaml-name-field-wrong.tar`), and try importing it
- see error is displayed and that assessment directory is deleted from course directory
- Upload a normal tarball / import from file system a normal directory and see that import succeeds
- Run `bundle exec rake spec` multiple times and see that tests pass

## Types of changes

## Checklist:

### DIFF
--- a/app/controllers/assessments_controller.rb
+++ b/app/controllers/assessments_controller.rb
@@ -115,7 +115,7 @@ class AssessmentsController < ApplicationController
         # add line break if adding to existing error message
         flash.now[:error] = flash.now[:error] ? "#{flash.now[:error]} <br>" : ""
         flash.now[:error] += "An error occurred while trying to display an existing assessment " \
-            "on file directory #{filename}: assessment file names must only contain lowercase " \
+            "from file directory #{filename}: assessment file names must only contain lowercase " \
             "letters and digits with no spaces"
         flash.now[:html_safe] = true
         next
@@ -129,8 +129,8 @@ class AssessmentsController < ApplicationController
                                ))
         unless props["general"] && (props["general"]["name"] == filename)
           flash.now[:error] = flash.now[:error] ? "#{flash.now[:error]} <br>" : ""
-          flash.now[:error] += "An error occurred while trying to display an existing assessment" \
-          "on file directory #{filename}: Name in yaml (#{props['general']['name']}) " \
+          flash.now[:error] += "An error occurred while trying to display an existing assessment " \
+          "from file directory #{filename}: Name in yaml (#{props['general']['name']}) " \
           "doesn't match #{filename}"
           flash.now[:html_safe] = true
           next
@@ -138,7 +138,7 @@ class AssessmentsController < ApplicationController
       else
         flash.now[:error] = flash.now[:error] ? "#{flash.now[:error]} <br>" : ""
         flash.now[:error] += "An error occurred while trying to display an existing assessment " \
-          "on file directory #{filename}: #{filename}.yml does not exist"
+          "from file directory #{filename}: #{filename}.yml does not exist"
         flash.now[:html_safe] = true
         next
       end
@@ -232,7 +232,7 @@ class AssessmentsController < ApplicationController
     end
 
     params[:assessment_name] = asmt_name
-    params[:isTarImport] = true
+    params[:cleanup_on_failure] = true
     importAssessment && return
   end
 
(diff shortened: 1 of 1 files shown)


app/controllers/assessments_controller.rb
```diff
@@ -121,8 +121,21 @@ def install_assessment
...
-      unless File.exist?(File.join(dir_path, filename, "#{filename}.yml"))
+      # each assessment must have an associated yaml file,
+      # and it must have a name field that matches its filename
+      if File.exist?(File.join(dir_path, filename, "#{filename}.yml"))
+        props = YAML.safe_load(File.open(
+                                 File.join(dir_path, filename, "#{filename}.yml"), "r", &:read
+                               ))
+        unless props["general"] && (props["general"]["name"] == filename)
+          flash.now[:error] = flash.now[:error] ? "#{flash.now[:error]} <br>" : ""
+          flash.now[:error] += "An error occurred while trying to display an existing assessment" \
```
damianhxy, line 132: nit: need space between "assessment" and "on". Perhaps use "from" instead of "on"?

damianhxy: Also could we exclude `name` from the yaml file altogether? Since `name` is already set and this seems like unnecessary work for instructors

This could be a separate PR, though

app/controllers/assessments_controller.rb
```diff
@@ -219,6 +232,7 @@ def importAsmtFromTar
     end
 
     params[:assessment_name] = asmt_name
+    params[:isTarImport] = true
```
damianhxy, line 235: Perhaps it would be more meaningful to call the param `cleanup_on_failure`? i.e. describe the behavior, rather than the scenario

damianhxy (approved):
Followed testing plan successfully, LGTM (other than some comments)

app/controllers/assessments_controller.rb
```diff
@@ -237,8 +252,11 @@ def importAssessment
                        but assessment file name is #{params[:assessment_name]}"
       # destroy model
       destroy_no_redirect
-      # need to delete explicitly b/c the paths don't match
-      FileUtils.rm_rf(assessment_path)
+      # delete files explicitly b/c the paths don't match ONLY if
+      # import was from tarball
+      if isTarImport
+        FileUtils.rm_rf(assessment_path)
```
damianhxy, line 258: nit: simplify to `FileUtils.rm_rf(assessment_path) if isTarImport`

app/controllers/assessments_controller.rb
20wildmanj: Let's save that for a future PR

PR #1985, opened 2023-09-22
Prevent spoofing the author of an annotation

NicholasMy:
### Summary generated by Reviewpad on 22 Sep 23 17:42 UTC
This pull request prevents spoofing the author of an annotation by setting the submitter as the current user's email.

## Description
This very small change overwrites the `submitted_by` parameter on `annotation_params` to the current user's email address to prevent client-side spoofing of the author.

## Motivation and Context
Previously, course assistants could modify the POST request sent by their browser to make annotations look like they were written by somebody else. This update ensures that the annotation's author is the currently authenticated user, ignoring who the client says the author is.

## How Has This Been Tested?
I saved an annotation like normal with the browser dev tools open. I recreated the request and modified it to change the `submitted_by` parameter to be something else. I verified the backend doesn't use my custom author, so the annotation's author is saved as the real user's email address.

## Types of changes

## Checklist:

### DIFF
--- a/app/controllers/annotations_controller.rb
+++ b/app/controllers/annotations_controller.rb
@@ -75,6 +75,8 @@ private
     params[:annotation].delete(:id)
     params[:annotation].delete(:created_at)
     params[:annotation].delete(:updated_at)
+    # Prevent spoofing the submitter
+    params[:annotation][:submitted_by] = @current_user.email
     params.require(:annotation).permit(:filename, :position, :line, :submitted_by,
                                        :comment, :shared_comment, :global_comment, :value,
                                        :problem_id, :submission_id, :coordinate)
--- a/app/assets/javascripts/annotations.js
+++ b/app/assets/javascripts/annotations.js
@@ -510,7 +510,6 @@ function elt(t, a) {
 function createAnnotation() {
   var annObj = {
     filename: fileNameStr,
-    submitted_by: cudEmailStr,
   };
 
   if (currentHeaderPos || currentHeaderPos === 0) {


app/controllers/annotations_controller.rb
```diff
@@ -75,6 +75,8 @@ def annotation_params
     params[:annotation].delete(:id)
     params[:annotation].delete(:created_at)
     params[:annotation].delete(:updated_at)
+    # Prevent spoofing the submitter
+    params[:annotation][:submitted_by] = @current_user.email
```
damianhxy, line 79: If we are overwriting `submitted_by` here, then we should be able to get rid of `annObj.submitted_by` in `annotations.js` https://github.com/autolab/Autolab/blob/14bc1f47c8f5c8682985ae18059cb80b68147ac2/app/assets/javascripts/annotations.js#L513

NicholasMy: Yes, good catch! I've removed that too.

KesterTan (approved):
LGTM

app/controllers/annotations_controller.rb
damianhxy: One last thing, we can get rid of `submitted_by` here too:

https://github.com/autolab/Autolab/blob/717a9153c2debb5d9b12b78547bf50f95655af04/app/controllers/annotations_controller.rb#L78

damianhxy: Nevermind, it's needed since we are setting `:submitted_by` right before this line

PR #1987, opened 2023-09-24
Refactor Assessment name rules, remove config file requirement

### LINKED ISSUE #1393: assessment names and assessment config modules
Autolab allows the creation of assessment names without many restrictions. Unlike course names, the only limit on assessment names is that they cannot begin with numbers. However, the assessment's name is used to form file names and ruby identifiers, so needs to have some more limitations. 
If an assessment with an inappropriate name is created, it triggers an exception whenever anyone tries to view it.
Based on https://stackoverflow.com/questions/3648551/regex-that-matches-valid-ruby-local-variable-names, you should limit assessment names to letters, numbers and _ (keeping the restriction that they cannot begin with numbers). 

It appears that some effort was put in to allow dots (".") in assessment names, which are stripped out before forming identifiers. If you want to keep that feature, then "." should also be allowed (except perhaps in the leading position).

Related to the "." stripping, construct_default_config_file should presumably refer to sanitized_name instead of name, to allow "." to be used when no ruby config file is provided with a newly installed assessment.

Lastly, given that various case conversions are applied to the name, the uniqueness validation for assessment names should presumably be case insensitive.

### LINKED ISSUE #1943: Allow lab directory names to contain safe punctuation (at least: hyphen, underscore)
`>` create a lab directory named "malloc-calibration" (for example)
`>` go to your course's "install assessment" page
`>` see banner: "An error occurred while trying to display an existing assessment on file directory malloc-calibration: assessment file names must only contain lowercase letters and digits with no spaces"
`>` sadface dot gif

I don't see any good reason not to allow '-' and '_' in lab directory names (and maybe also '.')

20wildmanj:
### Summary generated by Reviewpad on 06 Nov 23 22:44 UTC
This pull request includes the following changes:

1. The file `homework02-illegal-assessment-name.tar` has been deleted.
2. In the file `app/views/assessments/_edit_basic.html.erb`, a heading `<h4>Assessment Config</h4>` has been added, the `help_text` of the `f.file_field` has been modified, and a `<div>` containing a link to reload the assessment config has been added. These changes enhance the file by providing additional information and functionality related to the assessment configuration.
3. The file `spec/controllers/assessments_controller_spec.rb` has been updated with test cases for creating assessments, handling invalid display names, and testing the export and import roundtrip feature. Various methods in the `assessments_controller.rb` file have been modified to handle assessment name validation, file cleanup, error handling, and unique names and paths for assessment config files.
4. A new file named `autograde.tar` has been added in the `randomlab` directory. The file is a binary file and its contents differ from the previous version.
5. A new file `randomlab.rb` has been added in the `randomlab` directory. The file defines the module `Randomlab` and includes `AssessmentBase`. It also contains a method `assessmentInitialize` and initializes the instance variable `@problems` with an empty array.
6. A new file `autograde-Makefile` has been added in the `randomlab` directory. The Makefile extracts the contents of the `autograde.tar` file, changes the directory to `random`, and executes a script called `driver.sh`.
7. The file `randomlab/randomlab.rb` has been modified to include a require statement for "AssessmentBase.rb" and define the `Randomlab` module with a new method `assessmentInitialize`. The method calls the `super` method from `AssessmentBase` with two arguments.
8. A new file `driver.sh` has been added in the `randomlab/random` directory. The file is a bash script that generates random numbers and prints them in a JSON format.
9. The file `app/controllers/assessments_controller.rb` has been modified to update error messages, handle file cleanup, load and reload config files, and validate assessment names and special characters.
10. The file `app/models/user.rb` has been modified to remove the `return` keyword before `authentication&.user if authentication&.user` in several methods, ensuring that the methods always return the value of `authentication.user` if it exists.
11. The file `app/helpers/assessment_autograde_core.rb` has been modified to raise an `AutogradeError` for missing problems and handle the requirement of the assessment config file path.
12. The file `app/models/course.rb` has been modified to update the validation syntax for `validates :requested_at` and `validates_associated :oauth_application`.
13. The file `app/controllers/assessment/autograde.rb` has been modified to dynamically require configuration files based on assessment and course names.

These are the main changes across the diff. Let me know if you need further assistance or if there are any specific areas you would like to focus on.

## Description

- Refactor Assessment naming rules so that assessment names can contain hyphens and underscores
  - previously, only lowercase alphabetical characters and numbers were allowed
  - rules now essentially match Ruby Identifier rules, and also allows hyphens
- Remove requirement for config files to be included in importing assessments
   - they now will be automatically generated when no config file is found
 - No longer require module names to match the assessment name during importing
   - automatically rename module identifier to be `"#{sanitized_name}_#{id}".camelize`, so that assessment config files are unique among all assessments
 - Backwards compatibility: add code to migrate existing config files when config files are executed, while preserving the original source file, to prevent assessments from breaking if this commit is rolled back
 - Relax constraints on display name during assessment creation in `assessments/new`
   - previously, the display name here had the same naming rules as assessment names, which is inconsistent with editing display names, where there are no restrictions
 - update and add tests for naming requirements
 - add documentation outlining naming requirements, as well as a guide to troubleshooting importing errors
 
## Motivation and Context

- Address the needs of 15-213 and 15-122, while making naming rules more transparent and make more sense
- Streamline assessment import by removing config file existence and naming requirements

Closes #1943 
Closes #1393 

## How Has This Been Tested?

__This PR requires Autolab to be restarted in order to work__

New assessment:
- Go to a course->"Install Assessment"->"Assessment Builder"
- Create assessments with display names as follows:
  - valid:
     - .hello, ...test, test-test, test_test, TEST1-test, TEST2!, test111!@#$%^&*()
  - invalid:
    - ../,  ---, ____, ////, //---__44,  _-4 
- See that assessment config file is created in `courses/$course_name/$assessment_name/` and that config file has module name of `"#{sanitized_name}_#{id}".camelize`

Existing assessment:
- Ensure that viewing assessment still works
- Run an autograded assessment, see that autograding works (i.e. ensure that for an assessment that returns a non-zero score, it should still return a non-zero score)
- See that in `assessmentConfig` folder, there should exist "course-assessment.rb" and "course-assessment-${id}.rb" files, and "course-assessment-${id}.rb"  should have a module name of `"${assessment name}${id}" `

Import Assessment:
- from the documentation, those with valid names should be created successfully, those with invalid names should fail to create (try also names such as "../", see that creation fails)
- tar import: Import the tar file below, see that it successfully imports, generates a new assessment config file in the assessment folder and it also generated in the assessmentConfig folder
[bareminimum.tar.zip](https://github.com/autolab/Autolab/files/12709686/bareminimum.tar.zip)
- Try importing the tar file below, see that it fails to import
[badasmt.tar.zip](https://github.com/autolab/Autolab/files/12709708/badasmt.tar.zip)
Screenshot 2023-09-24 at 5 23 51 PM
- Take an existing assessment, copy the folder, rename the folder / `assessment.yml` / `name` field in `assessment.yml` to a name that is valid, see that it imports successfully
- Take an existing assessment, copy the folder, rename the folder / `assessment.yml` / `name` field in `assessment.yml` to a name that is invalid, see that it cannot be imported
Screenshot 2023-09-24 at 5 29 26 PM

- run docs locally, see that documentation looks good

## Types of changes

## Checklist:

- [ x I have run rubocop for style check. If you haven't, run `overcommit --install && overcommit --sign` to use pre-commit hook for linting


app/controllers/assessments_controller.rb
```diff
@@ -111,12 +111,14 @@ def install_assessment
...
         # add line break if adding to existing error message
         flash.now[:error] = flash.now[:error] ? "#{flash.now[:error]} <br>" : ""
         flash.now[:error] += "An error occurred while trying to display an existing assessment " \
-            "from file directory #{filename}: assessment file names must only contain lowercase " \
-            "letters and digits with no spaces"
+            "from file directory #{filename}: Invalid assessment name. "\
+            "Find more information on valid assessment names "\
+            '<a href="https://docs.autolabproject.com/lab/#assessment-naming-rules">here</a><br>'
+        flash[:html_safe] = true
         flash.now[:html_safe] = true
```
KesterTan, line 122: Only need one of the two, in this case since there's a redirect to install assessment, maybe just flash and remove flash.now.

app/models/user.rb
```diff
@@ -84,19 +84,19 @@ def ldap_reset
   def self.find_for_facebook_oauth(auth, _signed_in_resource = nil)
     authentication = Authentication.find_by(provider: auth.provider,
                                             uid: auth.uid)
-    return authentication.user if authentication&.user
+    authentication.user if authentication&.user
```
20wildmanj, line 87: rubocop changes

app/helpers/assessment_autograde_core.rb
```diff
@@ -450,7 +450,7 @@ def saveAutograde(submissions, feedback)
 
         scores.keys.each do |key|
           problem = @assessment.problems.find_by(name: key)
-          raise AutogradeError.new("Problem \"" + key + "\" not found.") unless problem
+          raise AutogradeError, "Problem \"" + key + "\" not found." unless problem
```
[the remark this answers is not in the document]
20wildmanj: not necessary

damianhxy:
Unable to create `TEST-test` and `TEST!` (probably because of `...test` and `test-test` ?)
Screenshot 2023-10-29 at 20 51 54
Screenshot 2023-10-29 at 20 51 40

app/controllers/assessments_controller.rb
```diff
@@ -290,20 +291,26 @@ def importAssessment
...
+      # We just want to prevent file traversal attacks here, and stop names that break routing
+      # first regex - try to sanitize input, allow special characters in display name but not name
+      # if the sanitized doesn't match the required identifier structure, then we reject
+      sanitized_display_name = @assessment.display_name.gsub(/[!@#$%^&*(),.?":{}|<>\s]/, "")
+      Rails.logger.debug(sanitized_display_name)
+      if sanitized_display_name !~ /^[A-Za-z_][A-Za-z][A-Za-z0-9_-]*$/
+        # TODO: update with docs link
         flash[:error] =
-          "Assessment name is blank or contains characters that are not lowercase letters or digits"
+          "Assessment name is blank or contains disallowed characters. Find more information on"\
```
damianhxy, line 304: Nit: Add space between "on" and "valid"

damianhxy:
> See that in assessmentConfig folder, there should exist "course-assessment.rb" and "course-assessment-${id}.rb" files, and "course-assessment-${id}.rb" should have a module name of "${assessment name}-${id}.rb"

I suppose you meant module name of "${assessment name}${id}"?

app/controllers/assessments_controller.rb
```diff
@@ -290,20 +291,26 @@ def importAssessment
...
-      # Validate the name
-      ass_name = @assessment.display_name.downcase.gsub(/[^a-z0-9]/, "")
-
-      if ass_name.blank?
+      # Validate the name, very similar to valid Ruby identifiers, but also allowing hyphens
+      # We just want to prevent file traversal attacks here, and stop names that break routing
+      # first regex - try to sanitize input, allow special characters in display name but not name
+      # if the sanitized doesn't match the required identifier structure, then we reject
+      sanitized_display_name = @assessment.display_name.gsub(/[!@#$%^&*(),.?":{}|<>\s]/, "")
+      Rails.logger.debug(sanitized_display_name)
```
damianhxy, line 300: Nit: remove this debug line, and the todo comment below.

Also, it does not let me create an assessment with name `_4utol4b-_4smt` (as per docs)

damianhxy: It allows me to create `.autolab.asmt` which is supposed to be invalid

app/controllers/assessments_controller.rb
```diff
@@ -111,12 +111,13 @@ def install_assessment
                                          filename)) || (filename == "..") || (filename == ".")
 
       # assessment names must be only lowercase letters and digits
-      if filename =~ /[^a-z0-9]/
+      if filename !~ /^[A-Za-z_][A-Za-z][A-Za-z0-9_-]*$/
```
damianhxy, line 114: Screenshot 2023-10-29 at 21 19 12

Regex is disallowing names that should be valid

damianhxy:
When I (attempt to) create the following assessments, in this order:
- autolabAsmt
- Autolab-Asmt
- Autolab_Asmt10
- .autolab.asmt

The last one fails because autolabAsmt is taken, but somehow it seems to overwrite the yml, so now it is invalid
Screenshot 2023-10-29 at 21 25 26

damianhxy:
Importing bareminimum (hellortar.tar), fails with `Error loading yaml: undefined method `key?' for nil:NilClass `

20wildmanj:
> Unable to create `TEST-test` and `TEST!` when following PR instructions (probably because of `...test` and `test-test` ?) Screenshot 2023-10-29 at 20 51 54 Screenshot 2023-10-29 at 20 51 40

So, upon further testing, each assessment name is valid, but saving those ones error because in the database, the names are not case sensitive, so validation fails. However, if you were to install each one in a different course, they would be successfully installed

20wildmanj:
> > See that in assessmentConfig folder, there should exist "course-assessment.rb" and "course-assessment-${id}.rb" files, and "course-assessment-${id}.rb" should have a module name of "${assessment name}-${id}.rb"
> 
> I suppose you meant module name of "${assessment name}${id}"?

Yes

20wildmanj:
> Importing bareminimum (hellortar.tar), fails with `` Error loading yaml: undefined method `key?' for nil:NilClass ``

Seems like the tar was wrong, this one should work:
[baremin 2.tar.zip](https://github.com/autolab/Autolab/files/13222690/baremin.2.tar.zip)

20wildmanj:
> When I (attempt to) create the following assessments, in this order:
> 
> * autolabAsmt
> * Autolab-Asmt
> * Autolab_Asmt10
> * .autolab.asmt
> 
> The last one fails because autolabAsmt is taken, but somehow it seems to overwrite the yml, so now it is invalid Screenshot 2023-10-29 at 21 25 26

the behavior of this has now changed due to some changes I just pushed, ".autolab.asmt" will now get converted to a name of "autolab," so I can't replicate this behavior.

app/controllers/assessments_controller.rb
20wildmanj: the behavior here has been changed to rather than just try to substitute out bad characters, take out a portion of the display name that can be used as a valid name. so now `.autolab.asmt` will become `autolab`. Does this sound reasonable?

app/helpers/assessment_autograde_core.rb
```diff
@@ -502,7 +502,19 @@
...
+    begin
+      if @assessment.use_unique_module_name
+        require assessment.unique_config_file_path
+      else
+        require assessment.config_file_path
+      end
+    rescue TypeError => e
+      raise AutogradeError, "could not find the assessment config file: #{e}"
+    rescue LoadError => e
+      raise AutogradeError, "could not load the assessment config file: #{e}"
```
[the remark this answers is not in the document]
20wildmanj: is a non issue because these errors get handled by assessments controller, which will only show errors to instructors and above

app/helpers/assessment_autograde_core.rb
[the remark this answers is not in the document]
20wildmanj: Rubocop recommends the way I am creating exceptions.

KesterTan:
LGTM, was able to follow the steps outlined

damianhxy:
> See that assessment config file is created in courses/$course_name/$assessment_name/ and that config file has module name of "#{sanitized_name}_#{id}".camelize

The module name for config files under that path appear to be "#{sanitized_name}".camelize
Screenshot 2023-11-03 at 23 40 36

damianhxy:
Installing `autolabAsmt` followed by `Autolab-Asmt` results in the first assessment's rb file becoming a `.bak` file

Screenshot 2023-11-03 at 23 47 16

damianhxy:
None of the "invalid assessment names" appear to be actually invalid.

Screenshot 2023-11-03 at 23 48 55
Screenshot 2023-11-03 at 23 48 14
Screenshot 2023-11-03 at 23 48 23

app/models/assessment.rb
```diff
@@ -46,7 +46,8 @@ class Assessment < ApplicationRecord
   # Constants
   ORDERING = "due_at ASC, name ASC".freeze
   RELEASED = "start_at < ?".freeze
-
+  VALID_NAME_REGEX = /^[A-Za-z][A-Za-z0-9_-]*$/.freeze
```
damianhxy, line 49: Ruby identifiers can start with underscores, should we allow that?

app/controllers/assessments_controller.rb
```diff
@@ -262,6 +263,7 @@ def importAssessment
       @assessment.load_yaml # this will save the assessment
     rescue StandardError => e
       flash[:error] = "Error loading yaml: #{e}"
+      Rails.logger.debug e.backtrace
```
damianhxy, line 266: Nit: remove debugging statements (also in assessment.rb)

app/models/assessment.rb
```diff
@@ -46,7 +46,8 @@ class Assessment < ApplicationRecord
...
+  VALID_NAME_SANITIZER_REGEX = /^[^A-Za-z]*([^A-Za-z]*[A-Za-z0-9_-]+)/.freeze
```
damianhxy, line 50: Nit: the second `[^A-Za-z]*` seems redundant, could simplify to `/^[^A-Za-z]*([A-Za-z0-9_-]+)/`

app/models/assessment.rb
```diff
@@ -396,6 +434,10 @@ def source_config_module_name
     sanitized_name.camelize
   end
 
+  def unique_source_config_module_name
```
damianhxy, line 437: Nit: appears unused

damianhxy (approved):
Generally LGTM, other than some bugs and nits noted

app/controllers/assessments_controller.rb
damianhxy: Makes sense

20wildmanj:
> > See that assessment config file is created in courses/$course_name/$assessment_name/ and that config file has module name of "#{sanitized_name}_#{id}".camelize
> 
> The module name for config files under that path appear to be "#{sanitized_name}".camelize Edit: This seems to only happen for creation from scratch
> 
> Screenshot 2023-11-03 at 23 40 36

This is probably because for new assessments there doesn't exist an id yet? Regardless, the actual config file that gets written to `assessmentConfig` wiill have `"#{sanitized_name}#{id}".camelize` (I removed the underscore since it gets removed by camelize anyways, so this is a non-issue

app/models/assessment.rb
20wildmanj: I left out underscores because in the current regex that would allow for all underscores to be valid. Essentially, it makes the validation more complicated so rather not allow it.

damianhxy:
LGTM

Issue #931, opened 2017-11-04
Leading whitespace in user's first name breaks extension manager for that particular user.


pcbouman-eur:
Recently we had a case where we were unable to provide an extension to a student who has a first name field in the database containing a leading space (e.g. " P. Bouman"). While the user shows up in the user-search field in the Extension Manager, an attempt to create the extension results in a "No student with id was found for this course." error message. I was able to provide an extension by manually inserting the relevant records into the database, after which the extension did show up in the extension manager.

The problem turned out to be reproducible for all students who have a leading whitespace in their first name field. I recorded the POST data submitted by the browser for such a student, an noticed the "extension[course_user_datum_id]" part of the form data is empty. This suggest something goes wrong in creating the POST request in the browser for such students.

cg2v:
This also seems to occur if there is a character that's encoded into an HTML entity (like ') in the name

jbman223:
The root cause of this issue is an issue with how we attach the auto-filled form's input to a student ID. The form input is not HTML safe/encoded while the dictionary relating user ids to user emails is. The fix for this is to base64 encode the raw user values in the array (so that characters like `<` or `'` are maintained) then when searching this dictionary with the value of the student box, also base64 encode that value. This keeps everything safe from script injections.

A separate fix is needed for leading and trailing spaces in student names. Either the HTML safe encoder that rails uses or the form value strips leading and trailing white space on strings, so we must also strip the student name before we encode it in the dictionary.

PR #2008, opened 2023-11-13
No line breaks when generating base64 strings

### LINKED ISSUE #2007: Autocomplete fails for long names + email adresses
**Your environment**
Does not matter.

**Steps To Reproduce**
1. Create a user with a very long name and email address e.g. "Christopher Nathaniel Devereaux (autolab@campus.tu-berlin.de)".
2. Go to "Manage Course" -> "Act as user" -> Type <the name of the user> -> Submit.

**Current behavior**
"User does not exist in the course" flashes.

**Expected behavior**
User impersonation.

ugogon:
The function `strToBase64` introduces a line break (I think after every 76 character). This is incompatible with 
rails `Base64.strict_encode64` where no line breaks are introduced. This leads to the bug that the user names and email are converted to different hashes in ruby and javascript.

This PR just aligns the javascript `strToBase64` implementation with the rails `Base64.strict_encode64` implementation. _Although I still do not get why this code duplication is necessary._ 

### Summary generated by Reviewpad on 13 Nov 23 22:44 UTC
This pull request fixes a bug related to long strings in the base64.js file. The patch modifies the base64EncArr function, specifically commenting out code that adds newlines to the encoded string, since it was causing an issue.

## Description
I just commented out the lines where the line breaks are added. I think that javascript code was taken from somewhere else (as these line breaks are neither compatible with `Base64.encode64` (after 60 characters) nor `Base64.strict_encode64` (no linebreaks)) and I do not want to diverge too much from the original code.

## Motivation and Context
Fixes #2007.

## How Has This Been Tested?
Manual, see steps to reproduce in #2007.

## Types of changes


damianhxy:
Hi @ugogon, thanks for the PR!

The implementation was taken from https://developer.mozilla.org/en-US/docs/Glossary/Base64, but it was recently updated to a much simpler implementation in https://github.com/mdn/content/pull/25253.

If possible, it would be great if you could replace the implementation with MDN's new implementation that uses `TextEncoder`.

As for why we use base64, there is a long history behind it (starting with #931)

ugogon:
Thanks. I updated it, although I'm not sure if a separate file is still necessary.

damianhxy:
Either way is fine, you can leave it as a separate file for now in case another piece of code needs to use it

damianhxy (approved):
Tested that long names now work, and unicode names still work (e.g. chinese)

PR #2018, opened 2023-12-01
Overwrite labs on reupload

### LINKED ISSUE #2017: Overwrite labs on reupload

jlge:
### Summary generated by Reviewpad on 29 Jan 24 17:53 UTC
This pull request modifies the `assessments_controller.rb`, `assessment.rb`, `course_logger.rb`, `install_assessment.html.erb`, and `assessments_controller_spec.rb` files.

In `assessments_controller.rb`, the changes include:
- Replacing a check for an existing assessment with a variable assignment.
- Adding a condition to skip certain files based on the existing assessment.
- Adding params for assessment overwrite.

In `assessment.rb`, the changes include:
- Adding methods `asmt_yaml_path`, `log_path`, `unique_source_config_file_path`.
- Modifying methods `dump_yaml` and `load_yaml` to use the new methods.
- Modifying method `path` to handle the new methods.

In `course_logger.rb`, the changes include:
- Modifying method `setAssessment` to use the assessment's log path.

In `install_assessment.html.erb`, the changes include:
- Adding a paragraph explaining the behavior of importing a tarball with the same name as an existing assessment.

In `assessments_controller_spec.rb`, the changes include:
- Modifying the description of a test case.

Overall, these changes seem to improve the management of existing assessments and handle the import of assessment tarballs with the same name.

## Description

- Edit `importAsmtFromTar` to allow duplicate assessment names and upload files for an existing assessment
  - Change `importAssessment` logic to not load any config files if the assessment already exists

## Motivation and Context

Resolves #2017 

## How Has This Been Tested?

Download and unzip these files:
[homework3.tar](https://github.com/autolab/Autolab/files/13614047/homework3.zip)
[homework3_new.tar](https://github.com/autolab/Autolab/files/13614045/homework3_new.zip)

#### Creating your own tar files: ####
1. Export an existing assessment, to get the corresponding tar file, e.g. `homework3.tar`
2. Copy the tar and add another file to the directory, name it smth else like `homework3_new.tar`

#### How to test ####
1. Go to 'Install Assessment' and upload `homework3_new.tar` (don't delete the original asmt)
2. Check that the new file you added to the directory was added and nothing else changed (such as handins/student data)
3. Go to 'Edit assessment' and change the writeup to the new file you added , e.g. `meme.PNG`
4. Save the changes and click 'View writeup', it should download the new file that was uploaded 👍 
5. Go to 'Install Assessment' and upload 'homework3.tar' 
6. Check that the writeup still downloads (ensures that uploading does not delete existing files)

## Types of changes

## Checklist:


app/controllers/assessments_controller.rb
```diff
@@ -202,6 +200,10 @@ def importAsmtFromTar
           FileUtils.chmod entry.header.mode, entry_file,
                           verbose: false
         elsif entry.file?
+          # Skip config files
+          next if overwrite && %W[#{asmt_name}.yml
```
damianhxy, line 204: Need to skip handin directory and file within

app/controllers/assessments_controller.rb
```diff
@@ -194,6 +192,7 @@ def importAsmtFromTar
         # Ensure file will lie within course, otherwise skip
         # Allow equality for the main directory to be created
         next unless Archive.in_dir?(Pathname(entry_file), Pathname(assessment_path), strict: false)
+        next if overwrite && relative_pathname.start_with?('handin/')
```
damianhxy, line 195: You should use the value of `handin_directory`, not hardcode it to `handin`

app/controllers/assessments_controller.rb
```diff
@@ -185,6 +180,7 @@ def importAsmtFromTar
         # Ensure file will lie within course, otherwise skip
         # Allow equality for the main directory to be created
         next unless Archive.in_dir?(Pathname(entry_file), Pathname(assessment_path), strict: false)
+        next if !existing_asmt.nil? && relative_pathname.start_with?(existing_asmt.handin_directory)
```
damianhxy, line 183: This check doesn't actually work, because `relative_pathname` is prepended by the assessment's name, e.g. `homework3/handin/`

damianhxy: I would suggest replacing the check with something like

```ruby
next if existing_asmt && Archive.in_dir?(Pathname(entry_file), existing_asmt.handin_directory_path, strict: false)
```

damianhxy (changes requested):

app/controllers/assessments_controller.rb
damianhxy: Verify that the check works by adding a file to the handin directory, and upload a tar where that file is replaced with something else -> shouldn't overwrite

app/controllers/assessments_controller.rb
```diff
@@ -193,6 +190,11 @@ def importAsmtFromTar
           FileUtils.chmod entry.header.mode, entry_file,
                           verbose: false
         elsif entry.file?
+          # Skip config files
+          next if existing_asmt && %W[#{asmt_name}.yml
```
damianhxy, line 194: Nit: this seems a bit clunky since we won't upload files that have a matching basename but at a different path. e.g. `newdir/{asmt_name}.rb`

suggest comparing against `existing_asmt.path { ... filename ... }`, e.g. `existing_asmt.path "#{asmt_name}.yml"`. Even better to define helper methods in `assessment.rb` to return these paths (just like `handout_path`, `handin_directory_path`, `writeup_path`).

Note:
- yml: `dump_yaml` and `load_yaml` both use `path("#{name}.yml")` directly. Consider extracting into a helper method with definition `path "#{name}.yml"`
- rb: there exists `existing_asmt.unique_source_config_file_path`, but its definition could be refactored to `path "#{name}.rb"`
- log: currently `course_logger.rb` uses `Rails.root.join("courses", @course.name, @assessment.name, "log.txt")`. Consider extracting into a helper method with definition `path "log.txt"`

damianhxy (approved):
Verified that new files get uploaded, and asmt.rb / asmt.yml / log.txt / handin dir (+ contents) are not overwritten

Left some nits

PR #2058, opened 2024-01-22
Remove redundant conditional to migrate config file, Allow empty config file to be loaded

### LINKED ISSUE #2051: Better handling for malformed assessment config file
If the config file is malformed, we get an exception due to the following code where `match` is `nil`.
```ruby
if config_source !~ /\b#{unique_config_module_name}\b/
  match = config_source.match(/module\s+(\w+)/)
  config_source = config_source.sub(match[0], "module #{unique_config_module_name}")
end
```

Screenshot 2024-01-15 at 14 17 39

We should consider displaying a more meaningful error to the user and/or backing up the file + generating a default config.

Furthermore, some methods that call `extend_config_module` (`handout`, `assesssments#show`) do not catch errors that may arise from calling `config_module` -> `load_config_file`.

### LINKED ISSUE #2052: Impossible conditional in load_config_file
The conditional
```ruby
if (File.exist? source_config_file_path) &&
   (source_config_file_path != unique_source_config_file_path) &&
   Archive.in_dir?(source_config_file_path, folder_path)
```

is impossible to satisfy. The second implies `sanitized_name != name` while the third implies `sanitized_name == name`.

```ruby
def source_config_file_path
  Rails.root.join("courses", course.name, sanitized_name, "#{sanitized_name}.rb")
end

# name is already sanitized during the creation process
def unique_source_config_file_path
  Rails.root.join("courses", course.name, name, "#{name}.rb")
end

def folder_path
  Rails.root.join("courses", course.name, name)
end
```

The intention was to migrate existing config files from the old path. However, the issue was that the old path might equal the new path of another assessment, causing the config file to be accidentally "backed up".

Perhaps we shouldn't attempt to migrate at all.

20wildmanj:
### Summary generated by Reviewpad on 22 Jan 24 18:37 UTC
This pull request includes two patches. The first patch removes a redundant conditional and fixes an issue with an empty config source file in the `assessments_controller.rb` file. The second patch adds error handling to the `handout#handout` method in the `assessment/handout.rb` file.

## Description

- Remove a conditional which was originally meant to migrate over old config modules over to the new naming system, due to differences in allowed assessment names / sanitization methods. However, in practice this check is not necessary.
- Add code to check if the regex match for a `module something` returned nil in the case that the provided configuration file did not have a module. If so, build a default config file and load that instead, while maintaining the source config file and backing up previous config file, as per previous behavior
 - Change Rescue to catch `StandardError` instead of just `AutogradeError` so that errors such as a malformed config file get caught as well
 
## Motivation and Context

- #1987 introduced changes to config modules which introduced breaking behavior on autolab production. Previously, empty configuration files were acceptable and able to be used in assessments, but with the new changes, such a file causes an error upon loading the assessment.

Closes #2051
Closes #2052

## How Has This Been Tested?

- (not necessary to recreate) Rebuilt autolab docker using commit 3676b13 (commit before #1987)
- Created Autopopulated course / assessments using autopopulator, to have courses created pre-1987
- switched to 1987 commit and checked that migration of assessments still worked

Necessary testing:
- imported an assessment with a config file different from default, with a module
  - verified that this file was loaded into `assessmentConfig` folder properly
- imported an assessment without a config file, verified it loaded
- imported an assessment with a config file but with nothing inside, verified that default config was loaded into `assessmentConfig`
- for an existing assessment, first changed source config file to include something other than default, verified that change was loaded properly
  - changed source config file to not have a module inside it, verified that default config file was loaded into `assessmentConfig`
 - directly modified config file in `assessmentConfig` to not have a module, saw that error is now caught
 
Screenshot 2024-01-21 at 9 05 04 PM

## Types of changes

## Checklist:

### DIFF
--- a/app/controllers/assessment/handout.rb
+++ b/app/controllers/assessment/handout.rb
@@ -14,7 +14,21 @@ module AssessmentHandout
 
   def handout
     # If the logic here changes, do update assessment#has_handout?
-    extend_config_module(@assessment, nil, @cud)
+    begin
+      extend_config_module(@assessment, nil, @cud)
+    rescue StandardError => e
+      if @cud.has_auth_level? :instructor
+        flash[:error] = "Error loading the config file: "
+        flash[:error] += e.message
+        flash[:error] += "<br/> Try reloading the course config file," \
+          " or re-upload the course config file in order to recover your assessment."
+        flash[:html_safe] = true
+      else
+        flash[:error] = "Error loading #{@assessment.display_name}. Please contact your instructor."
+      end
+      return
+    end
+
 
     if @assessment.overwrites_method?(:handout)
       hash = @assessment.config_module.handout
--- a/app/controllers/assessments_controller.rb
+++ b/app/controllers/assessments_controller.rb
@@ -514,7 +514,7 @@ class AssessmentsController < ApplicationController
     set_handin
     begin
       extend_config_module(@assessment, @submission, @cud)
-    rescue AutogradeError => e
+    rescue StandardError => e
       if @cud.has_auth_level? :instructor
         flash[:error] = "Error loading the config file: "
         flash[:error] += e.message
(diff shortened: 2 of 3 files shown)


evanyeyeye (approved):
LGTM King

PR #2060, opened 2024-01-25
Feature/GitHub personal access token set

brandonfranzke:
Add minimal (PoC) "update github" token with text field.  Allow user/student to manually enter an override AFTER the OAuth flow completes.  Useful for user controlled access tokens, potentially at a "per assignment" level.

## Summary
New button on user page (if Github integaration): "update github token".  Single field form.  Overwrite (ActiveRecord encrpting) existing token.  These appear to be universal across the API and do nor require individual attention.

Incomplete tests.  I am teaching an overload this term and already behind.  Not able to build out the spec for the submission.  Happy to assist others but I cannot be lead.


brandonfranzke:
I have a fork with updates to the Dockerfile.  Did not intend to commit those, only updates in /app and to routes.

I am going to close this with comment in case there is discussion and prepare a better merge source for the weekend.

damianhxy:
Does #2061 supersede this?

PR #2070, opened 2024-02-03
Update tests to use instance variables, add course/assessment teardown code

20wildmanj:
## Summary
reviewpad:summary

## Description

- Add code to teardown course files and assessment config files, since 100s of these files would be generated during testing and stay there, which was not ideal
- Refactor some testing code to use instance variable defined in "controllers shared context" and defined in the factory sets, to make code shorter / less complex
- Write a feature test to create a new assessment and edit fields, as well as a test to create an annotation for a student
- favicon.ico is needed in public/ or else rails throws a tantrum
## Motivation and Context

- updates tests to be easier to use, and further increases feature coverage

## How Has This Been Tested?

- run `bundle exec rails spec`, see that tests pass

## Checklist:


spec/controllers/assessments_controller_spec.rb
```diff
@@ -1,5 +1,7 @@
 require "rails_helper"
 include ControllerMacros
+require_relative "controllers_shared_context"
```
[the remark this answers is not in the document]
20wildmanj: are you sure about this

spec/features/manage_submissions_spec.rb
```diff
@@ -1,19 +1,18 @@
 require "rails_helper"
 require_relative("../support/controller_macros")
+require_relative("../controllers/controllers_shared_context")
 include ControllerMacros
 
 RSpec.describe "manage submissions user flow", type: :feature do
   describe "click button", js: true do
     context "when user is Instructor" do
+      include_context "controllers shared context"
```
jlge, line 9: nit: some inconsistency with whether `include_context` is inside the context block or the outer describe block, don't know if we want to standardize it or not

spec/models/annotation_spec.rb
```diff
@@ -8,6 +8,9 @@
       [CourseUserDatum.where(user: @students.first).first,
        CourseUserDatum.where(user: @students.second).first]
     end
+    after(:each) do
```
jlge, line 11: nit: why not just use the shared context here?

jlge (approved):
LGTM, ran tests and they passed (but I didn't test Capybara b/c I don't have it set up)

spec/models/annotation_spec.rb
20wildmanj: It's because the shared contexts exists for controllers (and lives within spec/controllers), we could do the same for models

spec/features/manage_submissions_spec.rb
20wildmanj: It kind of depends on whether or not you want it defined within a certain scope, whether you want it defined across all contexts in a describe block or within a certain context.

PR #2071, opened 2024-02-03
Use instance variables instead of flash to display duplicates during roster import

### LINKED ISSUE #906: Roster import should not fail on duplicate email addresses
Any fatal duplicates should be caught and presented to the user for cherry-picking. I described this more in #581.

damianhxy:
## Description

Use instance variables instead of flash to properly render roster import errors with large numbers of duplicates.

## Motivation and Context

Currently, roster errors are propagated to the view using a flash. However, this results in a cookie overflow error when there are too many duplicates. By using instance variables, we can avoid this problem.

Closes #906

## How Has This Been Tested?

[roster_duplicates.csv](https://github.com/autolab/Autolab/files/14154368/roster_duplicates.csv)
[roster_noduplicates.csv](https://github.com/autolab/Autolab/files/14154369/roster_noduplicates.csv)

**Before PR - duplicates**
Screenshot 2024-02-03 at 16 52 37

**After PR - duplicates**

Screenshot 2024-02-03 at 16 53 35

Also check that uploading rosters with no errors (`roster_noduplicates.csv`) works as expected.

## Types of changes

## Checklist:


app/views/courses/upload_roster.html.erb
```diff
@@ -1,5 +1,11 @@
+<% if @roster_error %>
+<div class="error" id="flash_error">
```
jlge, line 2: nit: visually maybe make it so that when you hover over a row it doesn't highlight? I feel like it isn't really necessary

jlge (approved):
LGTM 👍

app/views/courses/upload_roster.html.erb
damianhxy: Discussed offline as won't fix

PR #2081, opened 2024-02-10
Added a button to release/unrelease grades

### LINKED ISSUE #2063: Allow release of individual student scores
Currently, scores can be released via
- "release all scores" button
- "release section scores" button
- individual student/problem via gradesheet (clicking on the grid and selecting released)

However,
- might not want to wait for instructor to release all at once (if graders take different amount of times to grade)
- sections might not be properly configured (incorrect roster import format, etc.)

So it would be good to allow TAs to release a student's scores (for all the problems) easily. For example, on the speedgrader screen.

dwang3851:
## Description
- Added a button to the view source page to release grades
- Added an indicator next to the grades tabs to show if a problem's score is released or unreleased
- If all of the scores for a submission are released, there is a unrelease grades button

Screenshots:
image
image

## Motivation and Context
-Currently, scores can only be released all at once, or through the individual problems on a gradesheet
- There is no easy way to release grades for a single student
Resolves #2063 

## How Has This Been Tested?
- I tested in my local environment (autolab manual installation) with a student, TA, and instructor account
- Checked that none of the changes are observable from a student account
- Ensured both TA and course instructors are able to use the release grades button

## Types of changes

## Checklist:


db/schema.rb
```diff
@@ -418,4 +418,4 @@
     t.index ["risk_condition_id"], name: "index_watchlist_instances_on_risk_condition_id"
   end
 
-end
+end
```
jlge, line 421: nit: add newline

app/views/submissions/_release_grades.html.erb
```diff
@@ -0,0 +1,13 @@
+<span id="release-grades">
+  <% unless released(@assessment, @submission) %>
+    <%= link_to(releaseSubmission_course_assessment_submission_path(@course, @assessment, @submission),
+                class: "btn small", title: "Release Grades") do %>
+    Release Grades
```
jlge, line 5: I feel like calling the button "Release Grades" could be confusing for the instructor b/c it sounds like they are releasing the grades for all students. Possible alternative could be "Release grade" but idk if that captures what we want either

app/controllers/submissions_controller.rb
```diff
@@ -611,6 +612,28 @@ def view
     end
   end
 
+  action_auth_level :releaseSubmission, :course_assistant
+  def releaseSubmission
```
jlge, line 616: Maybe call this `release_student_grade` to be more descriptive, also use snake case to be more consistent (I know there are other endpoints with camel casing but I think snake case is better)

app/controllers/submissions_controller.rb
```diff
@@ -611,6 +612,28 @@ def view
...
+    @assessment.problems.each_with_index do |p, _i|
+      p_score = p_scores[p.id]
+      p_score&.released = true
+      p_score.save
+    end
+    redirect_back(fallback_location: root_path)
+  end
+
+  action_auth_level :unreleaseSubmission, :course_assistant
+  def unreleaseSubmission
```
jlge, line 627: Same thing here with naming

app/views/submissions/_release_grades.html.erb
jlge: I think it would be helpful to add a title (which shows when you hover over the button) to specify that it is only releasing the grade for the current student

app/controllers/submissions_controller.rb
```diff
@@ -611,6 +612,28 @@ def view
...
+    p_scores = @submission.problems_to_scores
+    @assessment.problems.each_with_index do |p, _i|
```
jlge, line 629: Is `_i` a typo? Should probably just be `_`

jlge (changes requested):
Testing locally works, just some notes about naming and style. Overall a slay ✨

jlge (approved):
LGTM
image

app/helpers/submission_helper.rb
```diff
@@ -6,4 +6,16 @@ def plus_fix(f)
       sprintf("%.2f", f.round(2))
     end
   end
+
+  def released(assessment, submission)
```
damianhxy, line 10: Easier/better to do `submission.scores.pluck(:released).all?`

So that it's considered released as long as all problems with scores are released.

e.g. the button will show "unreleased grade" despite the nil manual score

Screenshot 2024-02-19 at 22 01 58

app/views/submissions/_grades.html.erb
```diff
@@ -11,6 +11,13 @@
           <b> &ndash; </b>
         <% end %>
         <b> / <%= p.max_score ? sprintf("%.2f", p.max_score.round(2)) : raw("&ndash;") %></b>
+        <b>
+          <% if p_score&.released %>
```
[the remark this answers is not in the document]
damianhxy: Also, it wouldn't really make sense to show this message to students. Since they see a dash if it's unreleased, and a score otherwise.

i.e. only show the message if `@cud.instructor? or @cud.course_assistant?`

damianhxy (changes requested):

app/helpers/submission_helper.rb
damianhxy: But note that this would become pretty similar to `grades_released?` in `submission.rb`. Maybe put the methods together?

```ruby
  def problems_released?
    scores.pluck(:released).all?
  end

  def grades_released?(as_seen_by)
    include_unreleased = as_seen_by.course_assistant? || as_seen_by.instructor?
    problems_released? || include_unreleased
  end
```

damianhxy:
LGTM, other than that the buttons do not update when annotations are created (e.g. a nil score is now non-nil and unreleased).

app/views/submissions/view.js.erb
```diff
@@ -4,7 +4,8 @@ newFile = {
     symbolTree: '<%= j render("submissions/code_symbol_tree", ctag_obj: @ctag_obj) %>',
     url: '<%= url_for(params.permit(:header_position, :action, :controller).to_h) %>',
     versionLinks: '<%= j render("version_links") %>',
-    versionDropdown: '<%= j render("version_dropdown") %>'
+    versionDropdown: '<%= j render("version_dropdown") %>',
+    releaseGrades: '<%= j render("release_grades") %>'
```
damianhxy, line 8: Should not do any caching on a per-file basis, as the "release/unrelease grades" button applies to the submission as a whole.

damianhxy (approved):
LGTM

PR #2082, opened 2024-02-10
Remove custom pluralize method in application_controller.rb

20wildmanj:
## Description

- Remove custom pluralize method as there already exists a [method provided by rails](https://apidock.com/rails/ActionView/Helpers/TextHelper/pluralize) which is being used by views code. 
- The pluralize code doesn't even work with controllers (where it was being used, such as in `autograde.rb`),as it only handles strings but was being fed integers which caused error such as this when trying to run `regradeAll`

Screenshot 2024-02-10 at 2 42 03 PM

## Motivation and Context

- the pluralize code was throwing an error unrelated to regradeAll which was not caught

## How Has This Been Tested?

- Run `regradeAll`, see that it runs successfully and pluralizes students
Screenshot 2024-02-10 at 2 47 52 PM

## Types of changes

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -345,11 +345,6 @@ protected
   end
   ### END HELPERS
 
-  def pluralize(count, singular, plural = nil)
-    "#{count || 0} " +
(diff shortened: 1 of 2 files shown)


jlge (approved):
LGTM
image

PR #2089, opened 2024-02-13
Allow instructors to select favorite problems to prioritize when annotating assignments and grading

### LINKED ISSUE #2062: Allow instructors to define a default annotation problem
Currently, the default annotation problem is whichever comes first alphabetically.

It would be convenient to allow instructors to define a default problem so that e.g. they can default to "style" instead of "correctness deduction".

jhs-panda:
## Description

- Added star icon in "Problems" in front of each problem to denote if problem is starred or not. Toggling buttons will automatically reflect when making annotations.
- Added checkbox in the Edit Assessments > Problems > (Edit Problem under Options) to select if problem should be listed first when adding annotations in grading
- When adding annotations in grading, starred problems are listed first, then unstarred. Both types of problems are listed alphabetically within their respective categories. Two groups of problems are separated.

Screenshots:
Added stars in front of each problem to denote if problem is currently starred or not:
Screen Shot 2024-03-09 at 4 18 20 PM
Toggle star/unstar options for each problem on Edit Assessment page (no need to press "Save" button for changes to reflect):
Screen Shot 2024-03-09 at 4 20 32 PM
Or click into each problem individually to see if starred and change status:
Screen Shot 2024-03-09 at 4 19 53 PM
Favorite problems prioritized in ordering for grading:
Screen Shot 2024-03-09 at 4 21 20 PM

## Motivation and Context

In classes where part of grading involves making annotations on submissions as part of certain problems, TAs requested the ability to select certain problems that will show up at the top of the scroll bar to assign a problem to a certain annotation made in order to save time and effort.
Resolves #2062.

## How Has This Been Tested?

Test different variations of "favorite problems" selected and ensure prioritized annotation feature works for all of the variations.

## Types of changes

## Checklist:


app/controllers/api/v1/problems_controller.rb
damianhxy, line 1: If you're modifying these functions, be sure to update the documentation for the API too! (https://docs.autolabproject.com/api-interface/#problems)

app/controllers/submissions_controller.rb
```diff
@@ -218,7 +218,9 @@ def download
 
     elsif params[:annotated]
       @problems = @assessment.problems.to_a
```
damianhxy, line 220: tbh, this seems like dead code that can be removed altogether. To be sure, try making a pdf submission, annotate it, then download it.

app/views/problems/_fields.html.erb
```diff
@@ -20,6 +20,10 @@ allow a student to see their total score without having a score for that
 problem. Use this option carefully. In most cases, you should leave it
 unchecked." %>
 
+<%= f.check_box :starred, help_text: "By default, all problems are
```
damianhxy, line 23: Would recommend elaborating further on this.

e.g. "Starred problems are displayed first in the problems dropdown when creating annotations"

app/views/assessments/_edit_problems.html.erb
```diff
@@ -1,6 +1,32 @@
 <script type="text/javascript">
```
damianhxy, line 1: Use `content_for`

app/controllers/submissions_controller.rb
jhs-panda: I think I use the @problems declaration to retrieve the problems and sort them. Let me know if you think it's dead code and I'll test it out though!

app/controllers/submissions_controller.rb
```diff
@@ -452,7 +454,9 @@ def view
             end
 
     @problems = @assessment.problems.to_a
-    @problems.sort! { |a, b| a.id <=> b.id }
+    @problems.sort_by! do |problem|
+      [problem.starred ? 0 : 1, problem.name]
+    end
```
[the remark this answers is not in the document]
jhs-panda: I think this is a bit of a hassle, but lmk if I should implement it!

app/assets/javascripts/annotations.js
```diff
@@ -532,8 +532,21 @@ function newAnnotationFormCode() {
...
+            $('<option disabled></option>').text('Starred Problems')
+        );
+        processStarred=true;
+      }
+      if(!problem.starred && processStarred){
+        box.find('select').append(
+            $('<option disabled></option>').text('-------------------')
+        );
+        processStarred=false;
+      }
```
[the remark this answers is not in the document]
jhs-panda: Tested suggest code, functionality not the same.

app/assets/javascripts/annotations.js
```diff
@@ -694,8 +707,21 @@ function initializeBoxForm(box, annotation) {
...
```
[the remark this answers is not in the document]
jhs-panda: Tested functionality, different. Did not implement suggestion.

app/views/problems/_fields.html.erb
jhs-panda: Rephrased in recent commit.

app/views/assessments/_edit_problems.html.erb
jhs-panda: implemented in recent commit

app/controllers/api/v1/problems_controller.rb
jhs-panda: Changed in recent commit.

app/views/assessments/_edit_problems.html.erb
```diff
@@ -34,7 +62,10 @@
   <tbody>
     <% for problem in @assessment.problems do %>
       <tr>
-        <td><%= problem.name %></td>
+        <td data-url="<%= course_assessment_problem_path(@course, @assessment, problem) %>">
+          <%= link_to "<i class=\"material-icons left star-icon\" data-problem-id=\"#{problem.id}\" data-starred=\"#{problem.starred ? 'yes' : 'no'}\">#{problem.starred ? 'star' : 'star_border'}</i>".html_safe, "javascript:void(0);", class: "small star-toggle", remote: true %>
+          <%= problem.name %>
+        </td>
```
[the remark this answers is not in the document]
jhs-panda: @damianhxy Tried suggested code, appearance and functionality very different. Did not take coderabbit suggestion.

app/views/assessments/_edit_problems.html.erb
```diff
@@ -1,7 +1,35 @@
-<script type="text/javascript">
-  is_positive_grading = <%= @is_positive_grading %>;
-  has_annotations = <%= @has_annotations %>;
-</script>
+<% content_for :javascripts do %>
+    <script type="text/javascript">
+        is_positive_grading = <%= @is_positive_grading %>;
+        has_annotations = <%= @has_annotations %>;
+
+        document.addEventListener('DOMContentLoaded', function() {
```
damianhxy, line 6: Nit: for consistency with other files, can we use `$document.ready( ... )` instead?

app/views/assessments/_edit_problems.html.erb
```diff
@@ -1,7 +1,35 @@
...
+                });
+            });
+        });
+
+        function toggleStar(icon) {
+            var starred = icon.getAttribute('data-starred');
+            starred = (starred === 'yes') ? 'no' : 'yes';
+            icon.innerHTML = (starred === 'yes') ? 'star' : 'star_border';
+            icon.setAttribute('data-starred', starred);
+            var starred_bool = (starred==='yes') ? 'true' : 'false';
```
damianhxy, line 21: Nit: add spaces around `===`

app/views/assessments/_edit_problems.html.erb
```diff
@@ -1,7 +1,35 @@
...
+            // Update the corresponding problem.starred attribute
+            var problemId = icon.getAttribute('data-problem-id');
```
damianhxy, line 23: Nit: `problemId` is unused, can be removed

app/views/assessments/_edit_problems.html.erb
```diff
@@ -1,7 +1,35 @@
-<script type="text/javascript">
-  is_positive_grading = <%= @is_positive_grading %>;
-  has_annotations = <%= @has_annotations %>;
-</script>
+<% content_for :javascripts do %>
+    <script type="text/javascript">
```
damianhxy, line 2: General style nit w.r.t. javascript code: prefer using `let/const` instead of `var`

app/views/assessments/_edit_problems.html.erb
damianhxy, line 1: I think it would be good to add some kind of indicator as to what the star icon represents. e.g. add a title to the icon via `link_to ..., title: "Toggle 'starred' status of problem"`.

Also, perhaps explain that the starred status is immediately updated by adding some explanatory text on the page?

app/views/assessments/_edit_problems.html.erb
```diff
@@ -1,7 +1,35 @@
...
+        function toggleStar(icon) {
+            var starred = icon.getAttribute('data-starred');
+            starred = (starred === 'yes') ? 'no' : 'yes';
+            icon.innerHTML = (starred === 'yes') ? 'star' : 'star_border';
+            icon.setAttribute('data-starred', starred);
+            var starred_bool = (starred==='yes') ? 'true' : 'false';
+            // Update the corresponding problem.starred attribute
+            var problemId = icon.getAttribute('data-problem-id');
+            var dataUrl = icon.parentNode.parentNode.getAttribute('data-url');
+            $.ajax({
```
damianhxy, line 25: There is some subtlety with the use of `ajax` here due to what the `update` method in `problems_controller` does.
1. it sets a `flash`. Usually, when a user saves a problem, they are redirected to the list of problems and so this would appear for the user. However, when using Ajax, the flash is set but not immediately "consumed", so it only shows up on the next page the user loads.
2. It performs a `redirect_to`. Usually, this is what redirects the user back to the list of problems. But when we are performing an Ajax call, it seems to follow the redirect also with the `patch` method, which results in an error.

Screenshot 2024-03-12 at 13 35 20

Screenshot 2024-03-12 at 13 35 34

Of course, the starred flag is still updated, but this isn't ideal.

------------------

If we do not want to add a new `update` method just for `Ajax` calls, then perhaps we should use params to distinguish between Ajax updates and "normal" updates.

On the frontend side, this can be specified by adding the params to the end of a path helper, e.g. `course_assessment_problem_path(@course, @assessment, problem, ajax: true)`

On the backend side, we can check if the flag is set with something like `params.include?(:ajax)`. Then you can avoid setting flashes and redirecting in this scenario.

app/assets/javascripts/annotations.js
damianhxy, line 1: Just a note, but this doesn't handle pdf annotations (the relevant function is `newAnnotationFormTemplatePDF`). 

Screenshot 2024-03-12 at 13 44 20

That said, pdf annotations have already diverged from normal annotations in terms of functionality such as adding to shared comment pool, displaying grading type, etc. so it's not a big deal not to deal with this for now, but it's something to KIV.

Besides, there are global annotations now so there's even less use for pdf annotations.

damianhxy (changes requested):
Functionality works, but left some comments regarding things that should be touched up for completeness

Also, be sure to fix the failing test

app/views/assessments/_edit_problems.html.erb
jhs-panda: Done

app/views/assessments/_edit_problems.html.erb
jhs-panda: Done

app/controllers/assessments_controller.rb
```diff
@@ -462,6 +462,7 @@ def installProblems
         p.description = problem["description"]
         p.max_score = problem["max_score"]
         p.optional = problem["optional"]
+        p.starred = problem["starred"]
```
damianhxy, line 465: Note that this is dead code that will be removed in #2118

app/views/assessments/_edit_problems.html.erb
```diff
@@ -12,9 +40,12 @@
   </label>
 </div>
 <label for="switch">
-  Negative Grading is where you grade by deducting from the max score. Positive Grading is where you grade by adding to 0.
+  Negative Grading is where you grade by deducting from the max score. Positive Grading is where you grade by adding to 0.<br>
```
damianhxy, line 43: Maybe put the `<br>` between the labels instead?

app/views/assessments/_edit_problems.html.erb
```diff
@@ -1,7 +1,35 @@
-<script type="text/javascript">
-  is_positive_grading = <%= @is_positive_grading %>;
-  has_annotations = <%= @has_annotations %>;
-</script>
+<% content_for :javascripts do %>
+    <script type="text/javascript">
+        let is_positive_grading = <%= @is_positive_grading %>;
```
damianhxy, line 3: This could be `const` instead just like `has_annotations`

damianhxy (approved):
LGTM other than nits

PR #2105, opened 2024-02-24
Added a button to show students with missing submissions

### LINKED ISSUE #2080: Make displaying of missing submissions on gradesheet a toggleable option
The gradesheet is used by some classes to assign grading ranges to TAs.

#2019 began including missing students by default which complicates the process since instructors have to sort missing submissions to the start / end. This makes the process error-prone and also causes the numbering to diverge between the gradesheet and "view source" page (which only counts actual submissions).

Thus, we should make this a toggleable option (disabled by default) to effectively revert to the old behavior.

Note: regardless, the gradesheet / view source pages always counted submissions by instructors / CAs / dropped students. If there will be a significant amount of submissions by them, further action might be needed to make the distribution of work fairer.

dwang3851:
## Description

- Added a button to toggle between showing students with completed submissions and all students on the view gradesheet screen
- By default, will only show students with completed submissions
- Use the viewall param in the url to accomplish this behavior
Screenshots:
image
image

## Motivation and Context
- Easier than sorting by submissions a couple times
- Illiano asked for it
- Resolves #2080 

## How Has This Been Tested?
- Tested in local environment from instructor and TA account
- Used an assessment that had both missing and non-missing submissions
- Ensured that all non-missing submissions were present before the button is toggled, and that all submissions were present when it was not
- Ensured this behavior remained consistent with multiple button clicks, and for both instructors and TAs

## Types of changes

## Checklist:

- Noticed current bug where the button to sort submissions is broken on the first click, and resolves itself after two more clicks


20wildmanj:
Note: for the "How Has This Been Tested?" portion of the PR you should outline the specific steps you took to test the new behavior (and also how you want your reviewer to test the PR, taking note to what the reviewer should be looking for)

app/views/assessments/viewGradesheet.html.erb
```diff
@@ -74,28 +75,35 @@
                              else
                                @course.course_user_data
                              end %>
-      <% for c in matching_students.where.not(id: has_submissions) do %>
-          { "id": "",
+        <% url = request.original_url %>
```
20wildmanj, line 78: This style of getting the params does not really follow the rails paradigm, this kind of checking the params can be easily done in the controller action that runs before the view is rendered. Try doing params logic in  the `viewGradesheet` method in `controllers/assessment/grading.rb`

20wildmanj: actually you wouldn't even need to do anything in the controller, just use `params[:viewall]` which you seem to be already doing for the toggle logic?

app/views/assessments/viewGradesheet.html.erb
```diff
@@ -238,6 +246,61 @@ Interaction between gradesheet and annotations
 <span id="interaction_badge" class="badge" data-badge-caption="Feature Information"></span></a>
 </div>
 
+<script type="text/javascript">
```
20wildmanj, line 249: javascripts should be in a `content_for` block

app/views/assessments/viewGradesheet.html.erb
```diff
@@ -238,6 +246,61 @@ Interaction between gradesheet and annotations
...
+    var loadFunc = (elem) => {
```
20wildmanj, line 250: Is this function used anymore? Also, why are we using local storage here? There is also a console log that is unnecessary.

app/views/assessments/viewGradesheet.html.erb
```diff
@@ -238,6 +246,61 @@ Interaction between gradesheet and annotations
...
+<div class="switch">
+  <label>
+    <b>Show students with completed submissions</b>
+    <% if params[:viewall] == "true" %>
+      <input type="checkbox" onclick="clickFunc(this)" id="view_all_checkbox" checked>
+    <% else %>
+      <input type="checkbox" onclick="clickFunc(this)" id="view_all_checkbox">
+    <% end %>
+    <span class="lever"></span>
+    <b>Show all students </b>
```
20wildmanj, line 270: nite: space

app/views/assessments/viewGradesheet.html.erb
```diff
@@ -238,6 +246,61 @@ Interaction between gradesheet and annotations
...
+</div>
+<label for="switch">
+  All students includes those with missing submissions
+</label>
+
+<script defer>
+    document.addEventListener('DOMContentLoaded', function() {
+        var checkBox = document.getElementById('view_all_checkbox');
+
+        function updateURL() {
```
20wildmanj, line 281: why not just redirect to `viewGradesheet` with the param set, instead of doing all of this url manipulation and reloading?

app/views/assessments/viewGradesheet.html.erb
```diff
@@ -238,6 +246,61 @@ Interaction between gradesheet and annotations
...
+    <% end %>
+    <span class="lever"></span>
+    <b>Show all students </b>
+  </label>
+</div>
+<label for="switch">
+  All students includes those with missing submissions
+</label>
+
+<script defer>
```
20wildmanj, line 277: should be in a `content_for` block

20wildmanj (changes requested):
Functionally this works, but I would like to see some changes to how you're implementing them to be more rails-like, using params and also jquery instead of url manipulation. Let me know if you need help / have any questions.

20wildmanj (approved):
Looks much cleaner! LGTM :))

Issue #1713, opened 2023-01-29
Update new course error message UI


michellexliu:
- "_ errors prohibited this course from being saved" label is too big
- Should make the errors listed more apparent
![Image](https://user-images.githubusercontent.com/50491000/215334575-13141658-1522-4a9b-92ca-9b212d4b34d6.png)

damianhxy:
Potentially reference #1689

PR #2124, opened 2024-03-18
Updated new course error message UI

coder6583:
## Description
- Removed large error message from top of the page
- Added required attribute to text field
- Added error text under the input field for invalid name

## Motivation and Context
Issue #1713 
It is easy for the user to input an invalid course name, but the error messages are unclear on why the name is invalid. The error message is also too large, and it is not a very intuitive UI.

## How Has This Been Tested?
Before:
image
After:
image
Gives suggested name for invalid name under the corresponding input field.
image
Error popup when field is not entered.

Tested with different invalid names that include spaces and special characters.

## Types of changes

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Introduced inline error handling and messages for course creation form fields.
	- Added a custom validation method for course names with suggestions for correction.
	- Enhanced form fields with required validations and specific attributes for better usability.
- **Bug Fixes**
	- Removed duplicate style declarations and redundant newlines in form builder.
- **Style**
	- Updated styles for error messages to be more visible and user-friendly.
- **Refactor**
	- Simplified error state handling in forms for a cleaner user experience.
	- Refactored course name validation to improve reliability and user guidance.
- **Database Changes**
	- Adjusted database schema for better precision, data integrity, and added a new `starred` field to mark importance.


app/assets/stylesheets/style.css.scss
jhs-panda, line 1: (not file specific)
When I was testing the feature, I noticed that the alignment of the name entry was off when the course is created and an error is hit. However, I saw your screenshots where the alignment looks right, so I'm not sure if it's a universal issue.
Here are some screenshots: 
Entering fields to create course:
Screen Shot 2024-03-24 at 10 53 38 AM

Display if error hit while trying to create course:
Screen Shot 2024-03-24 at 10 53 21 AM

db/schema.rb
jhs-panda, line 1: maybe git pull from main branch to get rid of the schema addition since it's not tied to this specific pr?

db/schema.rb
```diff
@@ -292,6 +292,7 @@
     t.datetime "updated_at"
     t.float "max_score", default: 0.0
     t.boolean "optional", default: false
+    t.boolean "starred", default: false
```
jhs-panda, line 295: maybe git pull from main branch to get rid of the schema addition since it's not tied to this specific pr?

app/models/course.rb
jhs-panda, line 1: The suggested names are a nice addition!

jhs-panda (commented):
Overall, looks pretty good! The display looked a bit off when I was testing and created a new course with errors (see screenshots), but yours looked fine from the screenshots in the PR, so I'm not sure if the issue is universal. Maybe see if you can reproduce the issue and try fixing it, but I'm also not sure if this issue was caused by this PR specifically so... 😄

app/assets/stylesheets/style.css.scss
coder6583: I fixed this problem with the latest commit, but it might break other forms using FormBuilderWithDayTimeInput. I tested other forms, and they seem to work, but I might need extra testing with other pages using FormBuilderWithDayTimeInput.

jhs-panda: New commit seems to solve issue in this commit! As long as it doesn't break the other forms, I think this PR LGTM!

db/schema.rb
jlge: +1, there should be no changes to `schema.rb` in this PR

jhs-panda (commented):
Please check that nothing else breaks with the changes in the code and that the errors are fixed. Otherwise, LGTM!

app/views/courses/_create_new_course.html.erb
```diff
@@ -1,20 +1,35 @@
-<h3>Create New Course</h3>
+<% @title = "Create New Course" %>
```
jlge, line 1: Since this is a partial you don't need the `@title` tag

app/views/courses/_create_new_course.html.erb
```diff
@@ -1,20 +1,35 @@
...
+
+<h4>Create New Course</h4>
```
jlge, line 3: This is inconsistent with the Import Course tab header now, so I would change the header of that tab as well

app/models/course.rb
```diff
@@ -9,7 +9,8 @@ class Course < ApplicationRecord
   validates :grace_days, numericality: { greater_than_or_equal_to: 0 }
   validates :version_threshold, numericality: { only_integer: true, greater_than_or_equal_to: -1 }
   validate :order_of_dates
-  validates :name, format: { with: /\A(\w|-)+\z/, on: :create }
+  validate :valid_name?, on: :create
+  # validates :name, format: { with: /\A(\w|-)+\z/, on: :create }
```
jlge, line 13: Delete this line

app/views/courses/new.html.erb
```diff
@@ -1,5 +1,3 @@
-<% @title = "Create New Course" %>
```
jlge, line 1: Keep the title on the page itself rather than the partial (since partials could be used in multiple places)

app/views/courses/_create_new_course.html.erb
```diff
@@ -1,20 +1,35 @@
...
+                                        ""
+                                      else
+                                        @newCourse.errors.full_messages_for(:name)[0]
+                                      end %>
 
-  <%= f.text_field :semester, size: 3,
-                              help_text: "The current semester. Examples: f12, s13, m13" %>
+  <%= f.text_field :semester, size: 3, required: true,
+                              help_text: "The current semester. Examples: f12, s13, m13",
+                              error_text: if @newCourse.errors.full_messages_for(:semester).empty?
```
jlge, line 20: It doesn't seem like there is a case where there will be errors in semester so not sure if this is necessary, correct me if I'm wrong though

jlge:
image
This error message should probably be changes to something along the lines of "Course creation failed. Please review the fields below" or "Please check all fields and try again."

jlge (changes requested):
Some style changes + dead code should be removed, but otherwise functionally works! Also should try and fix the failing tests :')

app/views/courses/_create_new_course.html.erb
```diff
@@ -1,20 +1,26 @@
 <h3>Create New Course</h3>
 
 <%= form_for :newCourse, url: courses_path, builder: FormBuilderWithDateTimeInput do |f| %>
-  <%= f.error_messages %>
-
-  <%= f.text_field :name, required: true,
+  <%= f.text_field :name, size: 3, required: true,
```
jlge, line 4: I don't think `size` attribute is needed here, it just sets the width of the text field but the text field spans the width of the entire page

app/views/courses/_create_new_course.html.erb
```diff
@@ -1,20 +1,26 @@
...
+      and URL-able. Typically includes the course number and semester. Examples:
+      15213-f12, 15110-s11",
+                          error_text: if @newCourse.nil? || @newCourse.errors.full_messages_for(:name).empty?
+                                        ""
+                                      else
+                                        @newCourse.errors.full_messages_for(:name)[0]
+                                      end %>
 
-  <%= f.text_field :semester, size: 3,
+  <%= f.text_field :semester, size: 3, required: true,
```
jlge, line 14: same thing here with `size`

jlge (approved):
Some nits but LGTM otherwise!

PR #2137, opened 2024-04-01
Allow students to download all of their submitted files

### LINKED ISSUE #2068: Add a button that downloads all Autolab files
Currently, it requires the user to enter each individual assignment tab to download their previous code. I want to save my code from all my previous classes as reference/for practice later on, but I have to enter each class/lab in order to do so.

jhs-panda:
## Description

- added buttons on user's account page that when clicked, will either download all user's submissions from non-disabled courses in a zip file or flash a message if there are no submissions to download (give user option of downloading only final submissions or all submissions)
Screen Shot 2024-04-08 at 6 34 45 PM
- saves each submitted valid file in zip file for easy interpretation of contents
Only final submissions:
Screen Shot 2024-04-08 at 6 25 33 PM
All submisisons:
Screen Shot 2024-04-08 at 6 26 32 PM

## Motivation and Context

This PR was motivated by a feature request. Previously, students had to download hand in files individually in order to save. With this new feature, students can, with one click, download all of their downloadable submissions in one zip file.
Resolves #2068 

## How Has This Been Tested?

Tested the feature with normal course settings (no disabled courses), disabled courses (no download of submissions allowed), and no submissions to download (flashed message instead).
Also tested in case where user is instructor of one class and a student enrolled in another. In this case, the downloaded zip file should only contain the user's submissions as a student, not any of the submissions in the course they are an instructor of.

## Types of changes

## Checklist:

## Summary by CodeRabbit

- **New Features**
	- Introduced the ability for users to download all their submissions from their profile page.
- **Enhancements**
	- Updated user display settings for a more intuitive user experience.
- **Bug Fixes**
	- Made improvements to handle submission downloads and filenames in the `UsersController`.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -154,7 +154,8 @@ class UsersController < ApplicationController
     submissions = Submission.where(course_user_datum: CourseUserDatum.where(user_id: user))
     submissions = submissions.select do |s|
       p = s.handin_file_path
-      !p.nil? && File.exist?(p) && File.readable?(p)
+      is_disabled = s.course_user_datum.course.disabled
+      !p.nil? && File.exist?(p) && File.readable?(p) && !is_disabled
     end
     if submissions.empty?
       flash[:error] = "There are no submissions to download."
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -162,8 +162,9 @@ class UsersController < ApplicationController
     ###
     filedata = submissions.collect do |s|
       p = s.handin_file_path
-      email = s.course_user_datum.user.email
-      [p, download_filename(p, email)]
+      course_name = s.course_user_datum.course.name
+      assignment_name = s.assessment.name
+      [p, download_filename(p, course_name, assignment_name)]
     end
     result = Archive.create_zip filedata
     if result.nil?
@@ -173,7 +174,7 @@ class UsersController < ApplicationController
     send_data(result.read, # to read from stringIO object returned by create_zip
               type: "application/zip",
               disposition: "attachment", # tell browser to download
-              filename: "submissions.zip")
+              filename: "autolab_submissions.zip")
   end
 
   # PATCH users/:id/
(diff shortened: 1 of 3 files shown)


app/controllers/users_controller.rb
```diff
@@ -148,6 +148,40 @@ def edit
...
+      flash[:error] = "There are no submissions to download."
+      redirect_to(user_path(user)) && return
+    end
+    filedata = submissions.collect do |s|
+      p = s.handin_file_path
+      course_name = s.course_user_datum.course.name
+      assignment_name = s.assessment.name
+      [p, download_filename(p, course_name, assignment_name)]
+    end
+    result = Archive.create_zip filedata
```
jlge, line 174: Probably should add some error handling in case this fails

app/controllers/users_controller.rb
jlge, line 1: It would be a good idea to have nested directories for submissions rather than putting everything into one

app/controllers/users_controller.rb
```diff
@@ -148,6 +148,40 @@ def edit
...
+    if result.nil?
+      flash[:error] = "There are no submissions to download."
+      return
+    end
+    send_data(result.read, # to read from stringIO object returned by create_zip
+              type: "application/zip",
+              disposition: "attachment", # tell browser to download
+              filename: "autolab_submissions.zip")
```
jlge, line 182: nit:  you can add the date or some unique string to the filename so that if the user downloads multiple times it's not the same file name

app/views/users/show.html.erb
```diff
@@ -104,4 +104,19 @@
     </ul>
     <%= f.submit 'Save', { class: "btn primary" } %>
   <% end %>
+  <h4>Download Submissions</h4>
```
jlge, line 107: nit: I would put final before all but that is just my personal preference :D

jlge (changes requested):
Functionally works, but the directory structure for the final zip should be changed

app/controllers/users_controller.rb
jhs-panda: Done in recent commit

app/controllers/users_controller.rb
jhs-panda: Done in recent commit

jlge (approved):
LGTM, great job 😄

PR #2140, opened 2024-04-03
Use File.size? when checking for config files

damianhxy:
## Description

Use `File.size?` instead of `File.exist?` when checking the existence of config files, namely
- `smtp_config.yml`
- `github_config.yml`
- `oauth_config.yml`
- `lti_config.yml`
- `lti_platform_jwk.json`
- `lti_tool_jwk.json`

## Motivation and Context

For Docker compose installs, the `Makefile` automatically creates empty config files (using `touch`) when a user updates their installation. This is to avoid problems where the Docker volume mapping creates a directory instead of a file when the path does not exist.

However, the codebase currently uses `File.exist?` when checking for the existence of config files, which breaks since the empty config files do exist, causing the code to attempt to load configuration from blank files.

By replacing the size with `File.size?` instead, it returns `nil` when the file does not exist OR the file exists but is empty. This is the desired semantics.

## How Has This Been Tested?

Create blank config files in the `config` subdirectory, e.g. `touch smtp_config.yml github_config.yml oauth_config.yml lti_config.yml lti_platform_jwk.json lti_tool_jwk.json`. Run Autolab.

Previously, there would be a variety of errors, starting with a failure to load github config.

After, Autolab should run fine, and e.g. the upload buttons for the LTI `json` files will not have the message saying that the files already exist.

Ensure that all relevant File existence checks in the codebase have been replaced with `File.exist?`

## Types of changes

## Checklist:

### DIFF
--- a/app/controllers/admins_controller.rb
+++ b/app/controllers/admins_controller.rb
@@ -37,7 +37,7 @@ class AdminsController < ApplicationController
   def autolab_config
     @github_integration = GithubIntegration.check_github_authorization
 
-    if File.exist?("#{Rails.configuration.config_location}/lti_config.yml")
+    if File.size?("#{Rails.configuration.config_location}/lti_config.yml")
       @lti_config_hash =
         YAML.safe_load(File.read("#{Rails.configuration.config_location}/lti_config.yml"))
     end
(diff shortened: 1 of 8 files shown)


20wildmanj (approved):
Verified that fix works and allows autolab to load, doesn't break existing functionality. LGTM.

PR #2145, opened 2024-04-11
Order assessment attachments correctly

NicholasMy:
## Description

This update fixes the sorting order of assessment attachments.

## Motivation and Context

When rendering attachments on the **assessment** page, they're ordered how the database returns them instead of by the desired "release_at ASC, name ASC" ordering constant. **Course** attachments are correct and it already calls `.ordered` in the same way I did here. If an instructor wants to re-order attachments, they currently need to delete them and re-upload them. This fix would allow them to only update the release date to re-order them. I also believe this was the intended behavior, but forgetting to call `.ordered` was an oversight.

## How Has This Been Tested?
I created two course attachments, A and B, with B's release date after A's. I verified B appears **after** A in the list. I updated B's release date to be before A's. I verified B appears **before** A in the list. 

## Types of changes

Potentially breaking change: Some instructors may depend on the "manual" method of ordering attachments as described above. That was ultimately a bug, and it never should have behaved that way, but maybe this should wait until the semester is over.

## Checklist:

## Summary by CodeRabbit

- **Bug Fixes**
	- Updated the attachment fetching logic in assessments to include ordering based on user role, enhancing usability and accessibility.

### DIFF
--- a/app/controllers/assessments_controller.rb
+++ b/app/controllers/assessments_controller.rb
@@ -584,9 +584,9 @@ class AssessmentsController < ApplicationController
                       @cud
                     end
     @attachments = if @cud.instructor?
-                     @assessment.attachments
+                     @assessment.attachments.ordered
                    else
-                     @assessment.attachments.released
+                     @assessment.attachments.released.ordered
                    end
     @submissions = @assessment.submissions.where(course_user_datum_id: @effectiveCud.id)
                               .order("version DESC")
--- a/app/controllers/assessments_controller.rb
+++ b/app/controllers/assessments_controller.rb
@@ -71,9 +71,9 @@ class AssessmentsController < ApplicationController
                                       .or(announcements_tmp.where(system: true)).order(:start_date)
     # Only display course attachments on course landing page
     @course_attachments = if @cud.instructor?
-                            @course.attachments.where(assessment_id: nil)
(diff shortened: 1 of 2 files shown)


damianhxy:
Thanks for catching this - it was an oversight on my part

Just a thought: course attachments are ordered on the course page due to `@course_attachments.from_category(cat).ordered`.

For consistency, I think either
1. course attachments should be ordered on the backend instead, inside `assessments_controller.rb#index`
2. assessment attachments should be ordered on the frontend instead, e.g. `render @attachments.ordered`

Option (1) is probably better, in which case the same `.ordered` scope should be applied when defining `@course_attachments`. This presumes that applying a `where` filter (via `from_category`) preserves order, although I can't find anything information online regarding this.

What do you think?

damianhxy:
> I think doing the same for course attachments makes sense.

That sounds good to me, could you make the change to order course attachments on the backend too?

> If I correctly order them before applying the where, it keeps the correct order, and the resulting SQL still explicitly orders it, so I don't think this is a coincidence.

In that case, I think it's safe to remove the ordering for course attachments on the frontend. Thanks for investigating this!

NicholasMy:
No problem, how does this look?

damianhxy (approved):
Tested that attachments are ordered as expected on course / assessment pages

LGTM!

Issue #2111, opened 2024-02-26
Speedgrader file outline breaks on certain archives


damianhxy:
Speedgrader File Outline is wonky when viewing an archive submission where the files have been tarred in an "unconventional way".

To reproduce: submit (after unzipping) [tartraversal2.tar.zip](https://github.com/autolab/Autolab/files/14410963/tartraversal2.tar.zip) and view the submission
(it was created by having files traverse two levels up first, e.g. `tar -cf tartraversal2.tar ../../Desktop/MossTest/autograde-Makefile ../../Desktop/MossTest/dave.c ../../Desktop/MossTest/hello.c`)

![Screenshot 2024-02-26 at 15 36 10](https://github.com/autolab/Autolab/assets/9074856/af2aafb3-99de-49e1-a3f4-7ffc7d75cfb1)

The problem seems to lie in the `sanitize_directories` method of `archive.rb`

PR #2146, opened 2024-04-13
Fixed Speedgrader file outline on unconventional tar files

coder6583:
## Description
- Changed it so the filename of a path is the last element instead of the 2nd element

## Motivation and Context
The filenames in tar files made using `..` are weird, like [tartraversal2.tar.zip](https://github.com/autolab/Autolab/files/14410963/tartraversal2.tar.zip) made with `tar -cf tartraversal2.tar ../../Desktop/MossTest/autograde-Makefile ../../Desktop/MossTest/dave.c ../../Desktop/MossTest/hello.c`. See screenshot below:
image

See issue #2111.

## How Has This Been Tested?
After:
image

- Tested with  [tartraversal2.tar.zip](https://github.com/autolab/Autolab/files/14410963/tartraversal2.tar.zip) ↑
- Tested with tar files made with both bsdtar and gnutar: bsdtars made on mac specifically have metadata that also appear in the file outline
  - I could not find a way to fix this, so one way to fix this might be to limit the tar command to be gnutar
- Tested with tar files with various directory levels

## Types of changes

## Checklist:

Do you think this is enough, or should I try to hide the metadata in the tar files?

### DIFF
--- a/lib/archive.rb
+++ b/lib/archive.rb
@@ -45,7 +45,7 @@ module Archive
     end
     subFiles = []
     filesNestedSomewhere = files.select{|entry| entry[:pathname].start_with?(root[:pathname]) && !(entry[:pathname] == root[:pathname])}
-    for file in filesNestedSomewhere
+    filesNestedSomewhere.each do |file|
       fileDepth = file[:pathname].chomp("/").count "/"
       if(fileDepth == depth+1)
         subFiles << recoverHierarchy(filesNestedSomewhere, file)
@@ -67,11 +67,12 @@ module Archive
     starting_header = -1
 
     # add pre-existing directories to the set
-    for file in files
+    files.each do |file|
 
       # edge case for removing "./" from pathnames
+      # file[:pathname] = file[:pathname].gsub(/\.\.\//) {'..-'}
       if file[:pathname].include?("./")
-        file[:pathname] = file[:pathname].split("./")[1]
+        file[:pathname] = file[:pathname].split("./").last
       end
 
       if(file[:directory])
@@ -79,13 +80,13 @@ module Archive
       end
     end
 
-    for file in files
+    files.each do |file|
       # for each file, check if each of its directories and subdir
       # exist. If it does not, create and add them
       if(!file[:directory])
         paths = file[:pathname].split("/")
         mac_bs_file = false
-        for path in paths do
+        paths.each do |path|
           # note that __MACOSX is actually a folder
           # need to check whether the path includes that
           # for the completeness of cleaned_files
@@ -96,7 +97,7 @@ module Archive
              break
           end
         end
-        for i in 1..(paths.size - 1) do
+        (1..(paths.size - 1)).each do |i|
           new_path = paths[0,paths.size-i].join("/") + "/"
           if(!file_path_set.include?(new_path))
             cleaned_files.append({


lib/archive.rb
```diff
@@ -67,25 +67,26 @@ def self.sanitize_directories(files)
     starting_header = -1
 
     # add pre-existing directories to the set
-    for file in files
+    files.each do |file|
 
       # edge case for removing "./" from pathnames
+      # file[:pathname] = file[:pathname].gsub(/\.\.\//) {'..-'}
```
damianhxy, line 73: Could we get rid of this dead code?

lib/archive.rb
```diff
@@ -67,25 +67,26 @@ def self.sanitize_directories(files)
...
       if file[:pathname].include?("./")
-        file[:pathname] = file[:pathname].split("./")[1]
+        file[:pathname] = file[:pathname].split("./").last
```
damianhxy, line 75: Actually, this approach is rather fragile and would break (not show the correct structure) on paths like `./b/./a/./a.txt` (multiple `./` in the middle)

Screenshot 2024-04-15 at 13 54 04

What do you think about using something like `File.cleanpath(file[:pathname], rel_root = true).gsub("..", "__PARENT__")`?

`cleanpath` (https://www.rubydoc.info/gems/yard/File.cleanpath) will strip any `./`, and redundant `../` (e.g. `./b/../c` -> `c`)

The `gsub` would then help to indicate the correct relative hierarchy (for the remaining `../`):

Screenshot 2024-04-15 at 13 51 32

For testing, here are some files:
[test2.tar.zip](https://github.com/autolab/Autolab/files/14982686/test2.tar.zip)
**Contents**
```
./b/./a/./a.txt
```
[test.tar.zip](https://github.com/autolab/Autolab/files/14982691/test.tar.zip)
**Contents**
```
./a.txt
../b.txt
../../b/b.txt
../../c.txt
../../b/b.txt
../../../c/b/a/a.txt
```

damianhxy (changes requested):

lib/archive.rb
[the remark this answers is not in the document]
damianhxy: This is not the intention of the code, we want to preserve the path structure as far as possible

coder6583: I didn't know `cleanpath` was a thing, it looks really good for this purpose!
I don't think this would be an accurate representation of the directory structure, since `a.txt` should be inside the __PARENT__ directories. However, I understand it might look confusing for the user if the file that should have been in the base directory was inside a nested directory. What do you think?

coder6583: Also, I think having the __PARENT__ directories might obstruct the view of the users from seeing the actual file outline clearly, considering the small width of the file tree.
image

damianhxy: Given that parent directories shouldn't occur that often, I don't think the obstructed view is a large concern.

I also think the behavior is fine regarding `a.txt` not being inside the `__PARENT__` directory. That seems to be the behavior for other software too (e.g. The Unarchiver)

Screenshot 2024-04-20 at 12 49 29

damianhxy (approved):
LGTM

lib/archive.rb
```diff
@@ -96,7 +96,7 @@ def self.sanitize_directories(files)
              break
           end
         end
-        for i in 1..(paths.size - 1) do
+        (1..(paths.size - 1)).each do |i|
```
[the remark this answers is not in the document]
damianhxy: @coder6583 this seems reasonable

PR #2147, opened 2024-04-15
Nil check for file manager

KesterTan:
Fixed nil check for file manager (https://nightly.autolabproject.com/file_manager/comptest/hello/autograde-Makefile)

## Description
- Opens and reads files that have no mimetype

## How Has This Been Tested?
- Upload a file that has no extensions like README and autograde-Makefile, check that when you click into it that the file opens and contents are displayed as text (no nil errors should occur)
- Test that breadcrumbs and back button also work as expected

## Types of changes

## Checklist:

### DIFF
--- a/app/controllers/file_manager_controller.rb
+++ b/app/controllers/file_manager_controller.rb
@@ -33,17 +33,14 @@ class FileManagerController < ApplicationController
     elsif File.file?(absolute_path) && check_instructor(absolute_path)
       if File.size(absolute_path) > 1_000_000 || params[:download]
         send_file absolute_path
+      elsif !is_binary_file?(absolute_path)
+        @path = path
+        @file = absolute_path.read
+        render :file, formats: :html
       else
-        mime_type = MimeMagic.by_path(absolute_path)
-        if mime_type.nil? || mime_type.type.split('/').first == 'text'
-          @path = path
-          @file = absolute_path.read
-          render :file, formats: :html
-        else
-          send_file(absolute_path,
-                    filename: File.basename(absolute_path),
-                    disposition: 'attachment')
-        end
+        send_file(absolute_path,
+                  filename: File.basename(absolute_path),
+                  disposition: 'attachment')
       end
     end
   end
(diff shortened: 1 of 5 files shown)


app/controllers/file_manager_controller.rb
```diff
@@ -34,8 +34,8 @@ def index
       if File.size(absolute_path) > 1_000_000 || params[:download]
         send_file absolute_path
       else
-        mime_type = MimeMagic.by_path(absolute_path).type
-        if mime_type.split('/').first == 'text'
+        mime_type = MimeMagic.by_path(absolute_path)
+        if mime_type.nil? || mime_type.type.split('/').first == 'text'
```
damianhxy, line 38: This might not be the best approach because it would attempt to show binary files too (assuming there are no extensions)

How about doing it similar to submissions controller?

```ruby
  def is_binary_file?(file)
    mm = MimeMagic.by_magic(file)
    mm.present? && (!mm.text? && (mm.subtype != "pdf"))
  end
```

You can drop the pdf check, and display anything that isn't a binary file.

KesterTan: okay, fixed.

KesterTan:
The errors were the same ones we had in file manager, they're appearing again cause I moved the if else loop into a elsif loop

docs/instructors.md
```diff
@@ -178,4 +178,25 @@ Manually assigned grades are by default not released, and therefore not visible
...
+
+- You cannot create a folder or upload a file with a name that already exists.
+
+- Deleting a folder will also delete all of its contents.
+
+- Downloading a folder will create a tar with all of its contents.
+
+- Download Selected downloads each of the selected files/folders separately.
+
+- Clicking on a binary file downloads it else the file's contents are displayed.
```
damianhxy, line 201: I think this could be better phrased. Maybe something along the lines of "When clicking on a file, it will be automatically downloaded if it is larger than 1GB or a binary file. Otherwise, the file contents will be displayed."

damianhxy (approved):
Able to view extensionless files now

docs/instructors.md
KesterTan: changed it to this.

PR #2165, opened 2024-05-23
Improve scheduler visual run handling of return value from update method

damianhxy:
## Description

Check whether return value of `update` method has a `to_str` method (used for implicit conversions), rather than just if it's non-`nil`.

## Motivation and Context

The scheduler visual run output feature introduced in #1916 expects that users either return a string or `nil` from the `update` method.

However, if an existing user was not aware of the change, they may not have an explicit return. As a result of Ruby's implicit returns, they might return a value that is non-`nil` yet can not be converted to a string, leading to an error.

## How Has This Been Tested?

(Follow the steps in #1916 if unsure how to set up a scheduler)

**returning a string**
```rb
module Updater
    def self.update(course)
        out = ""
        out << "my output\n"

        out
    end
end
```
Screenshot 2024-05-23 at 08 17 19

**returning nil**
```rb
module Updater
    def self.update(course)
        nil
    end
end
```
Screenshot 2024-05-23 at 08 17 00

**returning non-nil value that can't be converted to a string**
```rb
module Updater
    def self.update(course)
        true
    end
end
```
Note: `true` has a `to_s` method but NOT a `to_str` method. So implicit conversions fail.
(same picture as above)

### Updated documentation
Screenshot 2024-05-23 at 08 16 50

## Types of changes

## Checklist:

### DIFF
--- a/app/controllers/schedulers_controller.rb
+++ b/app/controllers/schedulers_controller.rb
@@ -56,7 +56,7 @@ class SchedulersController < ApplicationController
         begin
           require mod_name
           output = Updater.update(action.course)
-          if output
+          if output.respond_to?(:to_str)
             fork_log << "----- Script Output -----\n"
             fork_log << output
             fork_log << "\n----- End Script Output -----"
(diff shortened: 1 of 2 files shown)


20wildmanj (approved):
worked locally, lgtm

PR #2170, opened 2024-07-04
Display start and due date-time for assessment according to system timezone

ecwu:
## Description

As described in #2169, the displayed date-time is not in the correct timezone as the system. This change convert the date object to the local time zone before formatting it into a string.

## Motivation and Context
To resolve this, I adjusted the `date_to_s` method in `models/assessment.rb` to utilize the `getlocal` method, which should fetch the system's timezone. This ensures the date and time are displayed correctly according to the system's local timezone.

## How Has This Been Tested?
I have tested this fix on my end with system time zone set to Asia/Shanghai, and the date and time are now displayed correctly.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have run rubocop and erblint for style check. If you haven't, run `overcommit --install && overcommit --sign` to use pre-commit hook for linting
- [ ] My change requires a change to the documentation, which is located at [Autolab Docs](https://docs.autolabproject.com/)
- [ ] I have updated the documentation accordingly, included in this PR

## Other issues / help required

If unsure, feel free to submit first and we'll help you along.

PR #2171, opened 2024-07-25
Revert "Display start and due date-time for assessment according to system timezone (#2170)"

20wildmanj:
This reverts commit b6a4387d703c95b5c29819871af80515baed62bf.

## Description

#2170 changed visual cues such that they would display due-dates based on the system's local time. However, this causes deployments which are located in a different timezone from the user to display a visual cue with the time of that deployment's timezone instead of the user's timezone, with no indication of the timezone. The timezone is shown on the handin page using the user's timezone, so the difference between the two is bound to cause confusion and thus the previous commit should be reverted for the time being.

## How Has This Been Tested?

With the timezone change, see that the visual cue times are shown in EDT but the user is in PDT, causing the times to be different, but importantly the visual cue does not show a timezone and does not match the user's timezone.
Screenshot 2024-07-25 at 12 02 29 AM
Screenshot 2024-07-24 at 11 56 24 PM

Reverting the change, see that there is no longer a mismatch of the user timezone values (in PDT):
Screenshot 2024-07-24 at 11 56 21 PM
Screenshot 2024-07-24 at 11 56 24 PM

## Types of changes

### DIFF
--- a/app/models/assessment.rb
+++ b/app/models/assessment.rb
@@ -430,7 +430,7 @@ class Assessment < ApplicationRecord
   end
 
   def date_to_s(date)
-    date.getlocal.strftime("%b %e at %l:%M%P")
+    date.strftime("%b %e at %l:%M%P")
   end
 
   def load_dir_to_tar(dir_path, asmt_dir, tar, filters = [], export_dir = "")


KesterTan (approved):
LGTM

Issue #2168, opened 2024-06-30
Internal error on sending email after a new autolab installation


abrantesasf:
**Your environment**
Let us know details such as 
- Universidade Vila Velha (UVV) Autolab at https://autolab.computacaoraiz.com.br
- Rocky Linux 9.4
- Autolab v2.12.0
- This is a new installation of Autolab, in a new server, to substitute our old autolab server (the old autolab is v2.10)

**Steps To Reproduce**
1. Install Autolab according "Autolab + Tango Docker Compose Installation", on Autolab Docs
2. Configure e-mail according "Mailing for Autolab Docker Installation", on Autolab Docs
3. Test e-mail using "Forgot your password", on UVV Autolab page

**Current behavior**
When I try to test e-mail using the "Forgot your password" feature, on https://autolab.computacaoraiz.com.br, an Error 500 Internal Server Error is triggered. Our currently email configuration, in `.env` file is this:

`# SMTP SETTINGS

SMTP_SETTINGS_ADDRESS=smtp.gmail.com

SMTP_SETTINGS_PORT=587

SMTP_SETTINGS_ENABLE_STARTTLS_AUTO=true

SMTP_SETTINGS_AUTHENTICATION=login

SMTP_SETTINGS_USER_NAME=our_user_name

SMTP_SETTINGS_PASSWORD=our_password

SMTP_SETTINGS_DOMAIN=computacaoraiz.com.br

SMTP_DEFAULT_FROM=autolab@computacaoraiz.com.br

HOST_PROTOCOL=http

HOST_DOMAIN=autolab.computacaoraiz.com.br`

This same configuration are working as expected on the old server, with Autolab 2.10. But, in this new server with Autolab 2.12, the same email settings are not working.

We test the email function after disabling SELinux and Firewall, but the internal error continues.

In the Autolab logs this message is displayed:

`autolab  | App 191 output: W, [2024-06-30T15:29:56.977753 #191]  WARN -- : An error occurred when sending a notification using 'email' notifier.Errno::ECONNREFUSED: Connection refused - connect(2) for "localhost" port 25

autolab  | App 191 output: /usr/local/rvm/gems/ruby-3.2.2/gems/net-smtp-0.4.0/lib/net/smtp.rb:631:in `initialize'

autolab  | App 191 output: /usr/local/rvm/gems/ruby-3.2.2/gems/net-smtp-0.4.0/lib/net/smtp.rb:631:in `open'

[...]

PR #2175, opened 2024-08-04
Fix an incorrect setting name (SMTP authentication)

GuillaumeDIDIER:
## Description
In the smtp yaml file, the setting is called 'authentication', however in config/environments/ it is looked up as 'plain'. Fix this issue by correcting the look-up to use the correct name.

## Motivation and Context
This solves an issue with e-mail when the SMTP server requires a different authentication method, such as cram_md5.

(Note it appears the name plain may come from a time at which the setting was a boolean plain)

## How Has This Been Tested?
Verified on my production server that this change fixed the ability to send e-mail (which resulted in error 500 otherwise)
My server requires cram_md5.

I would recommend this change be tested on a test environment with a different SMTP server.

## Types of changes

## Checklist:

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -111,7 +111,7 @@ Rails.application.configure do
       address: config_hash['address'],
       port: config_hash['port'],
       enable_starttls_auto: config_hash['enable_starttls_auto'],
-      authentication: config_hash['plain'],
+      authentication: config_hash['authentication'],
       user_name: config_hash['user_name'],
       password: config_hash['password']
     }
--- a/config/environments/production.rb.template
+++ b/config/environments/production.rb.template
@@ -129,7 +129,7 @@ Autolab3::Application.configure do
       address: config_hash['address'],
       port: config_hash['port'],
       enable_starttls_auto: config_hash['enable_starttls_auto'],
-      authentication: config_hash['plain'],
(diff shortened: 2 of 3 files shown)


jhs-panda (approved):
Hi! Thank you so much for this PR and for addressing this issue, the code changes look good.

jhs-panda (changes requested):
Hi! Sorry, this is a nitpick, but could you please edit the documentation (docs/installation/mailing.md) to say "authentication:       'plain', # Other options include: 'login', 'cram_md5'" instead of "authentication:       'login'" since with your PR other SMTP authentications work rather than it just working on 'plain'.

GuillaumeDIDIER:
I can probably make that change, but the documentation seems quite inconsistent with the current implementation.

(right now the default production.rb pulls from the SMTP yaml, and the docker environment variables appear to have no effect, so the doc should probably get rid of those variables)

damianhxy:
Docs are known to be outdated as noted in #2168

GuillaumeDIDIER:
I made the requested change to the documentation. It still doesn't reflect the current state of the template.

If I have time to investigate the various places mentioned in #2168, should I make a separate PR to update the docs and the config file to a consistent state?

jhs-panda (approved):

PR #2181, opened 2024-08-23
add validation of yaml fields before loading them during import

20wildmanj:
## Description

It is possible that after exporting an assessment, instructors may modify the yaml configuration file manually, leading to bad / blank values. Therefore, in order to ensure that if bad values / nil values are provided to assessment import, we must validate some important fields like `name`, `handin_directory`, `handin_filename`, `display_name`, and dates. There was some validation of dates, but the checks did not include the case where the dates could be empty strings which technically can be parsed but will cause an error later down the road.

For example there can exist these errors:
- if the `name` field in the yaml file is different from the folder / yaml configuration filenames, then the `name` field of the assessment will become the one in the yaml file, but then Autolab can no longer find the folder of the assessment, which causes an error
- if the `handin_directory` field is not provided, this may also cause an error when trying to execute `assessment.rb:construct_folder` (constructing a handin folder for the assessment)
- if the `handin_filename` field is not provided, this causes a 500 error when trying to view the homepage
- if `display_name` is not provided, the assessment import will not be successful
- if dates provided are blank strings, the assessment import will fail validation of start / due dates

The below changes fix all of these cases.

## Motivation and Context

Improves assessment import flow and overall robustness by handling invalid yaml file fields.

## How Has This Been Tested?

1. Use this configuration file: [homework04.yml.txt](https://github.com/user-attachments/files/16721111/homework04.yml.txt)
2. create a folder `homework04` within a course, and try using the import assessment feature, see that import is successful
3. delete the assessment, and then go into `homework04.yml` and either delete the field or make it a blank string, see that import is successful and display name is the same as the assessment name
4. delete the assessment, go into `homework04.yml` and change the name to different from the file name (for example homework05), see that import is successful
5. delete the assessment, go into `homework04.yml` and remove the handin_directory and handin_filename fields, see that import is successful and handin_directory / handin_filename are given default values
6. delete the assessment, go into `homework04.yml` and make all date fields empty string, import and see that dates are autofilled

Note:  currently after deleting the assessment, the yaml fields may end up inconsistent with their corresponding values in the UI.

## Types of changes

## Checklist:

### DIFF
--- a/app/models/assessment.rb
+++ b/app/models/assessment.rb
@@ -300,6 +300,36 @@ class Assessment < ApplicationRecord
     File.open(asmt_yaml_path, "w") { |f| f.write(YAML.dump(sort_hash(serialize))) }
   end
 
+  # If the name field in a yaml for an assessment differs from the folder name
+  # then the name field for the assessment gets overridden, which then causes
+  # the yaml file to no longer be found, and then for cleanup to fail because
+  # the folder can no longer also be found. For example a folder with
+  # homework01 and a homework01.yml file but with a name field of homework02
+  # would cause assessment.name = homework02 so then it would try to find
+  # homework02/homework02.yml. Instead we rename the name field to match the folder
+  # name.
+  # Similarly for a bunch of other fields in the yaml, if they are not provided,
+  # then we may run into issues loading the assessment
+  def validate_yaml(yaml_map)
+    unless yaml_map["general"]
+      raise "General section missing in yaml"
+    end
+
+    if yaml_map["general"]["name"] != name
+      yaml_map["general"]["name"] = name
+    end
+    # ensure display name is present, otherwise fill in with name
+    if yaml_map["general"]["display_name"].blank?
+      yaml_map["general"]["display_name"] = name
+    end
+    if yaml_map["general"]["handin_directory"].blank?
+      yaml_map["general"]["handin_directory"] = "handin"
+    end
+    return if yaml_map["general"]["handin_filename"].present?
+
+    yaml_map["general"]["handin_filename"] = "handin.c"
+  end
+
   ##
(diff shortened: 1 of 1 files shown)


app/models/assessment.rb
```diff
@@ -300,6 +300,36 @@ def dump_yaml
...
+  # the folder can no longer also be found. For example a folder with
+  # homework01 and a homework01.yml file but with a name field of homework02
+  # would cause assessment.name = homework02 so then it would try to find
+  # homework02/homework02.yml. Instead we rename the name field to match the folder
+  # name.
+  # Similarly for a bunch of other fields in the yaml, if they are not provided,
+  # then we may run into issues loading the assessment
+  def validate_yaml(yaml_map)
+    unless yaml_map["general"]
+      raise "General section missing in yaml"
```
KesterTan, line 315: Is this caught by importAssessmentsFromFileSystem? It seems to catch StandardError and I think this is a TypeError.

app/models/assessment.rb
```diff
@@ -300,6 +300,36 @@ def dump_yaml
...
+    end
+
+    if yaml_map["general"]["name"] != name
+      yaml_map["general"]["name"] = name
```
KesterTan, line 319: Yml file does not auto propagate the name field when it does not exist or it is different from name. Seems to be a bug that occurs for every yml file.

KesterTan: But does not affect the functionality otherwise and okay to ignore if theres no quick fix.

KesterTan (approved):
LGTM

app/models/assessment.rb
20wildmanj: Going to add to another issue

PR #2185, opened 2024-08-30
Hotfix: Fix scoreboard check of index on base case

20wildmanj:
#2092 introduced various error-checking mechanisms to prevent accidentally rendering bad scoreboard data. However, it introduced a bug where for assessments without column specifications that use the problem scores to populate the scoreboard, if the number of students on the scoreboard exceeds the number of problems, then further student submissions would show that their scoreboard entry errored, when in reality their scoreboard data is fine. 

This PR removes the faulty check.

## How Has This Been Tested?

1. Use the following autograded assessment with 1 problem:
[randomlab5_20240830.tar.zip](https://github.com/user-attachments/files/16821462/randomlab5_20240830.tar.zip)

2. Submit submissions for at least 3 students

3. On master, see that there is a display error when viewing the scoreboard

Screenshot 2024-08-30 at 3 35 05 PM

4. On this branch, see that the error icon goes away and a score is displayed:

Screenshot 2024-08-30 at 3 35 32 PM

## Types of changes

### DIFF
--- a/app/views/scoreboards/show.html.erb
+++ b/app/views/scoreboards/show.html.erb
@@ -75,7 +75,7 @@
             <% end %>
           <% else %>
             <%# this should be guaranteed to be an array, but for redundancy, check that entry is array %>
-            <% if grade[:entry].is_a?(Array) && grade[:entry].length > i %>
+            <% if grade[:entry].is_a?(Array) %>
               <% grade[:entry].each do |column| %>
                 <td><%= column %></td>
               <% end %>


coder6583 (approved):
Looks good to me!

KesterTan:
Should this line also be changed?
Screenshot 2024-08-30 at 10 09 18 PM

20wildmanj:
> Should this line also be changed? Screenshot 2024-08-30 at 10 09 18 PM

No, that line is necessary for the case where there is a column specification but the number of specified columns is greater than the number of score entries.

Issue #2204, opened 2024-09-15
Unable to add new user in nightly


KesterTan:
Unable to create a new user in Nightly, am getting a 500 error.

**Steps To Reproduce**
1. Navigate to Manage Autolab
2. Click on Manage Users
3. Click on Create User on top right hand
4. Fill out the form

**Current behavior**
![Screenshot 2024-09-15 at 12 36 28 PM](https://github.com/user-attachments/assets/5e0b249d-60cd-4070-9613-e66f31f6505c)
500 error. This seems to be due to an SMTP error because of a sendgrid configuration.
![Screenshot 2024-09-15 at 12 38 22 PM](https://github.com/user-attachments/assets/68a1e4d1-f1ae-40a6-ba9c-70919629a2ab)

**Expected behavior**
![Screenshot 2024-09-15 at 12 37 29 PM](https://github.com/user-attachments/assets/1159b88e-3624-4305-8c5e-19d5fed21bbd)
There should be a green tooltip indicating that a user has been successfully created.

PR #2206, opened 2024-09-15
Override password reset email send when creating user if SMTP not set up

jhs-panda:
## Description

If SMTP is not properly set up, creating a user will not send the typical password reset email. Instead, it will create the user but flash a message to the user.

## Motivation and Context

With SMTP Config set up with single-sender, the password reset email does not send successfully when creating a new user. Issue #2204  
In the case that SMTP is not set up successfully, a user is still created but a message is flashed that the password reset email was not successfully sent. 

## How Has This Been Tested?

Before:
Creating new user
Screen Shot 2024-09-15 at 2 06 51 PM

After:
Creating a user but SMTP not set up.
Screen Shot 2024-09-29 at 2 08 28 PM
Screen Shot 2024-09-29 at 2 08 45 PM

If error other than SMTP comes up, flash error message and destroy user (since already created but if unusable, no point in keeping).
Screen Shot 2024-09-29 at 2 10 04 PM

## Types of changes

## Checklist:


20wildmanj:
Could you change this so that instead of just commenting out the send password reset lines, there is code to catch the exception and flash the error?

KesterTan:
I'm getting this error with this config

Screenshot 2024-10-06 at 12 40 39 PM

Screenshot 2024-10-06 at 12 43 17 PM

Screenshot 2024-10-06 at 12 43 24 PM

app/controllers/users_controller.rb
```diff
@@ -117,9 +117,19 @@ def create
...
-      redirect_to(users_path) && return
+      begin
+        @user.send_reset_password_instructions
+        flash[:success] = "Successfully created user."
+      rescue Net::SMTPFatalError
+        flash[:success] = "Successfully created user but reset password instructions were not sent:
+          Net::SMTPFatalError"
+      rescue StandardError => e
+        error_message = e.message
+        flash[:error] = "Failed to create user: #{error_message}"
```
KesterTan, line 128: This flash error is slightly confusing to me, I've been getting error messages like localhost(25) and auth errors when SMTP is not configured and I don't think I would know it was related to my SMTP config unless I know that an email was trying to be sent. Could we change this error message to say something like `Failed to create user: incorrectly configured SMTP config : #{error_message}` or something.

jhs-panda: Done.

KesterTan (approved):
LGTM

PR #2210, opened 2024-09-20
Allow SMTP auth to be nil

### LINKED ISSUE #2172: Allow SMTP Authentication to be nil
**Your environment**
Carnegie Mellon Autolab Instance

From Chaskiel: "The new email sending config interface does not allow using an smtp server that doesn't require authentication. It should be possible to set authentication to nil."

**Steps To Reproduce**
1. Go to 'Configure Autolab'
2. See that there exist multiple authentication options but no option to not have authentication

Screenshot 2024-07-29 at 9 42 26 PM

**Expected behavior**

There should be an option to not have authentication

KesterTan:
## Description
Resolves #2172 

Previously, users had no way of specifying smtp authentication in smtp config to be nil. However, with this fix, users can do so by clicking the option and not specifying a username and password.

Screenshot 2024-09-20 at 3 42 43 AM

## How Has This Been Tested?
1. To test this out, configure a local SMTP without auth. To do so, I ran this command:
`sudo python -m smtpd -n -c DebuggingServer localhost:25`.
2. You then fill in the Mail Server Address as localhost, Mail Server Port as 25 and click none for authentication. You should also fill in the information about your Autolab URL deployment under Autolab URL.
3. Try out a test email and see that it is it dumped to the console.

## Types of changes

## Checklist:

### DIFF
--- a/app/controllers/smtp_config_controller.rb
+++ b/app/controllers/smtp_config_controller.rb
@@ -84,9 +84,9 @@ private
       address: params['address'],
       port: params['port'],
       enable_starttls_auto: params['enable_starttls_auto'] == 'true',
-      authentication: params['authentication'],
-      user_name: params['username'],
-      password: params['password']
+      authentication: params['authentication'] == 'none' ? nil : params['authentication'],
+      user_name: params['username'] == "" ? nil : params['username'],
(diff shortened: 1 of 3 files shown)


jlge (approved):
LGTM 👍

PR #1934, opened 2023-06-29
Use omniauthv2-shibboleth for omniauth v2 compat

cg2v:
### Summary generated by Reviewpad on 29 Jun 23 14:07 UTC
This pull request updates the usage of the `omniauth-shibboleth` gem to version `1.3.2` in the Gemfile, replacing the previous version `1.1.2`. This change allows for compatibility with `omniauth` version 2.0.0 or higher. The Gemfile.lock has been updated accordingly.

## Description
Swap out the omniauth-shibboleth gem for omniauthv2-shibboleth (from https://github.com/lukaskoenen/omniauth-shibboleth)

## Motivation and Context

The original omniauth-shibboleth gem has been unmaintained since 2018, and is not compatible with omniauth 2.x and ruby 3.x. A possibly unintended consequence of #1380 was that omniauth 2.0 was introduced.

The omniauth-shibboleth gem was forked and updated so that gitlab could use shibboleth with omniauth 2.0 (https://gitlab.com/gitlab-org/gitlab/-/merge_requests/121235). Autolab can also make use of it!
This change does NOT obsolete #1593. There are use cases for both generic SAML authentication and shibboleth

## Types of changes

- [ X ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] I have run rubocop for style check. If you haven't, run `overcommit --install && overcommit --sign` to use pre-commit hook for linting
- [ ] My change requires a change to the documentation, which is located at [Autolab Docs](https://github.com/autolab/docs)
- [ ] I have updated the documentation accordingly, included in this PR

## Other issues / help required

I will run tests shortly and update here.

PR #2215, opened 2024-09-22
Remove redundant and outdated shibboleth gem

20wildmanj:
Remove the `omniauth-shibboleth` gem, which was already removed in #1934 but was unintentionally re-introduced in #2028. 

Not really able to test on our end, but is pretty much the same as #1934, so shouldn't cause issues.


KesterTan (approved):
LGTM

PR #2229, opened 2024-10-04
File manager exception fix

### LINKED ISSUE #2228: Fix File Manager Authentication and Add File Manager Tests
File Manager should route the user to the index page and not allow loading if the user is not verified to access that path.

KesterTan:
## Description
This PR fixes the routing when a user is not allowed to access a path. This also adds unit tests which tests this functionality for this fix.

Resolves #2228 

## How Has This Been Tested?
- Create a course and assign a student to that course
- Navigate to that course in the file manager and copy the link
- Sign in as a student using dev login and try to access the link
- Note that you will not be able to see anything in that path and note the flash[:error]
- Run the new unit tests: `rake spec SPEC=./spec/controllers/file_manager_controller_spec.rb` and make sure that everything passes

## Types of changes

## Checklist:


jlge (changes requested):
When I login as a user and access the `file_manager` path, it shows the base screen, but no error flash is displayed. 
image
I think a more ideal behavior would be to redirect back to the Autolab homepage and flash the "You are not authorized to view this page." message there. Also when an unauthorized user views the file manager, they can still press the "create folder" and "upload files" buttons which I know don't work but I think it's unideal, so this can be prevented by redirecting out to a different page.

jlge (approved):
lgtm!

PR #2235, opened 2024-10-04
Create endpoint for public LTI JWKs

20wildmanj:
## Description

Create a way for other LTI providers to discover LTI's public JWKs so that they don't need to be manually uploaded to the tool provider, which simplifies setup and makes switching out the JWT convenient.

documentation to be updated after we update the canvas flow / other LTI changes.

## Motivation and Context

Part of upgrades to support better LTI integration with Canvas. This should also make setting up Autolab on Canvas easier.

There is now a public endpoint `/lti_launch/jwks` to support querying Autolab for its public JWKs.

## How Has This Been Tested?

This testing was done on Nightly since it seems that the LTI testing tool really doesn't like http requests for making a request to get the public JWK.

Use the following platform: [https://lti-ri.imsglobal.org/platforms/5276](https://lti-ri.imsglobal.org/platforms/5276)

Verify that platform has correct JWK url:
Screenshot 2024-10-04 at 4 14 29 PM

Use the following configuration:
Screenshot 2024-10-04 at 4 11 55 PM

Upload whatever JWK keypair you want on nightly. Then go to the following link:
[https://nightly.autolabproject.com/lti_launch/jwks](https://nightly.autolabproject.com/lti_launch/jwks)

See that you see the public JWK (not the private one):
Screenshot 2024-10-04 at 4 12 58 PM

Also see that you can be logged out and still view the JWK (as it should be public-facing)

Then try synchronizing the course roster here (for now the LTI link is here): [https://nightly.autolabproject.com/courses/hackingsession-david/users](https://nightly.autolabproject.com/courses/hackingsession-david/users)

See that link succeeds
## Types of changes

## Checklist:

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -798,6 +798,7 @@ private
   def write_cuds(cuds)
     rowNum = 0
     rosterErrors = {}
+    rosterWarnings = {}
     rowCUDs = []
     duplicates = Set.new
 
@@ -863,7 +864,7 @@ private
         new_cud.delete(:year)
 
         # Build cud
-        if !user.nil?
+        if !user.nil? && !existing
           cud = @course.course_user_data.new
           cud.user = user
           params = ActionController::Parameters.new(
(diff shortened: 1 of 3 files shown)


coder6583 (approved):
LGTM!

Issue #2191, opened 2024-09-11
log_submit should presumably check the assessment dates


cg2v:
`log_submit` doesn't check the assessment's dates. Assuming it's kept around, it should presumably check if it's ok to accept submissions for the requested user. One possibly elegant way to do this is to create a new `Submission` object (but never save it). If the submission would be `allowed?` then it's ok to record the log entry, otherwise it should be refused

PR #2251, opened 2024-12-08
Resolving Issue 2191 - log_submit now validates for date

ogometz:
The new additions add additional validation tot he log_submit functionality for assessments. Now, the dates are checked for validity before submitting.

## Description
This pull requests has modified the log_submit method in the AssessmentsController to ensure that no assessments are being submitted without proper date validation. The code checks if the current time is before the start time of the assessment, or if the current time is beyond the end time of the assessment. If either condition is met, and error is thrown. This pull request addresses pull request number 2191. This pull request suggests on potential solution: " One possibly elegant way to do this is to create a new Submission object (but never save it). If the submission would be allowed? then it's ok to record the log entry, otherwise it should be refused." This option was initially explored. However, given the brevity of the fix, it was determined that creating a new object would likely serve to confuse and further complicate the functionality. And since the original log_submit function already has certain validations and checks, adding the date check only made sense.

## Motivation and Context
The motivation for this pull request is to address an issue brought up, linked here: https://github.com/autolab/Autolab/issues/2191
With these changes, the proper checks will be ran for proper code quality and coverage.

## How Has This Been Tested?
Accompanied by these code changes are several new tests, which have been added to the AssessmentController test suite found here: spec/controllers/assessments_controller_spec.rb. Three new tests have been added. The first executes the log_submit functionality with validated dates. The second attempts to do the same but with the submittion date before the start date of the assessment. Lastly, the third test seeks to do the same but with a submission date after the end date. As expected, the original code causes the first test to pass, but the last two to fail. With the code change, all three tests pass.

## Types of changes

- [ X ] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [ X ] I have run rubocop and erblint for style check. If you haven't, run `overcommit --install && overcommit --sign` to use pre-commit hook for linting

### DIFF
--- a/app/controllers/assessment/handin.rb
+++ b/app/controllers/assessment/handin.rb
@@ -270,6 +270,13 @@ module AssessmentHandin
       render(plain: err, status: :bad_request) && return
     end
 
+    # Validate assessment dates
+    current_time = Time.current
+    if current_time < @assessment.start_at || current_time > @assessment.end_at
+      err = "ERROR: Submissions are not allowed outside the assessment period"
+      render(plain: err, status: :bad_request) && return
+    end
+
     @result = params[:result]
     render(plain: "ERROR: No result!", status: :bad_request) && return unless @result
 
--- a/spec/controllers/assessments_controller_spec.rb
+++ b/spec/controllers/assessments_controller_spec.rb
@@ -212,4 +212,99 @@ RSpec.describe AssessmentsController, type: :controller do
       end
     end
   end
+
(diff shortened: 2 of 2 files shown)


Nita242004:
Effectively enhances the log_submit function by adding robust date validation, improving the system's reliability and preventing errors from invalid inputs. Would be beneficial to include comprehensive tests for edge cases like leap years and diverse time zones to ensure the feature works universally.

Issue #2261, opened 2025-01-27
Trying to regrade/download/excuse/delete over 200 submissions turns into a bad request


coder6583:
**Your environment**
Newest Autolab branch

**Steps To Reproduce**
1. Select over 200 submissions
2. Try to regrade/download/excuse/delete

**Current behavior**
Goes to a bad request page
![Image](https://github.com/user-attachments/assets/1b27c994-f21e-4bdf-a008-9e36d1bac5f9)

**Expected behavior**
Successfully execute the regrade/download/excuse/delete

This issue is related to [PR#2248](https://github.com/autolab/Autolab/pull/2248). Because the regrade all button is gone, we are currently putting all of the submission IDs into the url of the regrade/download/excuse/delete request. For example:

![Image](https://github.com/user-attachments/assets/7ad3b849-d338-4f2d-8e41-8b550a2382b9)
Because the url is so long, it is turning into a bad request. A possible fix is splitting the requests up, or putting the ids into the body of the request.

KesterTan:
Create submissions still capped at 300 users

PR #2271, opened 2025-02-23
Changing the manage submissions button to be ajax requests

dwang3851:
## Description
- Changed the regrade selected, delete selected, and excuse buttons to be ajax post requests
- Can't do this for download selected because it is a get request, and we shouldn't pass data in request body for get requests

## Motivation and Context
- Previously, all request information was contained in the request url which was not ok for large requests
- Resolves issue #2261 

## How Has This Been Tested?
- Tested in local env in manage submissions

## Types of changes

## Checklist:

- I am not sure why only the download selected button is properly disabled, when I console.log, the other buttons are the correct class
- Not sure if this is that user friendly, any input would be appreciated

### DIFF
--- a/app/assets/javascripts/manage_submissions.js
+++ b/app/assets/javascripts/manage_submissions.js
@@ -313,7 +313,6 @@ $(document).ready(function() {
       buttonIDs.forEach((id) => {
         const button = $(id);
         if (state) {
-          console.log(`Disabling: ${id}`, button);
           if (id === "#download-selected") {
             $(id).prop('href', baseURLs[id]);
           }
@@ -378,7 +377,6 @@ $(document).ready(function() {
         }
       });
     }
-
 
     changeButtonStates(true); // disable all buttons by default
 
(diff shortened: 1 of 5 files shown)


KesterTan:
Are the buttons visibility toggle broken? If no checkboxes are selected, buttons should be disabled?

Screenshot 2025-03-16 at 12 25 41 PM

app/assets/javascripts/manage_submissions.js
dwang3851, line 1: after further research, it seems it is highly illegal for security reasons to use ajax to download files (also i am not sure it is possible). I left download batch as is, I fixed some bugs with the single regrade button and error handling

KesterTan (approved):
Buttons work as expected, LGTM

app/assets/javascripts/manage_submissions.js
KesterTan: Agreed that it does not make sense

PR #2277, opened 2025-03-09
Regrade all and select checkbox message

KesterTan:
## Description
Regrade all button regrades submissions for every page. Encountered a cookie overflow problem when grading large amount of submissions so I changed the flash message to include counts instead of links.

Not every submission is normal/excused/NG so I changed regrade all to also include if a submission is none of the above.

Added a message that appears when the select all checkbox is clicked and all submissions in the page are clicked.
Screenshot 2025-03-09 at 6 03 37 AM

## How Has This Been Tested?
- Create course with more than 100 users
- Check that regrade all button works as expected
- Check that message appears when you select all submissions in a page and that regrade all link in the button works as expected

## Types of changes

## Checklist:

### DIFF
--- a/app/assets/javascripts/manage_submissions.js
+++ b/app/assets/javascripts/manage_submissions.js
@@ -158,9 +158,15 @@ $(document).ready(function() {
                   ${submission.total}
                 </td>
                 ${submission.problems.
-                map((problem) =>
-                  `<td class="submissions-td">${data.scores[submission.id]?.[problem.id]?.['score'] ?? "-"}</td>`
-                ).join('')}
+                    map((problem) =>
+                        `<td class="submissions-td">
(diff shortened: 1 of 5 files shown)


evanyeyeye (approved):
Mostly looks good. 

I noticed that for certain types of submissions the "Regrade Selected" button is hidden (like  for homeworks in the autogenerated course). However, the "Regrade All" button is still visible. I'm not sure what the exact condition is, but I'd imagine the behavior should be the same?

image

As a future item, it may be helpful to also show the number of selected elements even when its not 100.

KesterTan:
Regrade all should be hidden now when regrade is hidden 
Screenshot 2025-04-13 at 11 15 02 PM

Issue #2223, opened 2024-09-24
Make regrade status of a submission more clear


jhs-panda:
Requested by 15213.
Currently, when a submission is regraded, TAs are unable to tell what the status of the regrade is since it only displays a score (can't tell if still regrading or if autograder finished running already and score did not change).
Proposed solution: Better visual indicator for submissions that are still being regraded (ex. clock symbol to replace scores while being regraded)

20wildmanj:
Adding on from 122, they recently had an issue where they could not easily figure out the progress of regrade all.  That makes sense. It is totally possible that I clicked "regrade all" twice given that, like many things on Autolab, there is no visual feedback that something is happening until the relevant page gets loaded, which can take a while. (For "regrade all", that's typically an error page several minutes in -- but we've learned that it's not a real error, just the GUI timing out or something).

PR #2278, opened 2025-03-10
Manage submissions page shows progress for regrading selected submissions

coder6583:
## Description
When a submission is being graded, a clock icon replaces the score. There is a message on top of the page that tells the user to either check the jobs page or refresh the page so that they update the score. The message only goes away when there is no more submissions that are being graded. 
image

## Motivation and Context
Currently, when the user pressed the regrade selected button on Manage Submissions, it is unclear when the regrade all ended. The only way a user knows if the regrade has finished is if they go to the jobs page and see that all the jobs have finished running. This was a problem raised in issue #2223, by 213 and 122 TAs. 

## How Has This Been Tested?
I first submitted files to an autograder set up incorrectly, such as not adding a problem to the assessment. After the submission has been graded, add the missing problem to the assessment, and regrade those submissions that have the wrong scores. Check that when the submissions are being regraded, a clock icon appears instead of the score. Refresh the page until the message on the top of the page disappears, and check that the scores have been updated to a correct score. 

## Types of changes

## Checklist:

One issue with my implementation is that all submissions that are currently graded will have the status of being regraded. Therefore, if I regrade all submissions, then another user decides to submit a new file while the regrade is running, the message will only go away when that new submission also finishes grading. One way I thought of to address this issue is just making the clock icons and the yellow message a default feature regardless of if a regrade is going on or not. Therefore, if an instructor goes to Manage Submissions while a student creates a new submission, the instructor will see on the page that one grading process is going on.

### DIFF
--- a/app/assets/stylesheets/_variables.scss
+++ b/app/assets/stylesheets/_variables.scss
@@ -25,3 +25,4 @@ $autolab-selected-gray: #f5f5f5;
 $autolab-border-gray: #f4f1f1;
 $autolab-gray-text: #676464;
 $autolab-highlight-comments: #008000;
+$autolab-yellow: #f7e788;
--- a/app/assets/stylesheets/style.css.scss
+++ b/app/assets/stylesheets/style.css.scss
@@ -1718,3 +1718,13 @@ table.sub td, th {
   margin-top: 5px;
   color: $autolab-grey;
 }
+
+.regrade-refresh {
+  color: $autolab-gray-text;
+  background-color: rgba($autolab-yellow, 0.3);
+  border: 1px solid $autolab-yellow;
+  border-left: 5px solid $autolab-yellow;
+  padding: 10px;
+  border-radius: 3px;
+  font-weight: 500;
+}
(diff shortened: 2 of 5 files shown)


jhs-panda (commented):
I think the functionality is a bit broken? I am submitting with Proxylab (takes a long time to run so can test the regrade update properly). 

After regrading some submissions and letting them finish running (no jobs left), I regraded one more submission that was not previously regraded. At this point, the Jobs list is empty and none of the submissions have the timer. 
The "current regrading" count includes some submissions that were already regraded, and these submissions had the timer marking even though the jobs had already been run and did not show up on the Jobs list "waiting" or "running" sections.  Eventually, after some refreshing, the timer symbol and regrading message disappear.
Screen Shot 2025-04-01 at 2 58 20 AM
Screen Shot 2025-04-01 at 2 58 40 AM
It also seems like each time this error occurs, the same submissions are considered "regrading again" even though they weren't.

Let me know if you can't replicate it and I'll try with another assessment/set of submissions. Looking at your code it doesn't seem obvious why this happening and tbh it isn't that big of a deal, but it could be confusing for instructors.

PR #2281, opened 2025-03-17
Add EC2 autograder features

evanyeyeye:
## Description
This PR enables the use and configuration of EC2 autograding instances via Autolab. 

Documentation will be added in a future PR. 

## How Has This Been Tested?
- Verified that requests to Tango properly utilize the EC2 vmms.
- Verified that configured defaults work correctly.
- Verified that custom access credentials can be passed to Tango on a per job basis.

## Types of changes

## Checklist:


KesterTan (approved):
LGTM -> tested on https://dev.autolabproject.com/

Futurecomputerfixer (changes requested):
Security (secrets handling)

Problem: access_key / access_key_id are stored and passed around as plaintext. They must be encrypted at rest, scrubbed from logs, and not echoed in the UI.

Changes:

Prefer IAM roles over keys

app/helpers/assessment_autograde_core.rb

In the EC2 block, if use_access_key is false, omit accessKey/accessKeyId entirely from job_properties (don’t send empty strings).

Encrypt at rest (follow GitHub integration precedent)

Add encrypted columns mirroring github_integrations.access_token_ciphertext.

New migration (replace plaintext columns introduced in this PR):

Files:

Replace db/migrate/20241205233214_add_ec2_ssh_fields_to_autograders.rb with one that adds:

access_key_ciphertext :text

access_key_id_ciphertext :text

instance_type :string, default: "" (kept)

Drop the plaintext access_key/access_key_id columns if already migrated.

Model:

app/models/autograder.rb

Use the same encryption library already used (e.g., Lockbox) to define:

encrypts :access_key
encrypts :access_key_id


where those virtual attributes map to *_ciphertext.

Gate presence validations with use_access_key (see “Validations” section).

Scrub from logs & params filtering

config/initializers/filter_parameter_logging.rb

Add: Rails.application.config.filter_parameters += %i[access_key access_key_id]

app/controllers/autograders_controller.rb

Ensure we never flash or log raw values.

UI should not echo secrets

app/views/autograders/_ec2_settings.html.erb

Change inputs to password-style and avoid prepopulating:

<%= f.password_field :access_key, display_name: "Access Key", value: nil, autocomplete: "new-password" %>
<%= f.text_field :access_key_id, display_name: "Access Key ID", value: nil, autocomplete: "off" %>


Add a small hint: “Leave blank to keep existing secret.”

Strong params & controller flow

Problem: Params are permitted in two calls; :use_access_key isn’t included; and empty-string secrets are sent downstream.

Changes:

app/controllers/autograders_controller.rb

Unify strong params and include all fields:

def autograder_params
  params.require(:autograder).permit(
    :autograde_timeout, :autograde_image, :release_score,
    :use_access_key, :access_key, :access_key_id, :instance_type
  )
end


In create defaults block, avoid initializing secrets to "". Prefer nil:

a.access_key_id = nil
a.access_key = nil
a.instance_type = "t3.micro" # see validations below


In update/create, if use_access_key is false, ignore any provided access_key/access_key_id (set to nil before save).

Validations & defensive checks

Problem: instance_type is free-form; secrets aren’t validated when enabled.

Changes:

app/models/autograder.rb

Add allow-list for instance types shown in the UI:

INSTANCE_TYPES = %w[
  t2.nano t2.micro t2.small t2.medium t2.large t2.xlarge t2.2xlarge
  t3.micro t3.small t3.medium t3.large
  c5.large c5.xlarge c5.2xlarge
  r5.large r5.xlarge r5.2xlarge
].freeze

validates :instance_type, inclusion: { in: INSTANCE_TYPES }, allow_blank: true


Gate secrets validation behind the toggle:

with_options if: :use_access_key do
  validates :access_key_id, presence: true, format: { with: /\A[A-Z0-9]{16,24}\z/, message: "looks invalid" }
  validates :access_key, presence: true
end


Add bounded timeout validation if not already present:

validates :autograde_timeout, inclusion: { in: 10..900 }

Tango payload assembly

Problem: Empty strings are sent; json built twice; feature flag should be respected server-side.

Changes:

app/helpers/assessment_autograde_core.rb

Build a Ruby hash first, then JSON once at the end; respect config.x.ec2_ssh and use_access_key:

job_properties = {
  "image" => @autograde_prop.autograde_image,
  "timeout" => @autograde_prop.autograde_timeout,
  "callback_url" => callback_url,
  "jobName" => job_name,
  "disable_network" => assessment.disable_network
}

if Rails.configuration.x.ec2_ssh
  job_properties["ec2Vmms"] = true
  job_properties["instanceType"] = @autograde_prop.instance_type.presence || "t3.micro"

  if @autograde_prop.use_access_key?
    job_properties["accessKey"]   = @autograde_prop.access_key
    job_properties["accessKeyId"] = @autograde_prop.access_key_id
  end
end

job_properties = job_properties.to_json


This ensures keys are omitted entirely when not used.

Frontend (JS & form UX)

Problem: Minor JS bug; console noise; better UX for disabled/secret fields.

Changes:

app/assets/javascripts/autograder.js

Fix .val() misuse and remove console log:

;(function() {
  $(document).ready(function () {
    function access_key_callback() {
      const checked = $(this).prop('checked');
      const $access_key_field = $('#autograder_access_key');
      const $access_key_id_field = $('#autograder_access_key_id');
      $access_key_field.prop('disabled', !checked);
      $access_key_id_field.prop('disabled', !checked);
      if (!checked) {
        $access_key_field.val('');
        $access_key_id_field.val('');
      }
    }

    $('#autograder_use_access_key').on('change', access_key_callback);
    access_key_callback.call($('#autograder_use_access_key'));

    // Optional: keep dropdown UI tweaks, but drop console.log
    $('#autograder_instance_type option').hover(
      function() { $(this).addClass('highlighted-option'); },
      function() { $(this).removeClass('highlighted-option'); }
    );
  });
})();


app/views/autograders/_ec2_settings.html.erb

Add pattern to Access Key ID (light client-side feedback):

<%= f.text_field :access_key_id, display_name: "Access Key ID",
                 value: nil, autocomplete: "off",
                 pattern: "[A-Z0-9]{16,24}",
                 title: "Looks like an invalid AWS Access Key ID" %>

Feature flag correctness

Problem: UI is gated by config.x.ec2_ssh, but server must guard too.

Changes:

Files:

config/environments/development.rb – OK to be true.

config/environments/production.rb.template – defaults to false (good).

Server guard:

app/helpers/assessment_autograde_core.rb

Ensure the EC2 payload branch only runs if Rails.configuration.x.ec2_ssh is true (already applied above).

Migrations / schema churn

Problem: schema.rb shows charset/timestamp engine churn unrelated to this PR; plaintext columns introduced for secrets.

Changes:

Files: db/schema.rb

Regenerate schema after replacing the plaintext secret columns with encrypted *_ciphertext. Avoid unrelated charset/timestamp diffs (ensure you run the same DB engine/version locally as CI).

Files: db/migrate/20241205233214_add_ec2_ssh_fields_to_autograders.rb, db/migrate/20241211042124_add_use_access_key_to_autograder.rb

Update to remove plaintext columns and add encrypted ones (see “Security” section). Keep use_access_key:boolean migration.

Tests

Problem: No tests cover new payload assembly and validations.

Add:

spec/helpers/assessment_autograde_core_spec.rb

Cases:

ec2_ssh=false → no ec2Vmms/instanceType/keys in payload.

ec2_ssh=true, use_access_key=false → ec2Vmms & instanceType present, no keys.

ec2_ssh=true, use_access_key=true → keys present.

spec/models/autograder_spec.rb

Validates inclusion for instance_type.

Validates secrets presence only when use_access_key=true.

dwang3851 (approved):

PR #2299, opened 2025-10-20
Revert ec2 dropdown

Futurecomputerfixer:
### Context
This PR reverts commit `afd0665` (“ec2-dropdown”), which replaced the free-text EC2 instance type with a curated dropdown list.

That change caused a silent fallback when a previously saved EC2 instance type (e.g., `m7.flex`) was no longer in the list—Autolab defaulted to the first dropdown option instead of warning the user.

### What this PR does
- Restores the original text input for `instance_type`
- Preserves all existing database values (no automatic defaulting)
- Leaves controller/model logic unchanged

### Why
A dropdown cannot cover all possible EC2 instance types instructors may want to use.  
Keeping a free-text input maintains full flexibility and prevents silent overwrites.

### Testing
1. Load an assessment that already has an `instance_type` set to a value not in the list.
2. Confirm the text field shows the exact saved value.
3. Save without changes → value remains unchanged.
4. Modify and save → new value persists.

### Follow-ups
- Add optional “suggested values” (HTML datalist) with guidance text.
- Add validation to prevent empty or unintended defaults.
- Add regression test to ensure no silent mutations.

### Risk
Low — this reverts to previously stable behavior.


KesterTan (changes requested):
As discussed, this PR should be about displaying an error message when the user uses an EC2 instance that is not in our list instead of reverting it back to a text input. Stretch goals can include some sort of text validation or dynamically pulling a list of updated instance types

PR #2303, opened 2025-10-30
Manage submissions speed

dwang3851:
## Description
This changes implements server-side pagination for the manage submissions page to improve loading times.

## Motivation and Context
Previously, it would take about 1.5 seconds for each page of the manage submissions page to load since we would render all of them at once. For example, for a submissions page with 400 submissions, it would take about 6 seconds. This change renders only one page at a time, to take a maximum of 1.5 seconds each time. 

## How Has This Been Tested?
Tested all of the manage submissions functions locally

## Types of changes

## Checklist:

### DIFF
--- a/app/assets/javascripts/manage_submissions.js
+++ b/app/assets/javascripts/manage_submissions.js
@@ -11,7 +11,10 @@ const buttonIDs = ['#regrade-selected', '#delete-selected', '#download-selected'
 let tweaks = [];
 let currentPage = 0;
 $(document).ready(function() {
-  var submission_info = {}
+  var submission_info = {};
+  var selectedStudentCids = [];
+  var selectedSubmissions = [];
+  var submissions_to_cud = {}; // Build dynamically from server data
   const EditTweakButton = (totalSum) => {
     if (totalSum == null) {
       return `
@@ -70,7 +73,7 @@ $(document).ready(function() {
   $(document).ready(function () {
     $('.modal').modal();
 
-    $('.score-details').on('click', function () {
+    $(document).on('click', '.score-details', function () {
       // Get the email
       const course_user_datum_id = $(this).data('cuid');
       const email = $(this).data('email');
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/manage_submissions.js
```diff
@@ -70,7 +73,7 @@ $(document).ready(function() {
   $(document).ready(function () {
```
KesterTan, line 73: Can we remove these double document.ready functions

app/assets/javascripts/manage_submissions.js
```diff
@@ -11,7 +11,10 @@ const buttonIDs = ['#regrade-selected', '#delete-selected', '#download-selected'
 let tweaks = [];
 let currentPage = 0;
 $(document).ready(function() {
-  var submission_info = {}
+  var submission_info = {};
+  var selectedStudentCids = [];
+  var selectedSubmissions = [];
```
KesterTan, line 16: should we make these sets, would this be faster when we check for inclusion?

KesterTan (changes requested):
Overall functionality works fine, nice that pagination finally works as expected. There's some delay in making the requests but sorting/searching still seem reasonably fast. Just have a few comments.

KesterTan:
Regrade all causes the whole page to re-render, the submissions seem to be empty for a while before it is loaded again.

As a result, the tooltip for regrade all also dissapears.

app/assets/javascripts/manage_submissions.js
dwang3851: some functions (e.g requestData) expect arrays, changing to sets would be slower for those

dwang3851:
> Regrade all causes the whole page to re-render, the submissions seem to be empty for a while before it is loaded again.
> 
> As a result, the tooltip for regrade all also dissapears.

The tooltip for regrade all does not disappear for me, was unable to reproduce this behavior

app/assets/javascripts/manage_submissions.js
dwang3851: i don't really know why, but the page does not work without both document.ready functions. I figured whoever made the double document.ready functions in the first place had some reason for doing it.

KesterTan:
https://github.com/user-attachments/assets/e6e5faa1-5870-4fae-811d-523e969dff74

Check recording above, there should be a tooltip that says x number of submissions regraded but none is shown.

Screenshot 2025-11-25 at 5 10 41 AM

Issue #2313, opened 2026-03-04
Security: Fail-Open Authorization — Students Can Export/Modify All Grades (bulkExport, bulkGrade_complete, regradeBatch)


lighthousekeeper1212:
## Summary

A security audit identified **8 vulnerabilities** in Autolab, with the most critical being a **fail-open authorization design** that allows enrolled students to export all grades and write arbitrary grade modifications.

## Root Cause: Fail-Open Authorization Framework

In `app/controllers/application_controller.rb` lines 110-122, when an action has NO `action_auth_level` declaration, `authenticate_for_action` silently allows access:

```ruby
def authenticate_for_action
  controller_whitelist = @@global_whitelist[params[:controller].to_sym]
  return if controller_whitelist.nil?
  level = controller_whitelist[params[:action].to_sym]
  return if level.nil?  # Missing auth level = NO CHECK
  ...
end
```

Any action missing its `action_auth_level` declaration has **zero role-based authorization**.

### 1. `bulkExport` Missing Auth — Students Export All Grades (CRITICAL - CWE-862)

**File:** `app/controllers/assessment/grading.rb` lines 6-14

`bulkExport` exports all student scores as CSV. It has NO `action_auth_level` declaration while sibling `bulkGrade` requires `:course_assistant` (1-of-N inconsistency). Any enrolled student can GET `/courses/:name/assessments/:name/bulkExport` to download every student's grades.

### 2. `regradeBatch` Missing Auth — Student Batch Regrading (CRITICAL - CWE-862)

**File:** `app/controllers/assessment/autograde.rb` line 88

The `action_auth_level :regradeBatch, :instructor` declaration is **commented out**. Any student can trigger batch regrading, causing DoS on the Tango autograding server.

### 3. `bulkGrade_complete` Missing Auth — Students Write Grades (HIGH - CWE-862)

**File:** `app/controllers/assessment/grading.rb` lines 50-75

The save step of bulk grading has no auth while the upload step requires `:course_assistant`. A student can POST directly to save grade modifications for any student.

### 4. `permit!` Mass Assignment in Assessment Edit (HIGH - CWE-915)

**File:** `app/controllers/assessments_controller.rb` line 1120

`edit_assessment_params` calls `ass.permit!` which permits ALL parameters including `course_id` (cross-course manipulation).

### 5. ScoresController Cross-Submission Score Creation (HIGH - CWE-639)

**File:** `app/controllers/scores_controller.rb` lines 47-49

[...]

PR #2317, opened 2026-03-30
Resolve security concerns from upstream issue #2313

NicholasMy:
This resolves the security vulnerabilities raised in #2313. (Opened after discussion with the maintainers.)

Problems 1, 2, and 3 were simply resolved by adding action_auth_level verifications. It looks like these were accidentally omitted during a refactoring in 2015.

I couldn't exploit problem 4, but it's still a problem in theory. I crafted requests that tried to manipulate parameters that we wouldn't want instructors to directly set, such as the assessment id and name, but the application ignored them. Even so, I updated edit_assessment_params to explicitly delete unwanted parameters for extra safety.

Problems 5 and 6 are from the ScoresController. It is indeed vulnerable, but it turns out this controller isn't used anywhere, so I removed it, its (empty) tests, and its routes.

Problem 7 did indeed allow an instructor to create a system-wide announcement. This was simple to fix by only allowing administrators to set the system parameter.

Problem 8, allowing anyone to use the bug report feature, is debatable. I didn't resolve it here, as this may be your desired functionality. I previously disabled this feature for UB.

## Types of changes

## Checklist:


jhs-panda (approved):
LGTM! Thank you so much for this patch!

